Issue #522, opened 2014-05-19
Show selected item on plot


pecko:
_[Migrated from Assembla ticket [524](https://www.assembla.com/spaces/Debrief/tickets/524) - reported by @IanMayo on 2013-11-21 07:10:26]_

It would be good if we highlighted the currently selected item on the plot.
1. I guess we could either extend Plottable objects with extra highlight()/unhighlight() methods.  Then the PlotEditor could register as a SelectionListener, and if the selection is a plottable, it could call the highlight() method.
2.  Alternatively (and more cheaply) the PlotEditor could draw a rectangle over the getBounds() area for the selected item,  Maybe this rectangle could have very feint shading.

Obviously, once something else gets Selected (even a non-Plottable) then the unhighlight() get called.

I guess that 2. is a quick-win that we can experiment with, and 1. is a larger block of work.

IanMayo:
Moving to GEF for edit support #2525 would improve this.

We'd render our track data (as normal) onto the canvas.  But, we'd provide selection cursors in a GEF layer over the top.  The cursor would appear over the *Selected* item, with corners/nodes as appropriate.

PR #604, opened 2014-05-20
Issue #522 - Using XOR mode to paint selected item(s)

pecko:
Migrated from Assembla - https://bitbucket.org/ianmayo/debrief/pull-request/74/issue-524-using-xor-mode-to-paint-selected


IanMayo:
Close, not worth fixing now.

PR #990, opened 2014-12-12
Issue984 ne

IanMayo:
This will store the new Debrief plugin to handle Natural Earth data.


IanMayo:
drop this, use Peco's PR

PR #1032, opened 2015-01-12
Issue984 sax persist - IGNORE, NOT FOR RELEASE

IanMayo:


IanMayo:
Close - we no longer need this.  We aren't storing this level of detail in NE.

Issue #1037, opened 2015-01-13
JRE clash


IanMayo:
For the new GeoTools branches we need JRE1.7 (I guess the GeoTools compiled jars are built using 1.7).

But, when I run the master branch on an Eclipse set to JRE1.7, I get:
    interface com.planetmayo.debrief.satc.model.generator.IContributionsChangedListener is not visible from class loader

(this is when I try to open the Maintain Contributions view).

It's taken a while to diagnose it, but when I switch the Eclipse default back to JRE1.6, the problem goes away.

So, it looks like there's a Java version clash between SATC & JRE1.7.

I hope you can reproduce this by switching your Eclipse prefs / Java / Installed JREs to use 1.7, then checkout the Debrief/master trunk, run Debrief, and try to open the Maintain Contributions View from the Show/View menu.

pecko:
What is our minimal JDK we want to support: 1.6 or 1.7?

IanMayo:
I was happy with 1.6. But I think we have some compiled GeoTools jars that require 1.7.

Based on this, I'm happy to move to 1.7

pecko:
If some plugin requires 1.7 we requires 1.7. Will check 1.7 compatibility.

pecko:
> I hope you can reproduce this by switching your Eclipse prefs / Java / Installed JREs to use 1.7, then 
> checkout the Debrief/master trunk, run Debrief, and try to open the Maintain Contributions View from 
> the Show/View menu.

I can't reproduce the issue. Could you try clear configuration (Launch dialog, Configuration tab, Clear checkbox)?
What JDK/JRE you are using in the Launch configuration (Main tab, Execution environment...)?

Seems you have some specific settings in your Eclipse environment because there isn't any issue when starting the Debrief distribution with JDK 1.7.

I have added the PR: https://github.com/debrief/debrief/pull/1038
It set target JDK to 1.6 for the geotools plugin because we really doesn't need JDK 1.7 here.

IanMayo:
For the bug I was having: Debrief started ok, the problem was only when I opened the Maintain Contributions View.

pecko:
Weird. I could understand that you have a issue with JDK 1.6 since the geotools plugin requires 1.7 and shouldn't be started.

I can open the Maintain Contributions View when using distribution and JDK 1.7 too.

Could you try after the PR?

PR #1038, opened 2015-01-14
Issue #1037 - Set target JRE to 1.6 for o.m.c.geotools plugin

pecko:


IanMayo:
checked on OSX. Fix works.

PR #1050, opened 2015-01-20
Issue984 gtrenderer refactor

pecko:


org.mwc.cmap.gt2Plot/src/org/mwc/cmap/gt2plot/proj/GtProjection.java
```diff
@@ -89,7 +90,7 @@ public GtProjection()
 			// we'll tell GeoTools to use the projection that's used by most of our
 			// charts,
 			// so that the chart will be displayed undistorted
-			_worldCoords = CRS.decode("EPSG:3395");
+			_worldCoords = CRS.decode("EPSG:4326");
```
IanMayo: Hmm, we do the same CRS.decode in line 96.  Then we create a Transformation that appears to transform between two identical projections.

Do you understand what we're doing here?

pecko: I think that it isn't obliged anymore. You probably try to fix TIF images without *prj file.

Issue #970, opened 2014-12-01
Improved Windows Installer


IanMayo:
Our current Windows installer (Advanced Installer) requires every file to be listed.

But, after a new release the names of lots and lots of files change.  

After dragging in all of the updated files, there is still quite a lot of manual plumbing/connecting together to be done (shortcuts, etc).

It would be better if we could adopt an Installer builder that allowed source files to be specified by top level folder, not just by filename.  Then, we would just have to specify "plugins", "features", etc.  Or even, just our top-level deployed folder.

Ideally the app would be open source, so that supporting developers could produce their own builds, if required, for testing.

Extra-ideally, it would be great if the app could develop MS Windows installers under Linux/OSX - to save me having to switch context...
### Initial
1. Just require **.msi** installer, with embedded 32bit JRE. Ideally the installer would be for 32bit & 64bit windows.  If only one can be supported, then 32bit.
2. To be installed into "Program Files\Debrief\DebriefNG"
3. Require "Debrief" startup group with, "DebriefNG", "DebriefNG User Guide", "DebriefNG tutorial", "Uninstall DebriefNG"
4. Must not require admin privilege, just user rights
5. Registers file types, as with Kumar's register-files script 
### Fresh requirements (28th Apr 2015)
1. Silent install into c:\Progr..86
2. Create MSI installers for data:  Natural Earth, Charts, VPF
3. Users will manage via "System Centre Configuration Manager" (SCCM)
### Follow-on steps
1. Require 32bit and 64 bit installers
2. Require "lightweight" installers that don't include JRE (but check if JRE is present in installer)

IanMayo:
Once we have a text-based project file for a MSI builder, I think there will be value in Ant steps to update the configurable values in the Builder based on values in the rest of the project.

Significantly, the debriefng.product file contains a version number (something like 3.0.306).  We should certainly extract this, and insert it into the MSI build.  We should probably also insert today's date - and consider any other metadata that can be automated.

kumarbabu:
Hi,

I have tried to generate the MSI builder with help of command line arguments.Please find my analysis.

:Control Panel Icon: I am not able to assign an icon to debrief un-installer application in control panel.
:Launch File: Execute the register-files.bat file --> I trying to figure out the launch the register-files.bat file.
:Add registry keys and values to your package, here we can add the register keys, in case if we fails executing the register-files.bat file.

Batch Script:


@echo off
:Create Project:
[...]

PR #1058, opened 2015-01-21
Improved Windows Installer #970

kumarbabu:


IanMayo:
It's a great start.

When I look at the file, I can see that it lists every file.

Isn't is possible to just specify the top-level folders?

IanMayo:
Drop, to be handled separaetly.

PR #1075, opened 2015-01-27
Issue984 naturalearth

pecko:


IanMayo:
@pecko - we have a problem ;-)

I can't quite see how this occurred in this branch, but we've got a problem zooming in.  We can't zoom in closer than an area that's about 100,000 yards across.

If I comment out line 211 of ZoomIn.java the problem appears to go away, but I'm keen to understand the underlying problem.

Hmm, but even if I do comment out line 211, the BR-TL zoom out doesn't work properly in some cases.  I don't know if you know about BR-TL zoom out:  when in Zoom In mode, if you drag the cursor from BR to TL (instead of TL to BR), the plot resizes so that the current view fits into the selected area.   So, if we have *sample.pdf* open, and we do fit to window, the data will fill the screen.  Now if we drag a BR to TL selection area covering the bottom right-hand quarter of the screen, the map will zoom out so that the data now all fits into that quadrant.

Hopefully it's possible to either recognise where these changes happened (from memory), or step back through the commit log to find the breaking commit(s).

org.mwc.cmap.NaturalEarth/src/org/mwc/cmap/naturalearth/wrapper/NELayer.java
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+
+	@Override
+	public void setMap(MapContent map)
+	{
+		// store the map object.  
+		_myMap = map;
+		DirectLayer background = null;
+		// ok, now sort out which dataset we're looking at
```
IanMayo: the following getVisible() causes troubles.  setMap() only gets called once, when NE is loaded.

If the user has NE in his plot, but saves the plot with NE visible, this block of logic doesn't get processed - and it's not possible to switch NE on.

So, IMHO we should perform this NE loading functionality - even if NE isn't visible.

Ok, fixed in 511cff75af017b856346daed492cdf9625e63cc5

Issue #1089, opened 2015-02-07
XY Plots don't show in OSX


IanMayo:
A couple of days ago I found out that XY Plots don't open under OSX.  An error message is displayed instead.  When I googled the error message, it appeared to indicate that I needed an extra property line in the startup arguments.

![screen shot 2015-02-07 at 10 04 49](https://cloud.githubusercontent.com/assets/1108513/6091703/d4b58b7c-aeb0-11e4-9d21-4410213110fe.png)

pecko:
The PR: https://github.com/debrief/debrief/pull/1090

@IanMayo Could you, please, test. I can't test because I haven't some Mac. It should be work with newer JDK 7.

IanMayo:
No, the experiments don't work.  Googling suggests always switching to the newest version.  

But, that hasn't worked either.

I see that JFreeChart does include a SWT.jar file.  I don't think we use that, I think we use the earlier SWT_AWT utility.

Would you mind building a version that uses the JFreechart SWT jar?  Maybe that will fix the osx issue.

pecko:
> I see that JFreeChart does include a SWT.jar file. I don't think we use that, I think we use the earlier 
> SWT_AWT utility.

We use the swt plugin from Eclipse 3.8 distribution (our target platform). I can't check jfreechart because sourceforge is currently down

I can't test the issue because I don't have any Mac.
You can try to run mentioned swt snippet (or some swt_awt snippet) with our swt plugin as well as with swt.jar from jfreechart.

IanMayo:
Hmm, it does look like the only achievable fix is to create an OSX build in a newer version of Eclipse, since the plots "just work" in Luna.


@pecko - Is it very difficult to change the build so that the OSX version of the app is build on Luna? 

Or, can you suggest an alternative strategy to provide an OSX Luna build?

pecko:
> Is it very difficult to change the build so that the OSX version of the app is build on Luna?

We should create a separate target platform for Luna. It isn't to hard - we already had Juno target. This would be good to test future builds with E4.

> Or, can you suggest an alternative strategy to provide an OSX Luna build?

We could check what is changed in SWT on Luna. The main issue is that I don't have some Mac to test.

IanMayo:
Separate target platform - would this target platform still get build automatically by Travis?

pecko:
I think that we can set it so it builds with E3 and E4.
[...]

PR #1090, opened 2015-02-07
Issue #1089 - XY Plots don't show in OSX

pecko:


IanMayo:
Just tried in OSX.  The XYPlot View pane now opens, but it's empty.

pecko:
Could you try some SWT snippet, for instance, http://git.eclipse.org/c/platform/eclipse.platform.swt.git/plain/examples/org.eclipse.swt.snippets/src/org/eclipse/swt/snippets/Snippet154.java

IanMayo:
Have conducted SWT Snippets Trial.

In Eclipse LUNA we don't need to include the "special" java lines to run the demo.

Under 3.8.2, the window was blank when it opened, but when I resized it the table of items appeared - though it wouldn't fill the window area - there was a permanent border.

The last thing I can think of is to trial switching to using charts based on the jfreechart-swt.jar component.

Could you try that?

IanMayo:
This PR didn't fix the OSX Charting issue - so we'll have to ditch it.

PR #1127, opened 2015-03-16
1118 ais text files

IanMayo:


IanMayo:
@pecko  - this seems finished to me.  Are you aware of anything else we need to do?

pecko:
>  this seems finished to me. Are you aware of anything else we need to do?

Agree. Maybe we can enhance algorithm to recognize AIS format. Currently, we just check if the first line contains "!AIVDM". We have to care that this algorithm is fast. Hoping current state is fine for now.

Issue #1163, opened 2015-04-02
Strange UNDO problem on opening plot


IanMayo:
It appears that UNDO only works after the cursor has left, then returned to the plot.

Steps to reproduce:
1. Start Debrief
2. Open boat1.rep  (select Over the Ground if asked)
3. Right-click on track, and go into "Nelson" drop-down menu, and select "Color" / "Green".
4. Click on Undo menu, and select "Undo color for multiple items"
5. The operation doesn't undo.
6.  Select the *Navigator* view, then select the plot (*boat1.rep*) again
7. Click on Undo menu, and select "Undo color for multiple items"
8. Undo will have worked this time.
9. Undo will continue to work.  The problem will only re-occur when a fresh file is opened in a new plot.

Note that after step 4., the event isn't removed from the Undo Buffer - so something is *rejecting* the undo operation.

pecko:
I can't reproduce the issue.
See http://www.snpe.rs/issue1163.ogv.
How I can reproduce the issue described at https://github.com/debrief/debrief/issues/1156?

pecko:
> 4. Click on Undo menu, and select "Undo color for multiple items"

I suppose you mean "Click on Edit menu, ..."

pecko:
There is a problem when opening two (or more) plots. I have tried boat1.rep and boat2.rep and Undo/Redo operation are sometime done in wrong editor (non active editor). I'm not sure if this issue is related.

pecko:
Debrief uses one undo context for all plot editors (defined statically in CorePlugin). This is wrong and we will have different issues when using more plots or open/close some editor.
I will try to fix it, but, please, let me know how to reproduce the issue from #1156.

IanMayo:
Hi there,
tomorrow morning I'll switch to another PC and produce another test case.  But, on my linux PC the steps above worked every time.

IanMayo:
Aah, I can't reproduce it on my Mac.  Undo works correctly on there.

I'm certain the #1156 is the same error as this - the undo event gets ignored.  This issue was generated as an easier-reproduce version of the same issue.  It's unfortunate that the problem is hard to reproduce.  

I guess our best way ahead is to first fix the "single undo history", then for me to see if that resolves the issue.

pecko:
OK.
There is certainly issue when editing more files or open/close one file.

Test case:
- open a plot file
- make some editing so there are a undo/redo actions
- open second file
[...]

PR #1171, opened 2015-04-08
Issue #1163 - Strange UNDO problem on opening plot

pecko:


IanMayo:
Quick piece of feedback - making edits from the Outline View (such as changing a track color) aren't being added to the undo history.

On first impressions, the Undo problem doesn't seem fixed, but I'll investigate more to see if I can see a pattern.

pecko:
I haven't tested the Outline view. Will check...
The PR separate Undo/Redo for each editor's instance. 
IMO, it was main issue in previous code. For instance, you change color for the NELSON track in boat1.rep and get the same Undo action in boat2.rep that doesn't have the NELSON track. Also you keep Undo/Redo action when opening/closing editors.

IanMayo:
Yes, the new behaviour is certainly more sensible.

If I remember correctly, the previous behaviour was implemented so that "Undo" would work when a view (not the editor) had focus.

Users would make an edit from the Outline view, then not be able to undo it until they gave focus to the editor.  They found this non-intuitive.

pecko:
I will check, but Undo worked when a editor had focus. For instance, you can right-click a track in editor area and change track's color. I have followed your case this way (without the Outline view).

> Users would make an edit from the Outline view, then not be able to undo it until they gave focus to > the editor. They found this non-intuitive.

Agree, but it would be good that the actions in the Outline view be undoable/redoable. We can make a separate undo/redo actions in the Outline view, but I have to check. These actions should be active (doable) when the focus is in the Outline view or corresponding editor.

IanMayo:
> These actions should be active (doable) when the focus is in the Outline view or corresponding editor.

Yes, exactly.  Undo operations should apply to the current editor, even if the editor isn't the currently selected RCP pane.

pecko:
Could you please check the actions in editor, for now. I will check the Outline view.

IanMayo:
Hey - the Two-click undo problem issue is fixed in this branch!

Issue #1137, opened 2015-03-18
Put new Debrief icon into Windows launcher


IanMayo:
The MS Windows launcher .exe is using old Debrief icons.

We should update it to use the new ones.

IIRC, the launcher needs a composite bitmap containing the icon at a range of resolutions.  The various resolutions are available in this repo:
https://github.com/debrief/debrief-icons

kumarbabu:
Hi,
Could you please help me to convert the debrief_icon.ico file, I am not able to find  16X16, 32X32,  48X48, 256X256 debrief_icon files inr https://github.com/debrief/debrief-icons repository.

Regards,
Kumar

IanMayo:
16px: https://raw.githubusercontent.com/debrief/debrief-icons/master/16/icon.png
48px: https://raw.githubusercontent.com/debrief/debrief-icons/master/48/icon.png

No, we don't have 32px or 256px versions.  I'll ask the designer to create them.  Or I guess you *could* try re-sampling other icon sizes:
24px: https://raw.githubusercontent.com/debrief/debrief-icons/master/24/icon.png
533px: https://raw.githubusercontent.com/debrief/debrief-icons/master/others/533x533.png

kumarbabu:
Hi,

I have generated  32X32, 64X64, 256X256 derief png iconswith help of existing png icons, please see the attached generated one.

![debrief_32](https://cloud.githubusercontent.com/assets/835065/6808410/cd451a00-d279-11e4-9cb3-efd35edc9102.png)
![debrief_64](https://cloud.githubusercontent.com/assets/835065/6808411/cd483230-d279-11e4-9a23-48ba66915f2f.png)
![debrief_256](https://cloud.githubusercontent.com/assets/835065/6808440/fdbf5f6a-d279-11e4-93a2-bc5ed4db4e1e.png)

IanMayo:
Ok. Those images look like you *scaled up* a small icon.  The images would look better if you produced them by *scaling down" from the nearest larger icon.

kumarbabu:
Hi,

I have used http://image.online-convert.com/convert-to-png web site to produce the images.

IanMayo:
Sure - but the largest 256px image looks like a scaled up version of 16px image.  If you scaled-down the this 533px image (link below) it would look a lot more crisp.

 https://raw.githubusercontent.com/debrief/debrief-icons/master/others/533x533.png

IanMayo:
[...]

PR #1175, opened 2015-04-09
 Put new Debrief icon into Windows launcher #1137

kumarbabu:


IanMayo:
Quick feedback: the smaller icons are really pixelated.  We have icons specifically drawn at the smaller resolutions - we should use them.

IanMayo:
Ahh, maybe OSX Preview isn't the best viewer.  If the small icons are crisp for you, I'm sure they will be crisp for me.  But, if they aren't crisp for you, then you need to replace them with better ones.

kumarbabu:
Hi, in windows photo viewer, those icons are looking well. One more thing I would like to share. we are manipulating the pixels and bits for those generated 7 icons, the way did for old debrief launcher icon. see below,

debrief ico2

IanMayo:
Check in MS Windows - working fine :-)

Issue #1169, opened 2015-04-08
Use of Nebula chart widget instead of JFreeChart


IanMayo:
We're facing JFReeChart troubles of ranging complexity on some user machines.

Specifically we're using the SWT_AWT bridge to show the AWT JFreeChart control in the SWT environment.

Nebula widgets provides a native SWT graph control:
https://eclipse.org/nebula/widgets/visualization/visualization.php

Step one would be to integrate the nebula graph control into the build (prob a task for Peco), then for Peco or Ian to try replacing one or more JFreeChart controls with the nebula equivalent.

The two graphs at the foot of the Maintain Contributions View are candidates for being tried first.

pecko:
The PR #1177 implements step one - integrate the nebula graph control into the build.

BTW We would probably test pure SWT JFreeChart as follows: http://alvinalexander.com/java/jwarehouse/jfreechart/swt/org/jfree/experimental/chart/swt/demo/SWTTimeSeriesDemo.java.shtml
It doesn't use SWT_AWT bridge.

IanMayo:
Hey - the pure SWT JFreeChart would be better - since it's certainly the minimum API change.

pecko:
OK. Will try to implement that in MaintainContributionView tomorrow.

pecko:
The PR https://github.com/debrief/debrief/pull/1184 uses JFreeChart in MaintainContributionsView without AWT bridge.

I think that it is possible to use this technique in XYPlot, but we have to implement the StepperChartPanel class in SWT.

IanMayo:
Works fine for Ubuntu. Will check for OSX.

IanMayo:
WORKS FINE IN OSX!  That's great news.

Would you mind starting the steps to implement StepperChartPanel?

I'm pretty sure we should use a JFreeChart ValueMarker instead of our custom line drawing. That should make it simpler.

pecko:
OK. will try.

pecko:
I have added new PR: https://github.com/debrief/debrief/pull/1190

I have tested timing (stepping from the Time Controller view) and XVPlotView's actions. Some actions are required adapting (export WMF to file, export bitmap to clipboard).

Please test.

IanMayo:
> Some actions are required adapting (export WMF to file, export bitmap to clipboard).

Do you mean they still need to be fixed?

IanMayo:
No, all seems to work.

[...]

PR #1177, opened 2015-04-12
Issue #1169 - Use of Nebula chart widget instead of JFreeChart

pecko:
Step One - integrate the nebula graph control into the build


IanMayo:
Given the success of our use of the SWT version of JFreeChart, I think we no longer need this nebular chart.  Right?

pecko:
I think so.

IanMayo:
Aah no, let's keep it. 

I'll need to check the SWT JFreeChart with my users who were experiencing the JFreeChart freezing problem (#1169)

pecko:
OK. It's just this PR, that is, adding required nebula plugins.

IanMayo:
Sure.  If the SWT JFreeChart  introduces a problem then we will persevere with this PR.

IanMayo:
IAN - DON'T MERGE THIS, IT IS NO LONGER NECESSARY.  The SWT version of JFreeChart worked just fine.

Issue #1188, opened 2015-04-14
Moving shapes  (dynamic annotations)


IanMayo:
Add an enhancement to debrief to allow the depiction of moving shapes.
The moving shape would be listed as a shape in the objects list, but would have additional time / position data in the same way as a track object.
Three new annotations would be added to achieve this; Circle, rectangle and polygon (;CIRCLE_TRACK, ;RECT_TRACK, ;POLY_TRACK).The shape would take a colour from the symbology attribute assigned in the rep file and potentially also line style and fill options. The shapes name would be specified in quotes between the DTG and position terms of the data line and this would be used to group together multiple shape entries into a single track.

To speed up the process of customising the display the additional symbology parameters suggested in earlier proposal (#1186) should be adopted.

Use cases (all cases could be data from trials / exercises or ODIN model runs):
* Displaying wakes and other signatures
* Growing furthest on circle / target uncertainty areas
* Displaying ping front and echo front
* Moving and expanding op areas

## Example

**Existing circle**
```
;CIRCLE: @@ DD MM SS.S H DDD MM SS.S H YYY XX.XX 
;; symb, centre lat, centre long, radius (yards), label 
```
**New circle track**
```
;;format
;CIRCLE_TRACK: @@@@@ YYMMDD HHMMSS "NAME" DD MM SS.S H DDD MM SS.S H YYY XX.XX 
;; symb, date-time, parent track name, centre lat, centre long, radius (yards), label 
;CIRCLE_TRACK: @@011 151212 143000 "Circle Track A" 54 05 12.1 N 4 0 12.4 W 1000 At time 1430 
;CIRCLE_TRACK: @@011 151212 143100 "Circle Track A" 54 05 14.1 N 4 0 12.1 W 1100 At time 1431 
```

## Implementation
We will introduce a new shape wrapper class, **DynamicShapeLayer**.  This will store a series of shape objects.  The Shape objects do not need to be of the same style: they may be rectangles, circles, or polygons.

The DynamicShapeWrapper will be a top-level layer in Debrief, sitting alongside Tracks and other layers.

### Import
When reading in an instance of CircleTrack, RectangleTrack or PolygonTrack the importer will first test to see if it has an object of that name ("Circle A" above).  If the object doesn't exist then a new DynamicShapeWrapper of that name will be created.

If the object does exist, then the shape will be added to that dynamic shape wrapper.

### Painting - Normal mode

A new interface, DynamicPlottable will be created. This will have a single method:
```
[...]

PR #1197, opened 2015-04-22
Issue #1188 - Moving shapes (dynamic annotations)

pecko:
Step 1 - Importing


org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -785,6 +791,20 @@ public HiResDate readLine(final String theLine) throws java.io.IOException
...
+			DynamicShapeWrapper thisWrapper = (DynamicShapeWrapper) thisObject;
+			String trackName = thisWrapper.getTrackName();
+			DynamicShapeLayer dsl = (DynamicShapeLayer) getLayerFor(trackName);
+			if (dsl == null) 
+			{
+				dsl = new DynamicShapeLayer();
+				dsl.setName(trackName);
+			}
+			addToLayer(thisWrapper, dsl);
+			addLayer(dsl);
```
IanMayo: hmm, if an existing layer has been found (dsl != null), won't line 806 make it get added twice?

I would expect addLayer(dsl) to only  occur inside the if-block.

IanMayo:
@pecko  - this is all fine.  But, I'd like us to complete the implementation before we merge it into the trunk.

pecko:
> @pecko - this is all fine. But, I'd like us to complete the implementation before we merge it into the trunk.

Of course, I will continue on painting and exporting/importing to dpf

PR #1212, opened 2015-05-02
Issue #1169 - Make 'Only draw dots for visible data points' un-selected ...

pecko:


IanMayo:
Hey, I'm really surprised that this is all that was needed for the fix - but it certainly makes it work...

pecko:
The commit https://github.com/pecko/debrief/commit/cb530632c974241858312d529529da8a6675b5de fixes the issue with Bearing/Frequency Residuals views.

Issue #1204, opened 2015-04-28
Improve hover text when dragging track


IanMayo:
When dragging a manual TMA track, we show a hover box that displays the course and speed for the solution.

This box gets increasingly difficult to read over a cluttered plot.

Analysts would like the box to be made easier to read, by making it a solid box, shown above the data. 

Reproduce this by opening **midflow2.dpf**.  Then click on **Drag Track Segment**, then start to drag one of the red track segments (from the end point).

You will see the popup hover box appear.  In Ubuntu/64Bit Java 8 - the text isn't legible at all.

pecko:
I can't reproduce the issue on Ubuntu 64-bit 14.04 (JDK 1.8.0_25 or JDK7). I suppose that you have some specific ubuntu (or Eclipse) theme/font settings. 
Could you, please, create a screenshot (or screencast)?

pecko:
I suppose you mean the following popup:

![image](https://cloud.githubusercontent.com/assets/534584/7446465/bb13049e-f1d9-11e4-9aed-db477901c494.png)

pecko:
Try to fix (will create the PR tomorrow if it's correct):

http://www.snpe.rs/issue1204a.png

pecko:
The PR: https://github.com/debrief/debrief/pull/1214 fixes the issue as shown in second screencast.

IanMayo:
http://www.snpe.rs/issue1204.ogv - Yes - it clearly works on your machine.

On my machine I just get a white rectangle:
![draghighlight2](https://cloud.githubusercontent.com/assets/1108513/7470022/50b9e8d6-f311-11e4-9976-7624632136dc.png)

pecko:
It's weird. I'm using the same OS (Ubuntu 14.04 64-bit)
Could you try the PR: https://github.com/debrief/debrief/pull/1214?

IanMayo:
Yes, it's weird.  The image was produced using the PR #1214 

There's another strange behavior: when I measure range and bearing under Linux I don't see the measured text on the plot.  I see the "elastic line" that represents the range/bearing being measured - but not text (just the set of measurements displayed down in the status bar):
![dragrangebearing](https://cloud.githubusercontent.com/assets/1108513/7474486/b1d5610e-f339-11e4-84cf-6f0994e368c6.png)

So, it looks like for "Range and Bearing" and "Drag Track Segment" - the dragging behaviour can display the line, and the shaded rectangle.  But - it doesn't display text.

[...]

PR #1214, opened 2015-05-04
Issue #1204 - Improve hover text when dragging track

pecko:


IanMayo:
@pecko  - it looks like I've mangled another of your PRs here.

This fix certainly resolves my Linux issue.  I'll merge it, and see if it fixes the client issue on Monday.

Issue #1160, opened 2015-04-02
Aesthetic color scheme for tracks


IanMayo:
In my Open Source maritime analysis application we use colored lines to show the paths where different vessels have travelled:

![introshot](https://cloud.githubusercontent.com/assets/1108513/6962635/203f7e0c-d92e-11e4-86e4-445088fd5d78.png)

We have 10 standard track colors in Debrief.
![image](https://cloud.githubusercontent.com/assets/1108513/6962394/081b452e-d92c-11e4-8c18-c195e7161f98.png)

The 10 specific colors used are shown in the screenshots at the end of this posting.

These colors are used for both displaying vessel tracks (routes) on the map, and in UI listings.

Vessel route:
![routes](https://cloud.githubusercontent.com/assets/1108513/6962423/4525af5e-d92c-11e4-811e-3bcd9f8a7fdd.png)

Listing:
![narrativelist](https://cloud.githubusercontent.com/assets/1108513/6962431/55b3b6a4-d92c-11e4-9aba-ad5fa15cb468.png)

Unfortunately some of these shades have poor contrast against a white UI backdrop, or a white/light-gray map backdrop.   Within our chosen technologies it is possible to use custom shades as a backdrop for the lighter shades, but not on the map backdrop.

The shades also lack coherency.  Ideally it would be possible to devise a series of color shades that complement each other.  Diagrams exported from the software will typically be used in government or industrial technical reports, so the colors cannot be to "garish".

This two-stage task includes the following:
* design 10 new color shades.  A, B, and C will remain as some shade
 of Blue, Green and Red (resp).  But the other 7 shades may be created as fresh shades/colors.  The 10 shades should remain legible and clear against Black, White, Gray and Blue-Ocean shades (see the sample images posted below).
* for lighter shades, devise complimentary background shades (possibly gray) to be used in UI listings, to retain the legibility of lighter-shade text.

Note 1: Users will frequently copy/paste the chart images into MS Word-style documents, where the image will always have a white backdrop.
Note 2: The Debrief application has recently had a UI makeover, where two particular shades of Blue and Red have been used extensively RGB(00,00,51) and RGB(204,0,51).  If aesthetically possible, I would be interested in the designer investigating use of these specific shades for Blue and Red.

**The deliverable for this task will be 10 foreground and a number of background shades, presented in RGB color format**

Gray backdrop:
[...]

PR #1220, opened 2015-05-06
Issue #1160 - Aesthetic color scheme for tracks

pecko:


IanMayo:
@pecko  - I had to make a change to the shade of red used to make the tests run, presumably to match our new shades.

Does it look right to you?

pecko:
Yes. It's right.

Issue #1225, opened 2015-05-08
Display of range/bearing text in Rng Brg Measurement tool


IanMayo:
This worked, then it didnt. 

It isn't working right now (64 bit JRE7, Ubuntu 14.04)

When I use the measure range/bearing tool, I don't see a displayed text.

I guess we need a fix similar to the one we used for #1204

pecko:
What is your background? Could you save you current plot and attach them here?

IanMayo:
It fails with light or dark backgrounds. 

This is with **sample.dpf**:
![screenshot from 2015-05-08 15 44 00](https://cloud.githubusercontent.com/assets/1108513/7538624/3a0d0890-f599-11e4-8189-dfa422a2c0ac.png)

pecko:
You mean: It fails with sample.dpf from the master branch. Right?

Do you agree with the following solution: create text always as black on white background regardless plot's background. It would enable us to avoid XOR mode.

IanMayo:
Yes - it fails with sample.dpf from the master branch.  It isn't visible in any plot I open.

Yes - your solution is satisfactory.  But, does it trigger a whole plot repaint?

(using XOR was a way of avoiding plot repaints during drag operations).

pecko:
I'm not sure. Have to try. I would like to reproduce the issue. Have to some idea when/why it happens?

IanMayo:
I get it every time I try to measure range/bearing - for every plot.

pecko:
... but yesterday it worked fine. Right?

IanMayo:
Yes - during our screen sharing. But I did have a suspicion that we rolled back one or more commits from that version.

pecko:
You reproduce the issue with current master?

IanMayo:
Yes

pecko:
Trying ...

pecko:
what eclipse version you are using?

IanMayo:
3.8.2

pecko:
I can't reproduce. I'm using Debrief master, JDK 1.8.0_45, Eclipse 3.8.2 (from target platform), Ubuntu 14.04 64-bit...
Could you try to reproduce it with Debrief distribution: https://github.com/debrief/debrief/releases/download/Stable_20150508_3.0.323/DebriefNG-Linux64Bit.tar.gz (I will also try it).

IanMayo:
Sure. Will not be able to try it for a few hours - sorry.

pecko:
[...]

PR #1227, opened 2015-05-08
Issue #1225 - Display of range/bearing text in Rng Brg Measurement tool

pecko:


pecko:
Closing... You want https://github.com/debrief/debrief/pull/1227 for all platforms.

Issue #1218, opened 2015-05-06
Possible zoom problem with new XY Charts


IanMayo:
I've got a problem with the new XY charts.  

* open **sample.dpf**
* right-click on **Track:Collingwood** and select *XY Plot**
* select **Course** as the attribute to view
* when it opens, make the xy plot about 650 pixels wide (chart 1 below)
* zoom in on the "peak" in the course at around 09:30
* the zoomed in area is shown
* do a reverse drag (upwards & to the left) to rescale the axes
* make the chart 50% wider (about 900 pixels)  (chart 2 below)
* repeat the zoom on the peak 
* an empty area is shown. The time period is actually to the left of the desired area  (chart 3 below).

I can't reproduce this issue with the JFreeChart originals.  It looks like an SWT resize event isn't being propagated to the underlying chart - so the chart things a different area has been zoomed in on.

### Chart 1
![chart1](https://cloud.githubusercontent.com/assets/1108513/7491118/3183aec0-f3de-11e4-80c2-d3c1c6171a22.png)

### Chart 2
![chart2](https://cloud.githubusercontent.com/assets/1108513/7491120/3187480a-f3de-11e4-8d02-e62c93ea6e1d.png)

### Chart 3
![chart3](https://cloud.githubusercontent.com/assets/1108513/7491119/3186a166-f3de-11e4-9439-dd2032f7bd50.png)

pecko:
@IanMayo Could you create screencast?

IanMayo:
Video made, and uploaded here:
https://www.youtube.com/watch?v=zNinfubFzdA

(though it may take a minute or two more for YouTube to finish processing it)

In the first part of the video you will see me select an area to zoom in on.  The area displayed will always be to the left (before) the selected area.

In the second part of the video you will see that when I click on a data-point, the cross-hairs actually jump to a point further left (before) the one I clicked upon.

pecko:
Sorry, I don't understand the issue.

You call "zoom in" three times. They are correct. Right?
Your issue is just click and shown cross-hairs. 
I can reproduce/see the issue with cross-hairs, but can't reproduce some issue with zooming.

In any case, it is a JFreeChart bug and we have to fix it in the JFreeChart code.

IanMayo:
Yes, I call "zoom in" 3 times. But, it zooms in on the wrong location each time.  The time period displayed in the zoom doesn't match the area I dragged with the mouse.

I'm pretty sure this is related to the cross-hair clicks issue.  

[...]

PR #1236, opened 2015-05-14
Issue #1218 - Possible zoom problem with new XY Charts

pecko:


IanMayo:
@pecko  - I've made some changes, to increase the size of the graph before it starts to scale the internal image.  Oh, and some XY Plot formatting.

Issue #1161, opened 2015-04-02
English language property names


IanMayo:
For years and years we've presented the Java Getter/Setter method names to users, both in the properties window (below), and in right-click context menus.

![propertynames](https://cloud.githubusercontent.com/assets/1108513/6964031/432b4e6e-d93d-11e4-9602-34365f1e7f52.png)

It's probably time to switch to properties that use proper english language & spaces.  I propose we modify the prop() and expertProp() methods to introduce another String parameter. This String will be presented to the user.

The UI logic will use this String, not the raw property name.

We'll use mixed case, not all capitals - so "LineThickness" above will be represented as "Line thickness".

Eclipse will help us do this, by highlighting all instances using <ctrl-G>.  There's a chance that the methods have already been overloaded with extra String parameters - so be sure to start with the longest signatures, then work on the shortest signatures.

An example of how the properties are declared is in SensorWrapper.SensorInfo class.  The Editable methods called ultimately create PropertyDescriptor objects. I think the current implementation relies on the PropertyName to determine the read/write methods. So, I believe our code will have to use the longer PropertyDescriptor method that explicitly provides the read/write methods.

IanMayo:
Here's an example of where you will introduce the new english-language name:
SensorWrapper.SensorInfo

It's the PropertyDescriptor object that we're affecting.  We're currently using the Bean name - so that PropertyDescriptor can determine the bean getter/setter methods.

But, we'll use an English name.  So, I expect our "prop(" method will still have to receive the bean name, but it will have to determine the getter/setter methods, and explicitly provide these to the PropertyDescriptor.

OR, we could extend the PropertyDescriptor class to include a new attribute ("Display Name").  Maybe it would be called "NamedPropertyDescriptor".

When we build the UI to edit properties,  we check for NAmedPropertyDescriptor.  If it is a named property descriptor we provide the DisplayName to the user, else we provide the existing (Bean) name.

Here's an example of where we access the bean data to build the UI:
RightClickSupport.generateListEditorFor()

In there you'll see what PropertyDescriptor objects we access.

pecko:
@IanMayo Thanks.
Will try just to solve #1163

IanMayo:
@pecko - you have a queue of issues, so I've given this fairly straightforward one to Kumar.

pecko:
OK. I don't seen that.

kumarbabu:
Hi,

[...]

PR #1251, opened 2015-05-23
Issue #1161 - English language property names

pecko:


IanMayo:
@pecko - I've got a new user that I'd like to send a CD to, including this new update.  Can you give me an estimate on when you'll get chance to do the automatic testing?

IanMayo:
@pecko  - I'm keen to have a greater understanding of the test strategy here.  Why to they need to be JUnit Plugin tests?  They don't seem to need UI - so can't they just be JUnit tests?

IanMayo:
@pecko - a second query.  I've just made a breaking change (extra "L" on the end of "Label", to check that the tests work:

fixlabel

But, no test errors were reported:
editabletests

The tests were run from Maven. But, it looks like they aren't doing the job we require of them.

pecko:
> @pecko - I'm keen to have a greater understanding of the test strategy here. Why to they need to be 
> JUnit Plugin tests? They don't seem to need UI - so can't they just be JUnit tests?

We use JUnit Plugin tests because we have to create a environment to find hierarchy and to instantiate Editable classes (projects, classpath ...). It probably can be done in another way, but this is the simplest.
Tests don' run in UI mode, but SWTCanvasAdapter have to be created in UI mode and this test is handled specially.

pecko:
> a second query. I've just made a breaking change (extra "L" on the end of "Label", to check that the 
> tests work:

I have tested it in PDE (IDE), but not in maven build. Maybe I have changed something later. Namely, just calling getPropertyDescriptors will throw an exception if there is invalid property name.

Will check ...

pecko:
> I've just made a breaking change (extra "L" on the end of "Label", to check that the tests work:

You have changed getGriddablePropertyDescriptors, but not getPropertyDescriptors
The test checked only getPropertyDescriptors. I have added checking for getGriddablePropertyDescriptors 9will commit it shortly)

BTW: 
- we can just discover the class where some property name (or names) are wrong. We can't discover exactly property because Debrief catch the IntrospectionException and hide it (return null mostly).
- there are cca 240 classes which are tested

pecko:
I have updated the PR.

@IanMayo you can repeat your test with the label property.

IanMayo:
Great - working fine.

Issue #1248, opened 2015-05-22
Add boolean option for XY Plot View show legend


IanMayo:
Sometimes the XY Plot View legend isn't necessary.

We should add a property to the XY Plot View properties window that allows a user to switch the graph legend on and off.

If there is just one series displayed, I suggest that the legend is off by default.  But, it should be displayed by default if more than one series is present.

pecko:
I have prepared this fix, but will make a PR when you apply https://github.com/debrief/debrief/pull/1251 because possible conflict.

pecko:
The PR: https://github.com/debrief/debrief/pull/1260

Screenshot:
![image](https://cloud.githubusercontent.com/assets/534584/7991418/42b91c76-0af8-11e5-906b-42f1a426b04d.png)

IanMayo:
Fixed in #1260

PR #1260, opened 2015-06-04
Issue #1248 - Add boolean option for XY Plot View show legend

pecko:


IanMayo:
Nice implementation - I like the decision whether to show the legend.

Issue #1203, opened 2015-04-28
Introduce "Lock to 5 degs" in the Bearing Residuals error plot


IanMayo:
An analyst would like the Bearing Residual Errors to have a fixed x-axis.

This is possible by clearing "Keep plot sized to show all data" button.  But, this doesn't fix the error at the middle.

The analyst would like a "Fix error display to 5 degrees" button.  This will lock the error-plot x-axis to 5 degrees either side.

The icon for this new button will be a 5 degs (using degree symbol) with left-right vector arrows either side.

IanMayo:
[Peco - low priority]

IanMayo:
Introduced, in commit: https://github.com/debrief/debrief/commit/0ebdbc05f7981c027e1ddbe2482ce5cb196f980e

IanMayo:
We've got some confusing UI logic in the current implementation.

We have two buttons that affect the x-axis on the right-hand Bearing Residuals plot:
* clip bearing range to +/- 5 degs
* keep plot sized to show all data

These allow three modes:
1. clipped to 5 degrees
2. resize x-axis to show all data
3. don't resize x-axis, leave it unchanged

It's fiddly & non-intuitive to move between these three modes with the two toggle buttons.

We don't actually need mode 3.  We only need the first two modes.

A better implementation would involve just a single toggle button (the clipped to 5 degs one).

When this button is selected, we would clip the range to 5 degs.  When it isn't selected the bearing (x) axis would resize to show all available data.

You can "exercise" this functionality by loading **mid_flow.dpf**, then switching to **Drag Track Segment** mode.  As you drag the red track you'll see the bearing axis on the right-hand *Error (degs)* plot change.

IanMayo:
I'd to bring this forward too please @pecko , after the GriddableEditor fix is complete.

pecko:
> A better implementation would involve just a single toggle button (the clipped to 5 degs one).

> When this button is selected, we would clip the range to 5 degs. When it isn't selected the bearing (x) 
> axis would resize to show all available data.

The 5 degs (_5degResize) button affects only the _dotPlot plot (I suppose it is right plot). The "keep plot sized to show all data" (_autoResize) buttons affects both plots (_dotPlot and _linePlot).
If we remove the _autoResize button, will the _5degResize button affect both plots?

IanMayo:
Hmm, the line plot will always show all available data, regardless of the button setting.

I think that will work...

pecko:
[...]

PR #1269, opened 2015-06-09
Issue #1203 - Introduce 'Lock to 5 degs' in the Bearing Residuals error plot

pecko:


IanMayo:
Thanks - had a play.

But, it's not quite working for me.  When I click the binoculars the dot plot is trimmed to 5 degrees.  But, when I start to drag a track the plot auto-resizes.

Aah, there was a clash in the logic - the stacked dots clash was settting auto range to true if autoResize was true.   

So, I've refined the logic so that we either auto resize, or trim to 5 degrees.  Let's see what the users think.

PR #1274, opened 2015-06-15
Issue1258 - Time Display element

IanMayo:
Creating a PR - so I can start commenting on individual lines.


org.mwc.cmap.legacy/src/MWC/GUI/ExtendedCanvasType.java
```diff
@@ -64,5 +64,9 @@ public void semiFillRect(final int x, final int y, final int wid,
 	public void setXORMode(boolean mode);
 	
 	public boolean getXORMode();
+
```
IanMayo: @pecko  - we don't have a requirement for rounded rectangles, not for this issue.  

There was a previous issue where we wanted "fancy" UI, but for this issue, plain rectangular blocks are fine.

So, could you remove these functions from this API, and from SWTCanvasAdapter?  

thx.

org.mwc.debrief.core/src/org/mwc/debrief/core/editors/painters/TimeDisplayPainter.java
```diff
@@ -0,0 +1,642 @@
...
+			java.awt.Font.BOLD, 16);
+	
+	private String _format = "ddHHmm";
+
+	private PropertyChangeListener _timeListener = new PropertyChangeListener()
+	{
+		public void propertyChange(final PropertyChangeEvent event)
+		{
+			// right, retrieve the time
+			_DTG = (HiResDate) event.getNewValue();
```
IanMayo: @pecko  - if the user has multiple Time Display objects, won't this cause a screen update for each Time Display object?  So three Time Display objects would cause three screen redraws on each time change...

I'm surprised that our Editable has to listen to time updates, then fire plot updates.  The "highlighters" (normal & snail) already cause a plot redraw when time changes.  Is it possible to "hook into" their processing loop?

org.mwc.cmap.legacy/src/MWC/GUI/ExtendedCanvasType.java
pecko: will do it.

org.mwc.debrief.core/src/org/mwc/debrief/core/editors/painters/TimeDisplayPainter.java
pecko: I have to check

IanMayo:
@pecko - we need some more tidying to the relative time formats.

In relative mode, the time format string is being ignored. I think the list needs to include these options:
* Secs
* Mins
* Hours
* Days
* HH:MM:SS
* yy/mm/dd HH:MM:SS

The first four formats would all be formatted as integers.

org.mwc.debrief.core/schema/debrief_plot.xsd
```diff
@@ -681,6 +704,7 @@
 		<xs:attribute name="LineThickness" default="1" type="xs:integer" />
 		<xs:attribute name="Name" use="required" type="xs:string" />
 		<xs:attribute name="Visible" default="true" type="xs:boolean" />
+		<xs:attribute name="MovingLayer" default="false" type="xs:boolean" />
```
IanMayo: we're not using this attribute **MovingLayer** any more, are we?

org.mwc.cmap.legacy/src/MWC/Utilities/ReaderWriter/XML/LayerHandler.java
```diff
@@ -39,12 +39,17 @@
...
+import MWC.Utilities.ReaderWriter.XML.Features.TimeDisplayPainterHandler;
 import MWC.Utilities.ReaderWriter.XML.Features.VPFCoastlineHandler;
 import MWC.Utilities.ReaderWriter.XML.Features.VPFDatabaseHandler;
 
 public class LayerHandler extends MWCXMLReader implements PlottableExporter
 {
 
+	public static final String NAME = "Name";
+
+	public static final String MOVING_LAYER = "MovingLayer";
```
IanMayo: we don't use this any more

IanMayo:
Thanks for the update @pecko 

Unfortunately, to be able to create a TimeDisplay, our DynamicLayer needs to have its own "Add" method - to over-ride the BaseLayer implementation.  The current implementation won't allow a TimeDisplay to be added.

Surely you tested the new functionality before you committed it?  ;-)

Issue #1277, opened 2015-06-19
Put "Export to WMF" into an Eclipse job


IanMayo:
The "Export to WMF" job takes about 4 seconds on my machine.  I believe that for a very complex plot it can take almost a minute on user machines.

The first problem is that Debrief freezes during this process.  It would be great if we could avoid this.  But, I appreciate that we don't want the plot to change mid-way through the painting operation.

I'd also like to make a request - to put the export operation into an Eclipse job, so we get the progress monitor visible in the status bar, labelled as "Exporting to WMF". 

Finally, one small change.  I've currently got an invalid WMF export location - it's trying to export to "/d3__59_54.wmf" - but cant' because I'm not logged in as admin.  But, Debrief doesn't find out until the end of the export process (after a few seconds).  It would be better if we did this check at the start of the export, not the end.

pecko:
the PR: https://github.com/debrief/debrief/pull/1289

IanMayo:
Have given some feedback.  @pecko  - once we have a method of showing the progress bar without the background app freezing, could you apply the same method to the other WMF export process (export WMF embedded in RTF), please?

pecko:
I will to solve it during day. The action requires UI thread, but we could separate this part.

pecko:
@IanMayo  I have updated the PR with fixes for your suggestions. Please review!

IanMayo:
@pecko  - could you also put the "Export image as WMF embedded in RTF"  (**ExportRTF.java**) operation into a similar progress dialog?

Note, I've pushed a commit that fixes a minor typo. So please *Pull* before you commit.

pecko:
@IanMayo I have added the "Export image as WMF embedded in RTF" operation to a dialog. Your commit saved.

IanMayo:
Yes, working fine.

PR #1290, opened 2015-06-30
Issue #1277 - Put 'Export to WMF' into an Eclipse job

pecko:


org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/actions/ExportWMF.java
```diff
@@ -78,14 +89,49 @@ public static void init(final ToolParent theParent)
...
 
 		if (_theParent == null)
 		{
 			CorePlugin.logError(Status.ERROR, "Tool parent missing for Write Metafile", null);
-			return;
+			return Status.CANCEL_STATUS;
 		}
 
+		monitor.beginTask("Export to WMF", 10);
+		monitor.worked(5);
```
IanMayo: we don't know how long this process will take, so I believe we can use the "unknown length" version of the progress bar.  I think it shows a "rotating pattern" that indicates something is running, but does't give an indication of %age progress.

Here's an example I found:
````
      int totalUnitsOfWork = IProgressMonitor.UNKNOWN;
      monitor.beginTask("Performing read. Please wait...", totalUnitsOfWork);
      performRead(_fileName, monitor); // This only performs the tasks
      monitor.done();
````

File: org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/actions/ExportWMF.java [hunk 2]
```diff
@@ -78,14 +89,49 @@ public static void init(final ToolParent theParent)
...
+					throws InvocationTargetException, InterruptedException
+			{
+				executeInJob(monitor);
+			}
+		};
+		
+		try
+		{
+			ProgressMonitorDialog pmd = new ProgressMonitorDialog(getShell());
+			pmd.run(false, true, runnable);
```
IanMayo: Hmm, when I run "Export to WMF", the dialog shows, then Debrief freezes while the export happens.  Since I'm on ubuntu, Debrief goes dark whilst the app is frozen.

Is there a way of doing this so that the App remains "live"?  I guess this means so that the export happens in the background.

IanMayo:
[request added to issue - about adding progress monitor to another export operation]

IanMayo:
great, thanks.

Issue #1292, opened 2015-07-03
Let Eclipse support Local History for DPF & REP files


IanMayo:
Eclipse provides Local History support for recognised file types (I just played with an XML file, and it worked perfectly):
http://www.aaronwest.net/blog/index.cfm/2007/3/23/Do-You-Know-About-Eclipses-Local-History

But, this isn't supported for DPF files.  Maybe we have to register them. 

@pecko - could you investigate if/how we would add this support for DPF & REP files?

pecko:
The plot editor saved a file out of Eclipse. The PR https://github.com/debrief/debrief/pull/1297 fixes the issue.
It should be tested in network environment, because I only tested local file system (FileEditorInput).

IanMayo:
thanks. got other things on today & tomorrow - will test on Weds morning.

IanMayo:
1]  I've got the History window open, and **sample.dpf**, have right-clicked on a previous version, then selected "Get Contents" - "Do you wish to override contents".

The file gets overwritten, but the plot doesn't update.  I guess we can only update the plot to use the earlier version if we close then re-open the plot, right?

[I'm not asking for any implementation here, just your thoughts on the subject]

2]  With no plot open, when I right-click on an earlier version and select *Open in Debrief* I get this error:
````
!ENTRY org.mwc.cmap.core 4 0 Not expected file type:org.eclipse.team.internal.ui.history.FileRevisionEditorInput@7165b6002015-07-09 06:06:28.119
!MESSAGE data source for PlotEditor not of expected type:org.eclipse.team.internal.ui.history.FileRevisionEditorInput@7165b600
````

But, if I select *Open in text editor* I get the normal XML text contents.  Do we need to add support for the FileRevisionEditorInput type?

pecko:
1) Right.
The Plot editor doesn't implement reloading when their input (file) is changed.

2) Right.
I think we should convert file content to Debrief plot (import it)

IanMayo:
1).  Ok, let's leave that.

2).  Yes, I'd like you to investigate how to handle these files of type FileRevisionEditorInput. Please.

pecko:
Of course.

pecko:
I have fixed both issues.

1) It just close and open plot. I think that it is better to do it otherwise a user could have screen that isn't according to the plot file.
2) implement FileRevisionEditorInput

IanMayo:
Great - working fine.

PR #1297, opened 2015-07-03
Issue #1292 - Let Eclipse support Local History for DPF & REP files

pecko:


org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotEditor.java
```diff
@@ -1505,50 +1505,17 @@ else if (input instanceof FileStoreEditorInput)
...
+
+								InputStream source = null;
+								try
+								{
+									source = new FileInputStream(tmpFile);
+									file.setContents(source, true, true, monitor);
+								}
+								finally
+								{
+									FileUtil.safeClose(source);
```
IanMayo: FileUtil.safeClose has a warning against it:

````
Discouraged access: The method safeClose(Closeable) from the type FileUtil is not accessible due to restriction on required library /home/ian/Apps/eclipse-SDK-3.8.2-linux-gtk-x86_64/plugins/org.eclipse.core.resources_3.8.1.v20121114-124432.jar
````

The class javadoc comment also states that close() should be called & errors handled before calling safeClose().  See the safeClose() javadoc comment here:
http://grepcode.com/file/repo1.maven.org/maven2/com.tqlab.docbook/docbook-highlight/1.0.0/org/eclipse/core/internal/utils/FileUtil.java#254

pecko: This just close a stream without an exception, what we want here. I use such code in a lot places (in Debrief or other app).
The warning is the standard Eclipse warning when using internal Eclipse API. If we won't use the internal API we can replace this code with original code. However we use FileUtil.transferStreams (line 1540) that is slightly more complex. 
So, the code is correct, but question is if you want me to remove internal API. I think that we use it in other places too, so it isn't important here.

IanMayo: That's a good explanation.  In that case, I'm not worried about it.  Thanks.

PR #1322, opened 2015-07-23
Testing new travis disk limit

pecko:
- added linux 32-bit distribution to test disk limit
It will certainly exceed 3GB disk.

Don't apply.


pecko:
It works fine.

Issue #1318, opened 2015-07-17
Drop in performance for dragging tracks since March build


IanMayo:
I was with a user today, showing off the new "show bearing error via scaled circles" capability (#1294).

But, when we tried to drag his track, it was really slow.  We moved back through some previous versions, and found that a build from 16th March worked perfectly ( https://github.com/debrief/debrief/releases/tag/Stable_20150316_3.0.317 )

Just in case we need to reproduce the problem, here are the specifics of the data:  
* 8 hours of ownship data, with positions on the minute
* 980 sensor cuts, spanning 8 hours
* 3 legs of target track, roughly 2 1/2 hours each

The significant changes since then seem to be:
* switch to JFReeChart-SWT  #1169 
* switch from JFreeChart scaling up beyond a certain size to always rendering at native scale
* change to ensure we display the solution's course and speed #1204 

Steps to reproduce:
* load sample_lots_of_sensors.dpf
* under the Osprey track, switch on the Sensor titled **Sensor:sensor:0 (1505 items)**
* right-click on that sensor, and select **Generate TMA solution from all sensor cuts**
* from the wizard, click on **Next**, then **Finish**
* a new track will appear in the Outline view: **Track:TMA_120532.02**
* make this track the Secondary track (select it, then click on the "2" in the Outline toolbar)
* switch to **Drag track segment** view
* start to drag the new red track, you'll see it moves really, really slowly.

Hopefully you'll be able to reproduce the above performance problem, and identify that it doesn't occur in the March builds.

pecko:
The PR: https://github.com/debrief/debrief/pull/1323

I have used "Sensor 3:94 (5454 items)", have created a track and dragging the segment is pretty slow.
IMO, main reason is the BearingResidualsView.getValueAt(HiResDate) method that searching time using loop for all positions in time series.

Enhancements in the PR:

- using TimeSeries.getIndex(RegularTimePeriod) that optimizes above mentioned searching (I think that it is main enhancement)
- avoid creating Double object and Date object as well as unnecessary Double->double>int conversions (We call this method very often when dragging, so creating object can be significant)

I have added another enhancement (not related to speed):
- when getValueAt doesn't find any value it return 0. It is used to paint and Debrief paints nothing. I have changed so Debrief uses standard height=4 for this case.

[...]

PR #1323, opened 2015-07-25
Issue #1318 - Drop in performance for dragging tracks since March build

pecko:


org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/views/BaseStackedDotsView.java
```diff
@@ -678,25 +683,68 @@ public void run()
...
+	}
+
+	private void delay()
+	{
+		if (updateDotsJob == null || !updateDotsJob.isWorking())
+		{
+			return;
+		}
+		final Display display = Display.getCurrent();
+		long waitTimeMillis = 5 * 60 * 1000;
```
IanMayo: @pecko - doesn't this mean 5 minutes?    Surely that's too long for a timeout.

IanMayo:
Have removed the lazy screen update fix - it was making small/simple dragged tracks slower, with intermittent updates.

Issue #516, opened 2014-05-19
Composite Tracks need to be added to Debrief Schema


pecko:
*[Migrated from Assembla ticket [518](https://www.assembla.com/spaces/Debrief/tickets/518) - reported by @IanMayo on 2013-11-11 10:03:05]*

The Debrief schema ( *debrief_plot.xsd* ) doesn't include the new composite tracks, as exported by *CompositeTrackHandler.java*.  It should.

pecko:
I have fixed this old issue. The PR: https://github.com/debrief/debrief/pull/1325

pecko:
I have added your suggestions. The PR has been updated.

The composite_track element can't anymore extend the track element, because the Origin element is set before colour. I have to copy it and add new properties.

Test is also updated.

pecko:
I have changed the latest commit in the following way:

- Now, TrackHandler handles only track's elements. It also handled composite_track's element (planningSegment and closingSegment)
- CompositeTrackHandle handles composite_track's element (planningSegment and closingSegment)

It enables that we set Origin after all track's elements, but before planning and closing segments.
Also it enables that the composite_track type extends the track type, so we don't have to copy that.

@IanMayo Please review.

PR #1325, opened 2015-08-08
Issue #516 - Composite Tracks need to be added to Debrief Schema

pecko:


org.mwc.debrief.core/schema/debrief_plot.xsd
```diff
@@ -286,6 +287,51 @@
...
+		<xs:sequence>
+			<xs:element minOccurs="0" maxOccurs="unbounded" name="fix" type="fix" />
+			<xs:element minOccurs="0" name="WorldDistance" type="unitsValueType" />
+			<xs:element minOccurs="0" name="Speed" type="unitsValueType" />
+			<xs:element minOccurs="0" name="colour" type="colourType" />
+			<xs:element minOccurs="0" name="Duration" type="unitsValueType" />
+		</xs:sequence>
+		<xs:attribute name="LineStyle" type="xs:string" />
+		<xs:attribute name="Name" type="xs:string" />
+		<xs:attribute name="calcModel" type="xs:string" />
```
IanMayo: there is actually a limited set of model types, as recorded here:
https://github.com/debrief/debrief/blob/master/org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlanningLegCalcModelPropertyEditor.java#L90

pecko: Will fix it today later.

IanMayo:
@pecko  - have made minor change to how Composite Tracks are exported, so please remember to *Pull* before you push your changes,
thx.

pecko:
I have done that.

Notice: "export the Origin before the legs, that's the logical order" can affect old plots.

IanMayo:
"Can affect old plots"

The only effect I can think of is that if we save an old file, the order of the elements will change.  But, there won't be any software breakages, will there?

pecko:
No, it won't. The schema validation for old plots will not work. It will not affect current software.

IanMayo:
Duh, yes, of course.

Still, it does seem the right thing to do.

pecko:
Agree. It is minor issue.

Issue #1335, opened 2015-08-26
Dropping cheat sheets


IanMayo:
We're dropping support for cheat sheets.  We'll keep the sources in the repo "just in case". 

We need the following steps to remove them:
* [ ] don't register them in the plugin
* [ ] Swap any reference from the welcome page to a link to one of the two PDF documents.

Anything else @pecko ?

pecko:
You mean that we replace all tutorials links with one link to tutorial pdf? 

BTW I think that you mentioned all required steps.
If we keep source, we will be able to open cheat sheet with entering url manually.

IanMayo:
There will be just two PDF tutorials: a "Debrief Tutorial" and "TMA skills with Debrief".  The links will just go to the PDFs that we put in the installation folder.

pecko:
OK. Will add it today later.

pecko:
The PR: https://github.com/debrief/debrief/pull/1337

If you want to test links, you have to create distribution, because PDF files are created while building.

The following is tutorials page (maybe you want change something):

![image](https://cloud.githubusercontent.com/assets/534584/9529750/5a5f87e8-4cfc-11e5-83f0-1fd530262a0b.png)

IanMayo:
looks great - will review tomorrow AM

PR #1337, opened 2015-08-27
Issue #1335 - Dropping cheat sheets

pecko:


org.mwc.debrief.core/intro_content/debrief_tutorials.xml
```diff
@@ -4,29 +4,15 @@
 		alt-style="css/overview.properties" style="css/intro.css" name="Debrief bits" path="tutorials/@">
 		<group label="Finding your sea legs in DebriefNG" style-id="content-group"
 			id="orgmwcdebriefcore-introLink-group">
-			<link style-id="content-link" label="A Debrief walkthrough"
-				url="http://org.eclipse.ui.intro/showStandby?partId=org.eclipse.platform.cheatsheet&amp;input=org.mwc.debrief.help.started.tutorial"
+			<link style-id="content-link" label="Debrief tutorial"
+				url="http://org.eclipse.ui.intro/runAction?pluginId=org.mwc.debrief.core&#38;class=org.mwc.debrief.core.actions.OpenDebriefTutorialAction&#38;file=DebriefNGTutorial.pdf"
```
IanMayo: @pecko  - do we really need the custom action?  Isn't it possible to just refer to the relative URL of the pdf?  
It looks like the *openBrowser* or *openURL* actions would do the job, as described here:
http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2FintroContentFileSpec.html

pecko: I open pdf with OS's standard command and if it doesn't succeed open url in browser.
It is better way because we only depends on browser's plugins/addons.
We also have to find the location of pdf file (Debrief install location) what is easier in an action (not sure if it is possible to add install location to url).
In any case it is the simple action.

IanMayo: ok - that's fine.

Issue #1341, opened 2015-09-09
Time marker missing from XY Plots


IanMayo:
We can create an XY Plot (such as a graph of the range between two tracks).

When we move forwards and backwards in time (via the Time Controller), a vertical marker should move backwards and forwards through the XY Plot.

This marker is missing from the current version.

pecko:
It works in some older version (when we use swing chart composite). Will explore tomorrow (today later).

pecko:
The problem was redrawing in swt chart composite.
Fixed by the PR: https://github.com/debrief/debrief/pull/1347

IanMayo:
@pecko  - could you explain how the fix works?

The new code is quite different to the previous code:
https://github.com/debrief/debrief/pull/1347/files

pecko:
The old code was my trick  to refresh _plotControl. Unfortunately, it doesn't refresh time stepper :( . The new code is copied from JFreeChart's swt chart composite (not sure exactly where I found it, but I already used it in a mouse event). It refreshes complete chart properly.
I'm not sure if you want me to investigate deeply, that is, to investigate how exactly JFreeChart notification work.

IanMayo:
No, don't worry about investigating further.  Your explanation is fine.

PR #1347, opened 2015-09-12
Issue #1341 - Time marker missing from XY Plots

pecko:


org.mwc.cmap.xyplot/src/org/mwc/cmap/xyplot/views/XYPlotView.java
```diff
@@ -1611,10 +1611,11 @@ public void run()
...
-					_plotControl.setSize(0, 0);
-					_plotControl.getParent().layout(true, true);
-					_plotControl.redraw();
-					_plotControl.update();
+					JFreeChart c = _plotControl.getChart();
+					if (c != null) 
+					{
+						c.setNotify(true);
+					}
 				}
```
IanMayo: Here's Peco's description of this change:
    The old code was my trick to refresh _plotControl. Unfortunately, it doesn't 
    refresh time stepper :( . The new code is copied from JFreeChart's swt 
    chart composite (not sure exactly where I found it, but I already used it in a 
    mouse event). It refreshes complete chart properly.

Issue #1430, opened 2016-01-13
Ian stuck with Eclipse Properties


IanMayo:
I'm pleased with the progress in my issue1417_limpet_arch branch, particularly when it comes to properties.

Once we have a datafile open (such as boat1.rep), we can click on Limpet/Limpet Action and create some data.

Debrief correctly shows the properties for these new Limpet items.

But, the properties aren't editable.  I know we're creating the correct type of property (TextPropertyDescriptor, not PropertyDescriptor).

@pecko  - would you mind tracking what is happening, to find out why the properties aren't editable?

Note: our new property editors are retrieved in EditableWrapper.getPropertyDescriptors

That seems to be working properly.

cheers,
Ian

pecko:
@IanMayo 

I will start it asap. How it's urgent?

IanMayo:
Hmm, I think my preferred order is this fix first, then the #1429, which will allow me to start DIS coding next week.

But, it's not urgent - it's just the first task.

pecko:
OK. My plan is today later or (probably) tomorrow.

IanMayo:
Ok. I've just pushed my 1417 branch, so you have all my updates.

There have been a few limpet commits, but not many.

pecko:
Fine.

IanMayo:
@pecko - thanks for the UI tasks.

I'd appreciate this one next, please - it's quite a pain.

pecko:
@IanMayo You're welcome.

I will start it asap.

pecko:
The issue happens because Debrief doesn't use standard Eclipse property descriptors than its EditorType.

The PR: https://github.com/debrief/debrief/pull/1442 (against issue1417_limpet_arch) branch solve issue so adds default properties in DebrieIProperty and implementing getInfo (returns EditorType) for ItemWrapper. We should implement getInfo for each class we want to have editing properties.

@IanMayo 

Could you please review? After that and fixing #1441 we should rebase the issue1417_limpet_arch (maybe merge). Rebasing will be harder if you add a lot code here.

IanMayo:
Fixed.

PR #1442, opened 2016-01-20
Issue #1430 - Editing Limpet properties

pecko:


IanMayo:
Hey, well done. This seems a really tricky one.

pecko:
Thanks. Agree, it was tricky.

org.debrief.limpet_integration/src/org/debrief/limpet_integration/data/StoreWrapper.java
```diff
@@ -384,7 +390,32 @@ public boolean hasEditor()
...
+          public String getDisplayName()
+          {
+            return getDisplayName();
+          }
+
+          @Override
+          public String getName()
+          {
+            return getName();
+          }
```
IanMayo: hmm, I'm surprised that this doesn't cause a recursive stack overflow - where the methods keep calling themselves.

Just to help my Java skills - can you explain why getName() doesn't keep calling itself?

pecko: It calls the methods from parent class, but you are right. It is confused for review (maybe some Java version too). I will try to made it clearer.

pecko: @IanMayo 

Please see https://github.com/debrief/debrief/issues/1453.

IanMayo: Hmm,
super.getPropertyDescriptors() would call the parent class.

I'm sure getPropertyDescriptors() calls this class.

pecko: You are right. I have removed these methods in #1453. We need ordinal methods from EditorType.

Issue #1448, opened 2016-01-22
DIS graph layout issue


IanMayo:
We have a chart composite in the DIS Listener view.

But the layout is wrong. It doesn't stretch to fill the space available.

Could you fix this?

pecko:
I test this issue on the issue1440_initial_DIS branch. Is it OK?

IanMayo:
Yes.   

I you want to see it update,  you can press "Play" and then "Attach" and you should see points appear on the graph.

pecko:
Where is "Attach"?

IanMayo:
Aah, maybe it's "connect"

pecko:
Do I need set Server Preferences (IP, port ..)?

IanMayo:
If I remember correctly, they now default to the ones for CustonEspduSender.

But, you can edit them, if you need to.

pecko:
I can't set it to see points, but it's maybe isn't important. You want that the graph occupy all space in chart composite. Right?

IanMayo:
Agreed, it isn't important. Yes, I want the chart to fill the whole composite.

pecko:
trying ... chart composite layout  seems weird.

pecko:
It seems that the ChartComposite doesn't work properly for some layouts. I have fixed it by reflection in Debrief. I think that it's stable (JFreeChart team will not change the canvas field).

The PR: https://github.com/debrief/debrief/pull/1455

Screenshot:

![dislayout](https://cloud.githubusercontent.com/assets/534584/12542093/250076d6-c31f-11e5-96f2-14f39f5379c4.png)

pecko:
> Could you refactor the code/layout so that JFreeChart "just works", the way it does in our other uses?

@IanMayo 
I will try. I have checked and JFreeChart doesn't released a new version From  Jul, 2014 (https://sourceforge.net/projects/jfreechart/files/1.%20JFreeChart/1.0.19/), so I'm not sure if they will accept the patch. Maybe, we can use patched version.
I think, this case is different because we have other widgets within dialog/view. Have to test...

IanMayo:
Working fine. Acceptable for now

PR #1455, opened 2016-01-25
Issue #1448 - DIS graph layout issue

pecko:


IanMayo:
Hmm, I'm surprised we've had to make this fix.  I'm sure it was a difficult fix - sorry if it was a challenge.

We use the JFreeChart ChartComposite in the "Maintain Contributions" view,  the XY Plot View, and the "Sensor Fusion" view, and none of these need the workaround.

I would prefer it if we found a way of using ChartComposite that didn't need the workaround to be implemented.

Could you refactor the code/layout so that JFreeChart "just works", the way it does in our other uses?

IanMayo:
@pecko  - I'm going to merge this, so I have a working version for tomorrow's demo.  But, I'd still like you to provide a version that doesn't need the workaround.

Issue #1474, opened 2016-02-08
Include help links in DIS interfaces


IanMayo:
We have two DIS interface, the preferences page, and the view.

Provide link buttons on both of these pages that link to the DIS user guidance, once we've created it: #1473

IanMayo:
Fixed in #1478

PR #1478, opened 2016-02-08
Issue1473 dis guidance

### LINKED ISSUE #1473: Create skeleton for DIS support
We're going to need to produce user guidance for the DIS support.

Come up with some skeleton structure, and include it in the build, so we can start linking to it.

IanMayo:
Fixes #1473


IanMayo:
Fixes #1474

Issue #1594, opened 2016-03-15
Can't change location of Track Name


IanMayo:
Jason can't change the location of the track label.

We'll have to talk this one through.  The track location attribute is now ignored, since we put the label opposite the time marker, and we intelligently locate the time marker.

So, if we continue with this arrangement, then we should either:
* remove the track label attribute 
* respect the track label attribute if start/end time markers are disabled

PR #1600, opened 2016-03-17
track name relative location

IanMayo:
only force track name relative location if analyst has elected to view start/end time labels


IanMayo:
Fixes #1594

PR #1615, opened 2016-03-18
Show collision label on DIS collision PDU

### LINKED ISSUE #1606: Show spatial label for Collisions
We currently just show a narrative entry for DIS collisions.

We should show a time label, at the relevant location.

We may need a DIS generator update before we have a location to extract.

### LINKED ISSUE #1612: May get "-1" as entity id for Collision PDU
If an entity collides with the environment, we will receive a Collision PDU with -1 as id for one of the entities.

When this happens, we should indicate that the collision was with the environment.

IanMayo:
Fixes #1606 
also fixes #1612


IanMayo:
DROP - merged directly into DIS branch

Issue #1721, opened 2016-04-25
Change Save/Save as /Save all icons to match toolbar icon profile 


theanuradha:
Application toolbar has custom save icon, it should reflect on application menu than eclipse default versions:

![saveicons](https://cloud.githubusercontent.com/assets/1108513/14779909/41d9f868-0ad2-11e6-870f-7985a888ad4a.png)

theanuradha:
reassign when icons are available

theanuradha:
@IanMayo I can add to menu without an issue but it duplicate menu items and hide via perspective extention hide both as command id match . looking for better way to handle it. 
![undo-menu](https://cloud.githubusercontent.com/assets/535713/14796116/54366cb8-0b49-11e6-89ef-5e9fdfa5ea46.png)

theanuradha:
We may have same issue when we try to change save images so will find a better solution

theanuradha:
there is one solution i found is use bundle fragment for **org.eclipse.ui ** and update icon paths in it. just try to see is there better solution before try bundle fragment.

IanMayo:
Hmm, somehow we replaced the icon ok for *Save* in the *File* menu.  Can you investigate how that icon is over-ridden?

theanuradha:
Seems app not have custom **save** icon 

![saveicons](https://cloud.githubusercontent.com/assets/1108513/14779909/41d9f868-0ad2-11e6-870f-7985a888ad4a.png)

theanuradha:
@IanMayo  After some research on google with no luck, I dive into code base of eclipse :-). and come up with simple hack to do it I have tested it and works nice we can use same to update save icons as well will create PR so u can review and decide to accept it or not.

IanMayo:
Oh yes - you're right. Good investigating.   Thanks.

The commit is showing more lines than actually changed, probably because of a reformatting.  Could you check that you're using the eclipse formatting settings xml file in the repo's *contribs* folder?

theanuradha:
ah I thought formating setup at project pref. will create changeset with correct formatting.

theanuradha:
@IanMayo just push formatting to branch

theanuradha:
@IanMayo  I think we can close this now as PR got approved

IanMayo:
Closed.

PR #1729, opened 2016-04-26
#1721 added support for dynamically update workbench-window actions

theanuradha:
undo-menu-fix


org.mwc.debrief.core/src/org/mwc/debrief/core/Startup.java
```diff
@@ -110,101 +108,165 @@ public void run()
...
-		{
+			List<IPerspectiveDescriptor> descriptors, String id) {
+		IPerspectiveDescriptor perspectiveDescriptor = registry
+				.findPerspectiveWithId(id);
+		if (perspectiveDescriptor != null) {
 			descriptors.add(perspectiveDescriptor);
 		}
 	}
 
+	private static class ActionAccessSupport {
```
theanuradha: @IanMayo this is the hack allow access to workbench actions

IanMayo:
The commit is showing more lines than actually changed, probably because of a reformatting.  Could you check that you're using the eclipse formatting settings xml file in the repo's *contribs* folder?  

https://github.com/debrief/debrief/blob/master/contribs/eclipse_code_format.xml

Issue #1688, opened 2016-04-18
Show Context info in Properties view


sambireddyt:
Property editor doesn’t show which object properties are shown. I think it should show some context information about current selection.

ex:
![properties_context](https://cloud.githubusercontent.com/assets/1883173/14621607/1bfe1948-0592-11e6-9efc-6e14a42df0d3.png)

theanuradha:
<img width="532" alt="context_1" src="https://cloud.githubusercontent.com/assets/535713/14949469/dc4e6632-1067-11e6-81fd-0854a471765b.png">

theanuradha:
Multi-Select 

<img width="520" alt="context_multi_1" src="https://cloud.githubusercontent.com/assets/535713/14949473/e22841ea-1067-11e6-821b-1f931183fa25.png">

sambireddyt:
is it not possible to use empty space on view toolbar to add title?

theanuradha:
@sambireddyt so far not found way to get to toolbar space. Even Current impl was found after deep dive in to eclipse source how they build Properties View via Adaptors. :-) highly welcome give any directions on possibility to add on action toolbar space

IanMayo:
Fixed in #1737

IanMayo:
@russcrowley - the Properties window now shows the name of the item(s) being edited.    I've added a comment in tutorial 2 suggesting where it could be described:
https://docs.google.com/document/d/16h1-iKQ-oJswhUlA381xbWsaKfsMkYCVEt7rewzQATE/edit#

russcrowley:
Done.

PR #1737, opened 2016-05-02
#1688 Show Context/selection info in Properties view

theanuradha:
context_1

Multi-select 
context_multi_1

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotPropertySheetPage.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotPropertySheetPage.java
@@ -35,116 +35,119 @@ import org.mwc.cmap.core.property_support.EditableWrapper;
 public class PlotPropertySheetPage extends PropertySheetPage
 {
 
-	private PlotEditor _plotEditor;
-	private Composite _base;
-	private Label context;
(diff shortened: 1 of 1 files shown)


IanMayo:
Anu, what if 50 items are selected? What if the track has a really long name? Could you look into these edge cases?

IanMayo:
Oh, it's also possible to view the current highlighter (from the time controller), an XY plot (time-variable), or a file from the navigator.

theanuradha:
@IanMayo  it it long list it will show possible and tooltip will show full list

theanuradha:
@IanMayo time controller & file from the navigator. all use default version without context

theanuradha:
@sambireddyt  Agree on resource & SWTResourceManager, But this case it only create when editor is open and dispose when editor is closed. so editor itself self cache the bold font on PlotPropertySheetPage and unload itself when editor no longer needed.

IanMayo:
@theanuradha - would you mind double-checking the commit I made that limits how many labels are shown: https://github.com/debrief/debrief/pull/1737/commits/8af2c3b48b6403a8d4bd6ffc220395dbc45ed401

Does the logic look ok?

theanuradha:
@IanMayo Yes it looks good! in case you need tooltip to show full list lest me know I can fix it

Issue #1686, opened 2016-04-18
Multi line text editor enhancement


sambireddyt:
Text editor is not intuitive for name/description. Implement 
multi-line editor without scrolling.

![text editor](https://cloud.githubusercontent.com/assets/1883173/14621511/639071bc-0591-11e6-8621-70ff401e8d8b.png)

IanMayo:
A user experienced this confusion today.

If we can't replace the inline editor with a taller one (3 lines tall?), then we should provide a popup editor  with a multi-line input area.


@sambireddyt - Could you describe some options for me?

sambireddyt:
@IanMayo : I think quick window to accept value will  be easier to use. take a look at these options show in picture. last option is as mentioned, new popup dialog editor. I also think all color/font dialog should be popped up where user actually clicks to activate the editor, not in the middle of the screen. so that user doesn't have to move mouse cursor a lot.

![editor_poc](https://cloud.githubusercontent.com/assets/1883173/14696809/114ac3aa-074c-11e6-8893-ee6c4ac54ca7.png)

IanMayo:
Fixed in #1740

PR #1740, opened 2016-05-04
#1686 Multi line text editor

theanuradha:
multi-line-editor


IanMayo:
@theanuradha - are you able to produce an editor that appears in the same space as the property?  Like in Sambi's mockup:
https://github.com/debrief/debrief/issues/1686#issuecomment-212711199

theanuradha:
right now no. But I can make it

theanuradha:
multi-line-editor_inplace

theanuradha:
@IanMayo  changes on in-place mode is in.

theanuradha:
@IanMayo  by mistake seems I have push change that not belong to this branch, I have revert that change. SO it should be fine , if not let me know will recreate with new branch only have correct changes.

IanMayo:
Hi there, we have an untidy situation where the text looks different in view & edit modes.

In view mode, the text is multi-line, as it would appear on the plot:
viewmode

But, in edit mode, line-break chars are inserted:
editmode

Actually I think it would be acceptable to not allow inline editing - to just have the view mode, or click on the [...] button to open the popup editor (similar to how we handle color & font).

Could you make that change?

IanMayo:
Oh, and please note @theanuradha - I've added a commit that makes the editor slightly smaller.

theanuradha:
@IanMayo  Im on it

IanMayo:
Hi there Anu - I've also just pushed a commit that introduces scrollbars.  You'll need to merge your changes with that.

Oh, one last thing: Sambi's original suggestion included the ability for the dialog to resize - could you include this:

resizable dialog

theanuradha:
@IanMayo I added simplified mode. will be complicated to add resize support but I can try.

IanMayo:
Thanks for the simplification - that's great.

>  I added simplified mode. will be complicated to add resize support but I can try.

Ok, let's not add complexity. Let's only do it if we can support resize via SWT.RESIZE flags, or less than 10 extra lines of code.

theanuradha:
@IanMayo SWT.RESIZE only works with Dialog TRIM, custom code will be complex for sure.

IanMayo:
Aah, @theanuradha - there's one remaining UI glitch.

I don't know if this is Linux specific or not, but...  As we switch from view to edit mode, the text moves a few pixels up and to the left.  Can you prevent this?  I guess the edit-mode label needs a few pixels offset.

IanMayo:
> @IanMayo SWT.RESIZE only works with Dialog TRIM, custom code will be complex for sure.

Ok, let's not support resize.

theanuradha:
@IanMayo  can you send screenshot off offset issue ? 
on mac
multi-line-editor_update

IanMayo:
Ok, the "jump" is before the dialog has opened. So, when the "Label" property is selected, the text moves slightly up and to the left.

I've tried to shade in the pixels that are lost in this composite image that includes snippets of the property once selected:

viewmode2

theanuradha:
@IanMayo looks like OS specific, can you see same issue happen on font editor ?

IanMayo:
Yes, I get the slight shift for font editor on Linux.

Ok, it's probably OS-specific let's ignore it.

IanMayo:
Fixes #1686

Issue #1700, opened 2016-04-20
Hide/Reveal menu item state enhancement


sambireddyt:
1. Add Grid/ scale to plot
2. Select grid in outline view
3. right click --> hide item

we don't need two menu items to change visibility state.

When grid is invisible, right click should have "reveal" option. When grid is visible, right click should have "hide" option.

I think visibility column should have check editor to make it visible/invisible as enhancement
![hide_reveal_state](https://cloud.githubusercontent.com/assets/1883173/14659929/4b0d40a6-066e-11e6-87e1-3bdf0c8c64ea.png)

IanMayo:
1. **Popup menu**.  I believe we need both options for when multiple items are selected.

2.  **Checkbox editor in Outline view**.    Good suggestion @sambireddyt  Once we've got your comms sorted, and you working on issues, this would be a good one to start on.

sambireddyt:
@IanMayo : (1)popup menu: i think when multiple items are selected, i think we can be smart enough to figure out to show both options or just one option.  Ex: if all selected items are visible, show only hide and vice-versa. When visible and invisible item are selected, both options can be shown.

IanMayo:
@sambireddyt - yes, please investigate provision of tailored hide/reveal menu options.

You'll see that the "Set primary", "Set secondary" and "Add to secondary" menu items are already context sensitive.  I expect your coding will go in PlotOutlinePage.fillContextMenu()

IanMayo:
Fixed in #1755

IanMayo:
We now have 4-ways of hiding/revealing items:
1. click the the checkbox next to the item
2. click the checkbox in the View toolbar
3. right-click the item and select hide/reveal
4. use the View's context drop-down to select hide/reveal.

This is almost certainly too much duplication.  I'm sure we can drop *4.*    Maybe also *3*. Do you have any opinions on the UX @sambireddyt ?

@russcrowley - once this duplication is resolved we should include use of the new capability in the tutorial - I'll assign it to you when that happens.  Clicking on the checkbox next to the item name in the Outline view control is now the simplest/easiest way of setting object visibility - though we still need at least one of the other methods in order to support bulk hide/reveal.

sambireddyt:
@IanMayo I agree option 4 can be removed. I also think option(3) can also be removed unless users are already familiar with this, i just want to minimize number of mouse clicks  here.

IanMayo:
[...]

PR #1755, opened 2016-05-20
Hide/Reveal menu item state enhancement #1700

sambireddyt:
1. Reveal/Hide actions enablement 
2. Visibility column- inline flipping visibility state in outline view


IanMayo:
@sambireddyt - the implementation is fine.  The build problem hasn't been caused by you, but I'll wait until the build ruins correctly before merging.

theanuradha:
@IanMayo Build is valid now

theanuradha:
@IanMayo I have re-added limpet update site as now no dependency conflicts occur

PR #1787, opened 2016-06-14
change New project Wizard to support location selection and import if…

### LINKED ISSUE #1434: Prevent user creating project in home folder
The IT security policies at a user establishment are happy with Debrief storing preferences in USER.HOME/.debriefNG.

But, they do not want users to store Debrief plots in that folder.

So, when the user runs the "New Project" wizard, the IT staff would like users to be forced to choose "Alternate Location" - if possible.

Alternatively, we need to replace the "New Project" wizard with one that creates the project by default on a pre-assigned drive.

## Update: 10th June 2016

At a visit yesterday we devised a new strategy. Here is the current wizard:
createprojectwizard

We will modify this so that instead of entering a Project Name, the user has to use the System File Browser to an existing (or new) folder.  

The wizard will then check if this folder already contains a project. If it does, it will import the project.  If it doesn't, Debrief will create a project in the specified folder.

After the user has selected the target folder, if it's an existing folder I'd like the UI to be updated to reflect the fact that Debrief is "importing" the Project folder, rather than creating a new one.

theanuradha:
image

**When select existing Project** 
image

Fixes #1434

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/dialogs/CreateProjectDialog.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/dialogs/CreateProjectDialog.java
@@ -69,7 +69,7 @@ public class CreateProjectDialog extends TitleAreaDialog
   private static final String PROJECT_NAME_IS_REQUIRED =
       "The project name is required";
   private static final String PROJECT_IMPORT_FOLDER_INVALID =
-      "The project name is required";
+      "A project with this name is already loaded in Debrief:[%s]. Please select another";
 
   private IOverwriteQuery overwriteQuery = new IOverwriteQuery()
   {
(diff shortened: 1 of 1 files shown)


theanuradha:
@IanMayo  when sitch to import maybe better to update Wizard description as well . 

#Also I found one issue on this approach.
User has two projects on filesystem with same Folder name. So when one is already imported other wont allow to import.

IanMayo:
````
#Also I found one issue on this approach.
User has two projects on filesystem with same Folder name. So when one is already imported other wont allow to import due to it going to import as same name.
````

@theanuradha  - Could you explain this again? I don't follow. 
thx.

File: org.mwc.debrief.core/src/org/mwc/debrief/core/dialogs/CreateProjectDialog.java [hunk 1]
```diff
@@ -58,195 +61,257 @@
...
-	{
-		public String queryOverwrite(String file)
-		{
-			return ALL;
-		}
-	};
+  private static final String PROJECT_NAME_IS_REQUIRED =
+      "The project name is required";
+  private static final String PROJECT_IMPORT_FOLDER_INVALID =
+      "The project name is required";
```
IanMayo: shouldn't this message be different to the one in the line above?

theanuradha:
@IanMayo  Yes

PR #1804, opened 2016-06-20
attempt to fix layout issue on linux

### LINKED ISSUE #1802: Multipath analysis view not showing
There has been some interest in the multi-path view.

I just tried it.  It doesn't show at all on Linux.  The graph doesn't show on MS Windows.

Investigate.

## Steps to reproduce
* Open Debrief
* Open **MultiPath** folder in **sample_data**
* Open file **GapsTestTrack.dpf**
* Open **Multipath analysis** View
* Drag **GapsTestIntervals.csv** file onto **Intervals** placeholder
* Drag the **Depth** slider until the two graphs lie over each other

theanuradha:
@IanMayo  can you try this on linux 

Fixes #1802

### DIFF
--- a/org.mwc.debrief.multipath2/src/org/mwc/debrief/multipath2/MultiPathView.java
+++ b/org.mwc.debrief.multipath2/src/org/mwc/debrief/multipath2/MultiPathView.java
@@ -88,8 +88,7 @@ public class MultiPathView extends ViewPart implements
 	public void createPartControl(final Composite parent)
 	{
 	  parent.setLayout(new FillLayout());
-		_ui = new MultiPathUI(parent, SWT.EMBEDDED);  
-		// SWT.NONE allows some controls to appear - but not graph
+		_ui = new MultiPathUI(parent, SWT.NONE);  
 
 		createPlot(_ui.getChartHolder());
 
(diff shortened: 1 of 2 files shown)


org.mwc.debrief.multipath2/src/org/mwc/debrief/multipath2/MultiPathView.java
```diff
@@ -91,7 +87,9 @@ public MultiPathView()
 	 */
 	public void createPartControl(final Composite parent)
 	{
-		_ui = new MultiPathUI(parent, SWT.EMBEDDED);
+	  parent.setLayout(new FillLayout());
+		_ui = new MultiPathUI(parent, SWT.EMBEDDED);  
+		// SWT.NONE allows some controls to appear - but not graph
```
IanMayo: With this set to SWT.NONE I get this image:
somecontrols

With it set to SWT.EMBEDDED I get a blank RCP View.

SWT.Embedded does seem to be unit specific: http://www.eclipsezone.com/eclipse/forums/t29500.html

theanuradha: @IanMayo SWT.EMBEDDED should not be used here as we use SWT component , maybe before Chart must be used SWING component and used the SWT.EMBEDDED. I miss that it use SWT.EMBEDDED, maybe on mac/win it has no effect while it maybe having issue on linux

IanMayo: I think that we should be putting the Chart control into a ChartComposite rather than a plain Composite.  I'm playing with that now.

IanMayo:
Ok, have switched to ChartComposite:
chartlayout

The chart appears, but there is a layout problem.  Would you mind sorting the layout problem?

theanuradha:
@IanMayo  done now it should be fine.

PR #1821, opened 2016-07-11
added New Project Wizard and mask Project Dialog

theanuradha:
image

#on selection
image

(Please note this will show when **reset** Prespective )
image


IanMayo:
@theanuradha - I don't understand the overlap between this new wizard & the New Project wizard we worked on last month:
https://github.com/debrief/debrief/pull/1787

Shouldn't we use the same code in both instances?

theanuradha:
@IanMayo  we can Drop old Dialog and Always use this Wizard so no code duplicate

IanMayo:
Ok, could you update the PR to include removing the old code?

Have fixed minor typo(s).

theanuradha:
@IanMayo  Please review again I have removed old Dialog and port Wizard to reuse

IanMayo:
Hi there,
I need another minor change, please.

The dialog includes a message explaining why a folder is needed.  But, when the dialog first opens, this message is hidden by an error message:
messagehidden

I'm not sure if/how you can do it, but could you remove the error detection as the dialog first opens? Then only trigger validation when the cursor (for example) leaves the location text box?

theanuradha:
@IanMayo I think I can let me try this

theanuradha:
@IanMayo  done

PR #1834, opened 2016-07-19
removed WMF export as it now added at Limpet level

theanuradha:
This is cleanup after aduption of https://github.com/debrief/limpet/pull/376


theanuradha:
@IanMayo I rerun the build seems all good. this one cleanup WMF changes on debrief and added new freehp to product. please test when u merage it to see build is correctly include all libs

IanMayo:
Yes, have pushed tag, to force rebuild. Will download & test.

PR #1878, opened 2016-09-09
Issue 1871 Ktable to JFace Viewer 

### LINKED ISSUE #1871: Switch Narrative viewer from KTable to JFace table
The narrative viewer was introduced before the JFace Table was introduced, if I remember correctly.

If it's not too complex a task, I'd like us to switch the view to use a JFace table control.

These parts of the UI functionality should be present in the new implementation - please indicate if any aren't possible:
* [ ] double-click on *Source*** or **Type** header to open dialog to filter on that field
* [ ] entries color coded according to the narrative entry color
* [ ] optional word-wrap
* [ ] two-way sync, select row according to current time, set current time when row double-clicked

theanuradha:
Fixes #1871 

image

Pending is word wrap as it is not supported by default looking at workarounds


theanuradha:
@IanMayo I assume we can merge it now 

image

theanuradha:
Will test on Linux as well

IanMayo:
Works great on Linux!

theanuradha:
@IanMayo ok I can move to search option now maybe we can merge to master

IanMayo:
Sure, let me just do a bit more checking.

PR #1887, opened 2016-09-13
introduce narratives in other file formats

### LINKED ISSUE #1886: Import narrative in different formats (docx, pdf)
Occasionally the narrative documents are received in other formats, such as docx or pdf.

We should extend our functionality to support these formats.

Test files are present in `sample_data\other_formats\FCS_narrative.xxx`

**Files added to Branch/PR: https://github.com/debrief/debrief/pull/1887**

I expect we'll need to pull in other POI files.

There isn't a parent object that represents all these formats (pdf has to come in via PDFBox, not Apache POI).

So, the parent processing will collate a list of Strings then we'll pass those to our common importer.

The methods to be finished off are in `ImportNarrativeDocument.java`, `importFromXXX`

IanMayo:
Early creation of PR, to include test files

Fixes #1886


IanMayo:
Actually, I'm going to leave #1886 unfinished, but I want to merge this for release.

PR #1895, opened 2016-09-16
added support to only calculate wrap supported columns only

theanuradha:
@IanMayo  I have improved bit more on calculate so now only calculate **Entry** column only


theanuradha:
@IanMayo  I think this will improve performance significantly  will look on **NatTable**

PR #1927, opened 2016-10-01
refactor to support Filter & Phrases base highlighting

### LINKED ISSUE #1885: Highlighting text in narrative viewer
When the user conducts a search of the narrative viewer, it would be good to highlight the matching phrase.  I guess a pale yellow background would be fairly standard UI.

Also, it would be good to allow a set of "stock" phrases that get highlighted. A default set would be delivered with some phrases/words, but the user would be able to edit this in the Preferences page.

For this initial development, hard-coded values of ***important***, ***urgent*** and ***track*** should be highlighted.  **CHANGE**: let's have it user configurable.  We can commit to this now that we know the highlighting works well.

We will have a preference titled "Highlight phrases" in the preferences group "Narrative viewer".  This will contain a series of comma-separated phrases.  We will highlight those phrases in a background color that cycles through these shades
* 178 180 255
* 147 232 207
* 232 205 167
* 255 192 215

That will leave 238 255 176 (yellow) to be used for the search string:
shades

Oh, and support for RegExp would be a neat additional capability.

Here's an article on formatting parts of text:
http://stackoverflow.com/questions/26173834/highlight-particular-string-in-swt-tree-node

## Note:
The above swt-tree article is irrelevant. We can't use swt-tree, since it can't show different height rows on MS-Windows.

## Note 2:
I've attached a code snippet that *may* cover styling cells.
[demo_style.txt](https://github.com/debrief/debrief/files/475027/demo_style.txt)

## Note 3:
Actually, a good first implementation could be to highlight text that matches the current search phrase, as entered in the existing basic search interface.

## Note 4: 
the view drop-down menu should include an Action to *Edit highlight phrases*. This should open the preferences page.

theanuradha:
Fixes #1885 

image


org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
```diff
@@ -142,6 +169,37 @@ public void setInput(final IRollingNarrativeProvider entryWrapper)
     updateFilters();
   }
 
+  protected String[] getPhrases()
+  {
+    // TODO: read from pref store
+
+    return new String[]
```
theanuradha: @IanMayo this we change to user defined

IanMayo:
@theanuradha - Was the higlighting performance ok with 12k narrative entries?

IanMayo:
I'll have a deeper look at the code on Monday morning - but I like what I see.

Good luck with the preferences page.  I'll edit the issue to include the requirement to be able to open the preferences page directly from the Narrative Viewer.

theanuradha:
@IanMayo also added "Toggle button for narrative search field"

PR #1951, opened 2016-10-07
added Font change support for Narrative view

### LINKED ISSUE #1942: Change font size in narrative viewer
It would be good if the user could choose the font size for the narrative viewer.

I *guess* we could include this in the preferences page.  Or, just via a popup font selection dialog.

Yes, that's better.  A "Font-size" command in the drop-down menu that lets the user choose a new font size.  The font-size should be stored in the Eclipse settings for the next session.

theanuradha:
Fixes #1942

image


IanMayo:
@theanuradha  - I know it may not be easy, but could you display the font-name-text using the font itself?  Like a preview?

IanMayo:
@theanuradha - have fixed merge issue

theanuradha:
image

theanuradha:
@IanMayo  get latest change as well

PR #1958, opened 2016-10-10
don't try to process all time updates, just the last one.

### LINKED ISSUE #1949: Reduce number of time updates for Narrative Viewer
The Narrative Viewer highlights the entry nearest to the current time.

When we drag the time slider, the narrative viewer updates get queued up.  With a large narrative open this results in the viewer continuing to process the queue of updates once the time drag operation has completed.

The viewer update mechanism should handle this by only maintaining a queue of one time update. As a new update is received it replaces any queued up one.

IanMayo:
@theanuradha - would appreciate your feedback on how I've done this.

Fixes #1949

### DIFF
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/app/NViewerView.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/app/NViewerView.java
@@ -111,6 +111,13 @@ public class NViewerView extends ViewPart implements PropertyChangeListener,
    * 
    */
   private Action _followTime;
+  
+
+  /** we don't want to process all new-time events,
+   * only the most recent one.  So, take a note of the
+   * most recent one
+   */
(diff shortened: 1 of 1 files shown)


org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/app/NViewerView.java
```diff
@@ -967,11 +974,12 @@ protected void timeUpdated(final HiResDate dtg)
 
             public void run()
             {
-              if (_pendingTime != null)
+              if (_pendingTime.get() != INVALID_TIME)
               {
                 // quick, capture the time
-                final Long safeTime = _pendingTime;
-                _pendingTime = null;
+                final Long safeTime = _pendingTime.get();
```
theanuradha: @IanMayo move final Long safeTime = _pendingTime.get(); outside if and use it also in if condition

PR #1973, opened 2016-10-12
Narrative performance

### LINKED ISSUE #1928: Narrative viewer performance
@theanuradha  - I think we still have a rendering performance problem:
stack

I added **narr_12k.rep** to **sample.pdf**, then moved the time slider around.  We can see that **17** different time updates were sent.  For each new time we want to highlight the relevant row in the narrative viewer.    But, we see this results in 43076 calls to computeSize() - even when the size of the columns & rows have not changed.

Could you see if we can overcome this problem?

The excessive computeSize() calls don't have an impact for short narratives, but they do on long ones.

IanMayo:
Reduce the number of screen updates we perform when we're adding a new narrative

Fixes #1928

### DIFF
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/app/NViewerView.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/app/NViewerView.java
@@ -245,33 +245,40 @@ public class NViewerView extends ViewPart implements PropertyChangeListener,
        */
       private void updated()
       {
-        // indicate that there is an update pending
-        updatePending.set(true);
-
-        // queue up a screen refresh
-        Display.getDefault().asyncExec(new Runnable()
+        // is there already a pending update?
+        if (updatePending.get())
         {
-          @Override
-          public void run()
+          // hey, we don't need to fire another!
+        }
+        else
+        {
+          // ok, indicate that there is an update pending
+          updatePending.set(true);
+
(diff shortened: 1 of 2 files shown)


IanMayo:
Hi there @theanuradha - could you have a look at this?

I realised the cause of the narrative locking up - each time we added a single new narrative entry a screen update was triggered.  The attached fix reduces this in two ways:

1. only queue up a new asyncExec if we haven't already got one in the queue
2. only update the UI if we still have updates pending

I suspect that 2. isn't necessary given that we've already tested in 1.  

Would you mind having a look at it?

theanuradha:
@IanMayo I think best approach  is use Eclipse job and has 200 milisec delay 
then each new request cancel the job and recreate new job .

IanMayo:
Cool, could you explain the advantage of the Eclipse Job approach?

theanuradha:
@IanMayo we use same concept  in filterGrid as user keep type it will tigger lot of new search queries . so we have 200 milisec dealy and only do refresh after user idle more then 200 milisec

IanMayo:
Ok, I see.  But, each time we do it, we create a new job.  That's ok for typing, since we'll probably only do it 10 times (or as long as the search phrase is).

But, when we load the _narrative_bulk.rep_ we load 12000 lines. This will result in us creating a new Eclipse Job 12000 times.

theanuradha:
@IanMayo  well it cancel the old job before create new one. but I just notice ur change it looks fine!

PR #1974, opened 2016-10-12
1934 nmea import

IanMayo:
Oops, forgot to create this one.


org.mwc.debrief.core/src/org/mwc/debrief/core/dialogs/ImportNMEADialog.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  @Override
+  protected Control createDialogArea(final Composite parent)
+  {
+    final Composite base = (Composite) super.createDialogArea(parent);
+    final Composite composite = new Composite(base, SWT.NONE);
+    composite.setLayoutData(new GridData(GridData.FILL_BOTH
+        | GridData.GRAB_HORIZONTAL | GridData.GRAB_VERTICAL));
+    composite.setLayout(new GridLayout(2, false));
```
theanuradha: @IanMayo  I is little tick in java to get scope inside method. so no need to worry about method name duplicate or use wrong variable

Issue #1965, opened 2016-10-11
Too many screen updates when deleting lots of narratives


IanMayo:
I had about 7k narrative entries loaded.

I selected about 1000 of them, right-clicked, and selected **Delete 1000 items**

Debrief froze for ages. I had to kill it.

We're performing too many updates.  Investigate, and see if we can just fire one update at the end of the operation.

I've started this in the branch 1965_layer_updates in your repo.

Step to reproduce:
1. Open _sample.dpf_
2. Drag in _narrative_large.rep_
3. Expand **Narratives** folder in **Outline** view
4. Select a few hundred entries
5. Right-click, and select **Delete entries**

On my machine, the context menu doesn't appear for the last step.

IanMayo:
@theanuradha - I've switched this to you.  I'd appreciate you working on it before you work on the PDF-libs issue.

IanMayo:
Note: I've just made a start with this - the cost seems to be in generating the images.

IanMayo:
Done, reduced image creation

PR #1978, opened 2016-10-12
1965 layer updates

IanMayo:
To work on #1965

### DIFF
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
@@ -21,6 +21,7 @@ import java.util.regex.Pattern;
 
 import org.eclipse.jface.dialogs.Dialog;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.nebula.jface.gridviewer.GridColumnLayout;
 import org.eclipse.nebula.jface.gridviewer.GridTableViewer;
@@ -51,13 +52,13 @@ public class NarrativeViewer
     GridColumnLayout layout = new GridColumnLayout();
 
     filterGrid =
-        new FilteredGrid(parent, SWT.V_SCROLL |  SWT.MULTI |SWT.VIRTUAL , true)
+        new FilteredGrid(parent, SWT.V_SCROLL | SWT.MULTI | SWT.VIRTUAL, true)
         {
 
           @Override
           protected void updateGridData(String text)
           {
-            
+
             myModel.updateFilters();
             refresh();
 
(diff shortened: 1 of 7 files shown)


IanMayo:
@theanuradha  - I've done some pretty scary optimisation in this PR.

I'd appreciate you both having a look at the assorted changes and having a general play with Debrief to double-check it's ok.

theanuradha:
@IanMayo  im on it

org.mwc.cmap.legacy/src/MWC/TacticalData/NarrativeEntry.java
```diff
@@ -317,16 +334,26 @@ public final String toString()
...
-    result = prime * result + ((_DTG == null) ? 0 : _DTG.hashCode());
-    result = prime * result + ((_entry == null) ? 0 : _entry.hashCode());
-    result = prime * result + ((_track == null) ? 0 : _track.hashCode());
-    result = prime * result + ((_type == null) ? 0 : _type.hashCode());
-    return result;
+	  if(_hashCode == null)
+	  {
+	    final int prime = 31;
+	    int result = 1;
+	    result = prime * result + ((_DTG == null) ? 0 : _DTG.hashCode());
```
theanuradha: so we do not need to cache it _dtg should be simple sting and we delegate  'return _DTG.hashCode()'

theanuradha:
@IanMayo delete works nicely on my mac

org.mwc.cmap.legacy/src/MWC/TacticalData/NarrativeEntry.java
IanMayo: I had "random crashing" back in January, that was eventually diagnosed as incorrect hashcode implementation.  Something to do with the contract for use of a multi-threaded list class.  Aah yes, it was compareTo:
http://stackoverflow.com/a/27216587/92441

We could have more than one entry at the same DTG, so I think we really do need to use all the fields.

theanuradha: @IanMayo I notice property editor not load correct information debugging now

IanMayo: Ooh, I have changed that.   I modified it so there was a single static intstance of NarrativeEntry._myInfo

That was a silly mistake, since _myInfo contains an instance of the object being edited.

Changing it back.

theanuradha: so far all looks fine

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
```diff
@@ -249,7 +250,30 @@ public void setDTG(final HiResDate dtg)
     // ok, try to select this entry
     if (entry != null)
     {
```
IanMayo: @theanuradha - I introduced this extra step to only change the selected item in the viewer if it's different to the one that's currently selected.

Does it look ok?

theanuradha: Yes

PR #1991, opened 2016-10-13
make sure to update table cells when columns resize to recalculate cell heights 

theanuradha:


IanMayo:
@theanuradha - the refresh also needs to be called when the View is resized, not just the column.

PR #1996, opened 2016-10-14
make sure import dialog stay on top

theanuradha:
@IanMayo  Progressbar dialog is coming from Swing **ProgressMonitor** in MWC.Utilities.ReaderWriter.ReaderMonitor we can not change in unless we have to use new impl of progress dialog in swt


IanMayo:
@theanuradha - could you apply this fix to any other SWT dialogs that we open?

theanuradha:
@IanMayo  Sure I will check all

theanuradha:
@IanMayo  I only fount any other place use Dialog is **CoreFreqImportDialog**

theanuradha:
image


image

PR #1997, opened 2016-10-14
fix for limit number of repaint calls to improve performance in gridview

theanuradha:
@IanMayo seems fix has big impact 👍

### DIFF
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
@@ -286,11 +286,19 @@ public class NarrativeViewer
    */
   public void setEntry(final NarrativeEntry entry)
   {
-    // select this item
-    viewer.setSelection(new StructuredSelection(entry));
-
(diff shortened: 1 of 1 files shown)


IanMayo:
@theanuradha - can you explain how the fix works?

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewer.java
```diff
@@ -286,11 +286,19 @@ public void setDTG(final HiResDate dtg)
...
   public void setEntry(final NarrativeEntry entry)
   {
-    // select this item
-    viewer.setSelection(new StructuredSelection(entry));
-
-    // make sure the new item is visible
-    viewer.reveal(entry);
+    try
+    {
+      viewer.getGrid().setRedraw(false);
```
theanuradha: @IanMayo  by doing this it stop all rendering of ui so live animation of scroll not updated so soon after we do all processing we then enable update ui so it just need to update on it view port.

IanMayo: great - thanks

PR #1998, opened 2016-10-16
Issue 1993 update ui fix

theanuradha:
improved rendering to speedup cell update

### DIFF
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
@@ -550,6 +550,7 @@ public class NarrativeViewerModel
     public void setTimeFormatter(final TimeFormatter formatter)
     {
       myTimeFormatter = formatter;
+      formatCache.clear();
     }
   }
 
--- a/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
+++ b/org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
@@ -242,6 +242,14 @@ public class NarrativeViewerModel
     this.store = store;
 
   }
+  
+  void setViewerDataLoadProgress(ViewerDataLoadProgress loadProgress)
+  {
+    if(myColumnVisible!=null)
+    {
+      myColumnVisible.setViewerDataLoadProgress(loadProgress);
+    }
+  }
 
   public int setInput(final IRollingNarrativeProvider entryWrapper)
   {
@@ -379,6 +387,11 @@ public class NarrativeViewerModel
     return true;
   }
 
+  static interface ViewerDataLoadProgress
+  {
+      void next();
+  }
+  
   private  abstract class AbstractTextColumn extends AbstractColumn
   {
    
@@ -462,10 +475,17 @@ public class NarrativeViewerModel
   private static class ColumnVisible extends AbstractColumn
   {
 
+    ViewerDataLoadProgress loadProgress;
     public ColumnVisible(final IPreferenceStore store)
     {
       super(0, "Visible", store);
 
+    }
+
+    public void setViewerDataLoadProgress(ViewerDataLoadProgress loadProgress)
+    {
+      this.loadProgress = loadProgress;
+      
     }
 
     public Object getProperty(final NarrativeEntry entry)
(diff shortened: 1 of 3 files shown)


org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
```diff
@@ -483,7 +528,16 @@ public int getColumnWidth()
 
     public Object getProperty(final NarrativeEntry entry)
     {
-      return myTimeFormatter.format(entry.getDTG());
+      HiResDate dtg = entry.getDTG();
```
theanuradha: @IanMayo  with weakhashmap it use HiResDate object hashcode so if same value on two records it will use same formating as HiResDate override hashcode

theanuradha: opps HiResDate is mutable ?

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
```diff
@@ -332,36 +408,8 @@ public void dispose()
         @Override
         public Font getFont(Object element)
         {
-          String fontStr = store.getString(NarrativeViewerPrefsPage.PreferenceConstants.FONT);
```
theanuradha: @IanMayo  this was bottleneck so I moved it out and handle it better way.

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
theanuradha: @IanMayo I just review HiResDate it use **_micros** in hashcode method and it is imutable so user can not change the object so **HiResDate** is imutable so safe to use in WeakhashMap key

IanMayo:
Not at dev machine, so this is just my guess.  We may have two entries with the same DTG. These two entries may be in different colours, and have different type/source. So, we can't use the same format property for them, IMHO.

theanuradha:
@IanMayo **toStringHiRes** in NViewerView only use micros for fromat 

```
  public synchronized static String toStringHiRes(final HiResDate time,
      final String pattern) throws IllegalArgumentException
  {
    // so, have a look at the data
    final long micros = time.getMicros();
    // long wholeSeconds = micros / 1000000;

    final StringBuffer res = new StringBuffer();

    final java.util.Date theTime = new java.util.Date(micros / 1000);

    // do we already know about a date format?
    if (_myFormatString != null)
    {
      // right, see if it's what we're after
      if (_myFormatString != pattern)
      {
        // nope, it's not what we're after. ditch gash
        _myFormatString = null;
        _myFormat = null;
      }
    }

    // so, we either don't have a format yet, or we did have, and now we
    // want to
    // forget it...
    if (_myFormat == null)
    {
      _myFormatString = pattern;
      _myFormat = new SimpleDateFormat(pattern);
      _myFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
    }

    res.append(_myFormat.format(theTime));

    return res.toString();
  }
```

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer/NarrativeViewerModel.java
```diff
@@ -550,6 +550,7 @@ public Object getProperty(final NarrativeEntry entry)
     public void setTimeFormatter(final TimeFormatter formatter)
     {
       myTimeFormatter = formatter;
+      formatCache.clear();
```
IanMayo: aah yes, well spotted.

IanMayo:
The progress dialog is working ok, thanks.  But, we have a pause of 10s of seconds after it has completed, before the UI un-freezes.

Clearly there are two lengthy processes going on.  I suspect these are to:
1) put the data into the columns
2) format the rows (including row height)

We cover the first one with the progress dialog, but there is no indication about the second one.
viewerdraw

I've drilled down into the calls, and both spend most of their time in **NarrativeViewer.refresh()**.  But, the second seems to take advantage of the cached data.

Hmm, I wonder if we can avoid the second one?

IanMayo:
@IanMayo wants to suspend this PR.  It looks like it isn't possible to efficiently handle such large numbers of narrratives using Nebula Grid.

PR #2008, opened 2016-10-23
added DocX support with poi ooxml libs

theanuradha:
image


theanuradha:
@IanMayo  PDF read also added 

image

IanMayo:
Great - well done for finding the code to strip the text from *.docx* and *.pdf* documents.

PR #2019, opened 2016-11-30
added Custom Boolean Editor, where show Description correctly 

theanuradha:
I have added custom Editor , Default one works as up on selection it change the value so it make description hide (as value change event)  and unexpected edit of the value as just by click on property it change the value. I have added Check box to avoid those issues 
image


IanMayo:
Sorry, I think it will confuse the users.  Do you know how boolean property editors work in the rest of Eclipse?

theanuradha:
@IanMayo  it depends. some plugins has option with Checkbox and Label provider also show check Icon than "Yes,No" text. 

right now issue is current one in master soon as u select on Label of the property it switch the value . as other type editors it enable editor but not change value to user

IanMayo:
Are you able to switch to always showing the checkbox icon?

theanuradha:
@IanMayo yes i can do that

IanMayo:
@theanuradha - there's another outstanding item here - always showing the checkbox icon

theanuradha:
@IanMayo I try to show image but it has many issue with ui inconsistency (can not show OS specific checkbox images ).

SO I have go with Dynamic Drop-Down Option. upon selection it show the drop-down with values  and list box is visible automatically 

image

IanMayo:
OK. It's not as easy to use as the toggle image.  But, if it's the only way for the property description to show, then that's what we'll have to do.

PR #2020, opened 2016-11-30
prototype version of Zone Chart

theanuradha:
image


theanuradha:
- [x] drag the centre of the blue (or red) time period markers to "pan"
- [x] drag the edge of the time zone markers to expand/shrink the period "zoom"
- [x]  cursor change to show the user when the mouse is in the right position to pan/zoom the zone

 @IanMayo

PR #2024, opened 2016-12-19
undo/redo support for ZoneChart

theanuradha:
work complete

### DIFF
--- a/org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/controls/ZoneChart.java
+++ b/org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/controls/ZoneChart.java
@@ -10,7 +10,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.core.commands.operations.AbstractOperation;
+import org.eclipse.core.commands.operations.IOperationHistory;
 import org.eclipse.core.runtime.IAdaptable;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.MouseEvent;
 import org.eclipse.swt.events.SelectionAdapter;
@@ -418,7 +424,7 @@ public class ZoneChart extends Composite
           zones.add(adding);
           fireZoneAdded(adding);
         }
-
+        else
         {
           final XYPlot plot = (XYPlot) chart.getPlot();
           for (final Zone z : dragZones)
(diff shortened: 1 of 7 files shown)


IanMayo:
strategy makes sense.

theanuradha:
@IanMayo Zoom Undo is bit tricky as Jfreechart missing proper getZoom setZoom methods so had to workaround. let me know is current impl of zoom undoable ok or need more improvements.

IanMayo:
I played with undo and redo of adding/removing zones. Then got the following NPE. It looks like the in-memory zones don't match the JFreeChart ones?  To reproduce:
* open sample.dpf
* show bearing residuals view (click on **Drag track segment**)
* click on calculator (**Slice legs**) button of top plot 
* delete a segment
* undo delete operation
* move mouse over plot

The logic for undo/redo of Delete looks ok, but clearly it's failing.

````
!ENTRY org.eclipse.ui 4 0 2016-12-29 10:26:58.084
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException
	at org.mwc.debrief.track_shift.controls.ZoneChart$CustomChartComposite.mouseMove(ZoneChart.java:267)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:212)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1053)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4169)
````

**Comment by theanuradha** [comment]
@IanMayo ah I think I need to convert Slice legs to undoable

**Comment by IanMayo** [comment]
```` @IanMayo ah I think I need to convert Slice legs to undoable ````

Yes, it would be good for "slice legs" to be undoable.  But, that isn't related to this operation.   The problem occurs if I use *Slice legs* or use real data.

**Comment by theanuradha** [comment]
@IanMayo  yes NPE is fixed

**Comment by theanuradha** [comment]
will check on optation labels now

**Comment by IanMayo** [comment]
**Proper undo/redo of zoom operation**

It is possible to get/set the zoom using the data ranges 

````
          // get the the xy plot
          XYPlot plot = CustomChartComposite.super.getChart().getXYPlot();

          // get the range of the range axis
          Range rangeRange = plot.getDataRange(plot.getRangeAxis());
          
          // set the range of the range axis
          plot.getRangeAxis().setRange(rangeRange);
````

I'm sure you can work out the logic.  But, it's something like:
* **execute** store the existing range and selection before executing super.zoom(selection)
* **undo** set the range to the existing range
* **redo** execute super.zoom(saved range)

**Comment by theanuradha** [comment]
@IanMayo thanks for superb tip will tryout

**Comment by theanuradha** [comment]
@IanMayo lables update on actions now 
image

**Comment by theanuradha** [comment]
@IanMayo I manage to support 'proper Zoom history ' with Stack baseimpl works nicely

File: org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/controls/ZoneChart.java [hunk 1]
```diff
@@ -102,7 +110,14 @@ public IStatus undo(IProgressMonitor monitor, IAdaptable info)
             throws ExecutionException
         {
 
+          if (zoomHistory.isEmpty())
```
theanuradha: this shoulds be !zoomHistory.isEmpty() I added that in next commit

IanMayo: ok, just spotted this myself :-)

org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/controls/ZoneChart.java
```diff
@@ -78,6 +78,54 @@ private CustomChartComposite(final Composite parent, final JFreeChart chart)
...
+        @Override
+        public IStatus undo(IProgressMonitor monitor, IAdaptable info)
+            throws ExecutionException
+        {
+
+          CustomChartComposite.super.restoreAutoBounds();
+          zoomHistory.addAll(zoomRedoHistory);
+
+          for (Rectangle h : zoomHistory)
+          {
```
IanMayo: You've got me stuck here.  I can't see why the **undo** needs us to loop through a set of zoom operations.  I also can't see understand the roles of *zoomRedoHistory* and  *zoomHistory* objects.  We seem to create *zoomRedoHistory* using the contents of *zoomHistory*. But, we also fill *zoomHistory* with the contents of *zoomRedoHistory*.

I would think that an easier strategy would be to remember the existing rectangle when we do *execute* before we fit to win. Then we restore that rectangle in an *undo*.   Am I missing something?

theanuradha: Well issue seems on zoom I get rectangle relative to it current zoom state than from fit to window. so I need to work like a macro to correctly redo  multiple steps

IanMayo: Ok, I've had a play. You're storing the Rectangle of the selection (in pixels). But, if the graph/view is resized, those pixels will now point to another region of the chart.

We need to store the data coords of the operation for a Zoom, and for the "Fit to Window". There's a comment in the PR conversation that shows how to get the data coords.  I suggest we could benefit from a utility function that returns the current viewport as a Rectangle2D, oh, and another that sets the viewport to a Rectangle2D.

theanuradha: ah sure will give it a go

org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/controls/ZoneChart.java
```diff
@@ -447,29 +580,91 @@ else if (isResizeStart(z, event.x) || isResizeEnd(z, event.x))
...
-            intervalMarker.setEndValue(resize.end);
-          }
-          fireZoneResized(resize);
-          merge_1 = null;
-          merge_2 = null;
+            @Override
+            public IStatus execute(IProgressMonitor monitor, IAdaptable info)
+                throws ExecutionException
+            {
+
```
IanMayo: @theanuradha - I'm going through the class, doing some tidying.  the following 4 lines are in a block.  The block isn't necessary for the compiler, but I want to understand it before I delete the braces.  Do you remember why the braces are in there?

Cheers,
Ian

theanuradha: @IanMayo ah I use this to enclose variable context so I do not have to always  come up with new variable names

IanMayo: Aah yes, I remember you doing this before :-)  Thanks

Issue #2013, opened 2016-11-04
Show Error Log View is new item added in narrative import


IanMayo:
We're using the Error Log to give feedback if something screwy happens during a narrative import.

But, not all users will know to look at the Error Log for feedback.

So, if an error is logged from the ImportNarrativeDocument class, reveal the Error Log View

IanMayo:
Done some code tidying.

Aah, hold on.  The issue only requires the Errors view to show when we encounter an error from *ImportNarrativeDocument*.   There is lots of assorted logging in Debrief.  We only need the view to be forced to open if it's from ImportNarrativeDocument - else lots of log entries will be shown.  I'll add a flag to indicate if the Errors Log should be shown.

IanMayo:
Ok, API updated.  ImportNarrativeDocument now sends a flag to indicate the LogView should be displayed.  But, we need to amend the method in DebriefToolParent so that we use the UI thread (since getActiveWindow() is currently returning null).

Could you do that @theanuradha ?

theanuradha:
@IanMayo Sure

theanuradha:
@IanMayo done!

IanMayo:
Closed in #2028

PR #2028, opened 2017-01-02
prompt Error Log view to user up on error report.

theanuradha:
#2013

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/DebriefToolParent.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/DebriefToolParent.java
@@ -165,7 +165,8 @@ public class DebriefToolParent implements ToolParent, ProvidesModeSelector
 		    &&  PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage()!=null)
 		try
     {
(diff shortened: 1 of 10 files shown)


org.mwc.cmap.core/src/org/mwc/cmap/core/DebriefToolParent.java
```diff
@@ -160,6 +160,9 @@ public void logError(final int status, final String text, final Exception e)
 	{
 		CorePlugin.logError(status, text, e);
 		//prompt Error Log view to user up on error report.
+		if( PlatformUI.getWorkbench()!=null 
+		    &&  PlatformUI.getWorkbench().getActiveWorkbenchWindow()!=null 
+		    &&  PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage()!=null)
 		try
     {
       PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().showView("org.eclipse.pde.runtime.LogView");
```
IanMayo: Hmm, the Error Log doesn't open for me.  The second "if" test is always failing.  But, I know I have an Active Workbench Window available.  I expect it's because it's being called from a non-UI thread.

PR #2029, opened 2017-01-09
Incorporate new NatTable sources & libs. First step in integrating in…

IanMayo:
Working PR to track progress on integration of NatTable into Debrief.


IanMayo:
Hi there @theanuradha - I've got it building with the sources & libs.  But am stuck on a runtime problem.  I'd appreciate your support.

Eclipse can't find RichTextPainter - even though I can see it's present.   Hopefully you can make some progress on it. Then we can progress more easily.

````
!ENTRY org.eclipse.ui.workbench 4 0 2017-01-09 11:25:07.422
!MESSAGE Unable to create view ID org.mwc.cmap.narrative.views.NarrativeViewer: org/eclipse/nebula/widgets/richtext/RichTextPainter
!STACK 0
java.lang.ClassNotFoundException: org.eclipse.nebula.widgets.richtext.RichTextPainter
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:501)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:421)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:412)
	at org.eclipse.osgi.internal.baseadaptor.DefaultClassLoader.loadClass(DefaultClassLoader.java:107)
	at java.lang.ClassLoader.loadClass(Unknown Source)
	at org.eclipse.nebula.widgets.nattable.extension.nebula.richtext.RichTextCellPainter.<init>(RichTextCellPainter.java:101)
	at org.eclipse.nebula.widgets.nattable.extension.nebula.richtext.RichTextCellPainter.<init>(RichTextCellPainter.java:54)
	at org.mwc.cmap.NarrativeViewer.app.NViewerView.createPartControl(NViewerView.java:487)
````

**Comment by theanuradha** [comment]
Sure u will check now

**Comment by IanMayo** [comment]
Integration tasks:
* [ ]  setupPartListeners() - this can prob be cut/pasted from NViewerView, and then create the relevant fields/listeners
* [ ]   NarrativeViewerPart.setInput(IRollingNarrativeProvider data). Convert IRollingNarrativeProvider to List<MWC.TacticalData.NarrativeEntry> -  change NViewerPart to expect our Legacy narrative entries rather than the new one.  Or, maybe we should wrap our Legacy NarrativeEntry so it looks like NATNarrativeEntry (return String objects for all fields)
* [ ]   NarrativeViewerPart.setDTG(HiResDate date) - to select the entry on or after the specified time
* [ ]   NarrativeViewerPart.refresh() - after we've edited the list of highlight phrases, changed font size, changed time format, etc. The content hasn't changed, but the styling/presentation may have
* [ ]   NarrativeViewerPart.addSelectionListener() - so we can fire time update on new selection
* [ ]   NarrativeViewerPart.setTimeFormatter() - so we can control the format of the time field
* [ ]   NarrativeViewerPart.setSelection(NarrativeEntry) - to select a the specified entry
* [ ]   NarrativeViewerPart.setSerachMode(boolean) - whether to display free text search field box
* [ ]   NarrativeViewerPart.setWrappingEntries(boolean wrap) - in response to user toggle
* [ ]   NarrativeViewerPart.setFocus() - to handle the high level setFocus() method

**Comment by theanuradha** [comment]
- [x] setupPartListeners() - this can prob be cut/pasted from NViewerView, and then create the relevant fields/listeners
- [x] NarrativeViewerPart.setInput(IRollingNarrativeProvider data). Convert IRollingNarrativeProvider to 
- [x] List<MWC.TacticalData.NarrativeEntry> - change NViewerPart to expect our Legacy narrative entries rather than the new one. Or, maybe we should wrap our Legacy NarrativeEntry so it looks like NATNarrativeEntry (return String objects for all fields)
 - [ ] NarrativeViewerPart.setDTG(HiResDate date) - to select the entry on or after the specified time
 - [x] NarrativeViewerPart.refresh() - after we've edited the list of highlight phrases, changed font size, changed time format, etc. The content hasn't changed, but the styling/presentation may have
 - [ ] NarrativeViewerPart.addSelectionListener() - so we can fire time update on new selection
 - [ ] NarrativeViewerPart.setTimeFormatter() - so we can control the format of the time field
 - [ ] NarrativeViewerPart.setSelection(NarrativeEntry) - to select a the specified entry
 - [ ] NarrativeViewerPart.setSerachMode(boolean) - whether to display free text search field box
 - [x] NarrativeViewerPart.setWrappingEntries(boolean wrap) - in response to user toggle
 - [x] NarrativeViewerPart.setFocus() - to handle the high level setFocus() method

File: org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer2/NarrativeEntryConfigLabelAccumulator.java [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+	public void accumulateConfigLabels(LabelStack configLabels, int columnPosition, int rowPosition) {
+		INatEntry entry = dataProvider.getRowObject(rowPosition);
+
+    String uName = entry.getName().toUpperCase();
+    configLabels.addLabel(uName);
+    registerStylesForSource(configRegistry, uName, entry.getColor());
+	}
+
+	// TODO this could be optimized so the styles get only registered once with the known values
+	// this approach is very dynamic as it reacts dynamically on occuring values
```
IanMayo: @fipro78 - you've hinted that this method could be optimised, probably to reduce both processing and object creation.  Would it be better to keep a local list of which **source** objects have already been handled, or to check if the configRegistry already contains the style before we call **registerConfigAttribute** ?

fipro78: You mentioned that a user has the possibility to configure the styling of entries via the source object. I suppose via preferences, is that correct?
So I would suggest that you create a configuration that is able to read the configurations from the preferences and add the configuration to the ConfigRegistry like shown in the accumulator. And of course you need a preferences listener to update the configuration if a user changes the styling at runtime.

IanMayo: Perfect, thanks!

IanMayo: Aah, have come up with cacheing strategy:

````
	private List<String> processedStyles = new ArrayList<String>();

	@Override
	public void accumulateConfigLabels(LabelStack configLabels, 
                                          int columnPosition, int rowPosition) {
		NarrativeEntry entry = dataProvider.getRowObject(rowPosition);
		if (entry != null) {
			final String uSource = entry.getName().toUpperCase();
			configLabels.addLabel(uSource);
			if (!processedStyles.contains(uSource)) {
				processedStyles.add(uSource);
				registerStylesForSource(configRegistry, uSource);
			}
		}
	}
````

This has reduced thousands of *registerConfigAttribute" calls.

fipro78: Caching is also an option of course. But still I think this could be also a one shot if you register all the style information once based on the preferences. :)

IanMayo: Yes - that would be an option if they were stored in Workstation Preferences.  But, the colors aren't determined until the data is loaded, they're format attributes that are stored in user documents.

fipro78:
But wouldn't it be possible to load the user data and determine the color
configuration before creating the table? Well, just an idea. I don't think
it will make much difference.

Just one optimization, I would use a Set and not a List to collect the
already processed source styles. The contains() method scales better with
Set and you don't need the order capability of a List.

Am 12.01.2017 11:38 schrieb "Ian Mayo" <notifications@github.com>:

IanMayo:
````
Just one optimization, I would use a Set and not a List to collect the
already processed source styles. The contains() method scales better with
Set and you don't need the order capability of a List.
````

Let every day be a school day!  That's a great lesson, thanks @fipro78 !

IanMayo:
@theanuradha - one problem: we can't see the *entry* for the selected row.  I guess the font is correctly set to white, but the background isn't set to the blue shade:
selectedrow

IanMayo:
Another problem: check out the following entry - it's in two paragraphs:
image

When we switch off word wrap, two lines are shown. I guess it's one for each paragraph:
image

But, the intended behaviour is that we only see one line, even if there are multiple paragraphs of data.

Hmm, from drilling down, the problem "may" be in the RichTextPainter code. Is that right @fipro78  ?

IanMayo:
@theanuradha - could you check that we're using @fipro78 's recommended row selection strategy?  Feel free to ask if you're unsure about it.

fipro78:
@IanMayo @theanuradha 
The row selection strategy that is used is correct. Although I would not suggest to use the whole object as the row ID. I just recommended to use the row only selection bindings to make the user interaction consistent.

Regarding the word wrapping, well yes, you have HTML tags in your texts. I think it is a <br> tag. And the HTML tags are processed by the RichTextPainter. The word wrapping feature is used to automatically wrap, but it does not avoid HTML tag processing. Actually this was intended as a feature to be able to still add line breaks although no automatic word wrapping is enabled.

theanuradha:
- [x] Use Search Text Box from old Narrative Viewer (includes "erase" button on right-hand side when text present) (remember useNativeSearchField() )

theanuradha:
@IanMayo I cannot get to "background isn't set to the blue shade" can u get new changes and  double check ?

IanMayo:
@theanuradha - aah, found steps to reproduce "white background" issue.  When the new viewer first loads it works ok.  If I switch off word-wrap, then the entry cell for the selected row appears white. If I switch word-wrap back on, then the the entry cell for the selected row still appears white.

image

IanMayo:
@theanuradha - another minor issue related to the search box. The search box in the new viewer is missing the "eraser" icon when a search field has been provided:
image

theanuradha:
@IanMayo  "white background" is fixed with https://github.com/debrief/debrief/pull/2029/commits/e6d7b90ec08eaeb9bb7c40625dcb8391f932d71c

IanMayo:
Aah, new one for @theanuradha - when "hide" the search field, it still reserves the space.

theanuradha:
- [x] Toggle menu options for Show Source and Show Type columns

IanMayo:
*Toggle menu options* - yes, it's working. But, I'd really like the **Entry** column to *steal* the available space:

image

Currently, it doesn't steal the space. It isn't even possible to resize the **Entry** column.

theanuradha:
@IanMayo "available space" issue is fixed now part of https://github.com/debrief/debrief/pull/2045

Issue #2044, opened 2017-01-16
NAT Narrative font size buttons


IanMayo:
The users really like the "increase font size" and "decrease font size" buttons we have in the Stacked Charts View.

I'd like to add these buttons to the toolbar and menu for the new Narrative Viewer

Clicking on increase/decrease will set the stored preference for the narrative viewer font to be 1px larger or smaller.

**increase.png** and **decrease.png** icons are now available in the org.mwc.cmap.core/icons/16px folder: https://github.com/debrief/debrief/pull/2029/commits/6720c5a394d40dd2e03db77b3f8738c7930390f7

theanuradha:
https://github.com/debrief/debrief/pull/2045

IanMayo:
Great implementation, thanks.

PR #2045, opened 2017-01-17
make sure column hide rearrange widths on visble columns  on  nattables

theanuradha:


theanuradha:
@IanMayo  #2044 font buttons done in this PR

PR #2072, opened 2017-01-31
patched version of nattable core to avoid combo-dropdown issues

### LINKED ISSUE #2067: Trouble with drop-down check-box in NatTable (Linux & OSX)
On Linux (and I think osx), the selector drop-down to filter the Source/Type columns doesn't render properly.

image

I'm assigning this to @theanuradha , but if it's a problem that's common across NatTable then it's of more interest to @fipro78

theanuradha:
Fixes #2067 

@IanMayo Works great on MAC will test on Windows 10 64bit issue next see if this patch helps


IanMayo:
@theanuradha - we don't fix this by merging, do we?  I presume our real fix is to use an updated version of NatTable where the fix has been formally included in the master copy. Right?

theanuradha:
@IanMayo  it is safe to merge until official build happen. as we are not sure when will it be

PR #2076, opened 2017-02-05
make sure font change buttons works in linux

theanuradha:
make sure works on Linux


IanMayo:
@theanuradha - can you explain what was wrong with the initial implementation, please?

org.mwc.cmap.NarrativeViewer/src/org/mwc/cmap/NarrativeViewer2/NATViewerView.java
```diff
@@ -832,8 +833,11 @@ public void run()
               final FontData[] readFontData = PreferenceConverter.readFontData(fontStr);
               if (readFontData != null && readFontData.length>0)
               {
-                readFontData[0].height=readFontData[0].height +1;
```
theanuradha: Linux GTK  FontData class has issue on use variable direct due to it has it own cache useing a string . so needt to use setHeight() to make sure cache reload correctly. 

`JFaceResources.getFontRegistry().filterData( readFontData, Display.getCurrent());`
added to double check best possible fontdata on given OS.

IanMayo: Aah, perfect - thanks.

PR #2091, opened 2017-02-13
2086 update zones

### LINKED ISSUE #2086: Zone Charts should update on new plot/data
We improve the Stacked Dots performance by updating the Zone Charts much less frequently.

But, now it doesn't update when we load new data.

The Zone Charts should update (or be cleared) on these events:
* Plot changed
* Secondary track changed

Clearing the zone chart is probably enough.  We get lots of zone chart `refresh()` events, but we only do it if the zone chart is clear.

IanMayo:
Fixes #2086

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GenericData/TimePeriod.java
+++ b/org.mwc.cmap.legacy/src/MWC/GenericData/TimePeriod.java
@@ -163,6 +163,8 @@ public interface TimePeriod extends java.io.Serializable, Cloneable
     // member functions
     ////////////////////////////////////////////////////////////
 
+    
+    
     /**
      * get the start of this time period
      */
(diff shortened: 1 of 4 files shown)


org.mwc.cmap.legacy/src/MWC/GenericData/TimePeriod.java
```diff
@@ -171,6 +173,31 @@ public final HiResDate getStartDTG()
...
+        return true;
+      TimePeriod tp = (TimePeriod) obj;
+      if(tp.getStartDTG().equals(this.getStartDTG()) && tp.getEndDTG().equals(this.getEndDTG()))
+        return true;
+      else
+        return false;
+    }
+
+
+    @Override
```
IanMayo: @theanuradha - I'm not very experienced at *hash codes*.  Does this `hashCode()` function look correct/valid?

p.s. I don't know how/why examples seem to introduce values like **7** and **31** - but I've used them anyway,

theanuradha: looks good

Issue #1781, opened 2016-06-10
NullPointerException after opening SATC solution


IanMayo:
We created an SATC solution.

When I tried to expand it in Outline view, Debrief threw an NPE.

Then, almost whatever we did in Debrief, it threw an NPE.  The Error Log listed the source of the NPE as: AbstractTreeViewer.getSelection()

I guess part of the SATC solution isnt' rendering properly in the Outline view.
## Investigation

I'm unable to reproduce this on my sample data.  On a future visit I'll have to try running several SATC solution (including sample data ones) to try to isolate cause.

IanMayo:
I had this problem again today.  The NPE was apparently on line 2956 of *AbstractTreeViewer.java*.

We need to find a reproducible case.

@theanuradha - could you look out for getting a NullPointerException from AbstractTreeViewer.

I *think* it's arising from the Outline View, but couldn't reproduce it.

theanuradha:
@IanMayo any steps I can try to reproduce issue ? Btw when u get NPW do u have the full stack trace ?

IanMayo:
Hi there, the NPE is an intermittent one.  Next time it occurs in a repeatable way, I'll be sure to catch it.  I had been ignoring it, presuming it wouldn't impact users, but on Tuesday I saw that it did - so it's worth me/us directing some effort towards it.

IanMayo:
Just had the problem, after pressing "Collapse all layers" in the Outline view:
````
!ENTRY org.eclipse.ui 4 0 2017-01-24 11:06:54.502
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException
	at org.eclipse.jface.viewers.AbstractTreeViewer.getSelection(AbstractTreeViewer.java:2956)
	at org.eclipse.jface.viewers.StructuredViewer.handleSelect(StructuredViewer.java:1211)
	at org.eclipse.jface.viewers.StructuredViewer$4.widgetSelected(StructuredViewer.java:1241)
	at org.eclipse.jface.util.OpenStrategy.fireSelectionEvent(OpenStrategy.java:239)
	at org.eclipse.jface.util.OpenStrategy.access$4(OpenStrategy.java:233)
	at org.eclipse.jface.util.OpenStrategy$1.handleEvent(OpenStrategy.java:403)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
[...]

PR #2094, opened 2017-02-13
Workaround for NPE on Outline TreeView collpase

theanuradha:
#1781


IanMayo:
Hmm, I'll merge it, to get rid of the NPE.  But, could you invest some time in understanding the underlying problem? I'd like to get to the bottom of it - to make use it doesn't cause us problems elsewhere in the future.

theanuradha:
@IanMayo I will

PR #2138, opened 2017-03-10
2136 support infill color on icons color warp

### LINKED ISSUE #2136: Sensor Wrapper icon shading unclear
We currently color-code some icons, such as the Sensor icon - shown pink below:

image

With lighter shades it is difficult to recognise what the icon represents.  It would be better to have an icon in standard Debrief blue/red colors, but with a flash of relevant color shown on it.

The sensor icon is actually a two-tone image as shown below (zoomed in):

image

We should trial swapping the red-shade for the sensor color, instead of applying the sensor color to the whole shape.

The processing to apply the color is here:
https://github.com/debrief/debrief/blob/master/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/CoreViewLabelProvider.java#L204

theanuradha:
Fixes #2136 

image

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/CoreViewLabelProvider.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/CoreViewLabelProvider.java
@@ -31,6 +31,8 @@ import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.graphics.ImageData;
+import org.eclipse.swt.graphics.PaletteData;
+import org.eclipse.swt.graphics.RGB;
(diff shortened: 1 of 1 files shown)


IanMayo:
Working fine on Linux:
image


 But,  I suspect I'll have to see what it looks like with some larger datasets.  Will test, when I get chance.

IanMayo:
@theanuradha - can I ask for a change?  Could you switch to replacing Blue, rather than Red in the icons?  Most icons have more blue than red - so switching out the blue will make it easier to see the custom color.

Thanks :-)

IanMayo:
Oh, @theanuradha - it's ok, I've got it. The colors are in the DebriefColors.java.  It should be easy :-)

IanMayo:
Hold on, tried it, and failed :-) I'd appreciate you having a go, please.

theanuradha:
@IanMayo I change it to pick blue shades

PR #2158, opened 2017-03-18
Allow multiple file types on Open/Import action

### LINKED ISSUE #2154: Classic - let app open .dpf files
Debrief Classic expects its data to be stored in a `.xml` file.

Since then we moved to `.dpf` files. They still match the same schema, it's just that we adoped a different suffix.

The file import processing in Debrief Class should be modified so that it loads .dpf or .xml files.

theanuradha:
Fixes #2154 

image


theanuradha:
#2154

PR #2167, opened 2017-03-23
2166 dataset manipulation

### LINKED ISSUE #2166: TA Data - editing and manipulation
As a stepping-stone to track color-coding by array status (#1818), we need to be able to calculate the difference between fore & after headings (or some function of the multi-module dataset).

So we need:
* editable properties for Datasets (esp name)
* operations to add/subtract/multiply/divide (interpolated)
* ability to store new datasets

IanMayo:
Fixes #2166


org.mwc.debrief.limpet_integration/src/org/mwc/debrief/limpet_integration/measured_data/MeasuredDataOperations.java
```diff
@@ -0,0 +1,204 @@
...
+  {
+
+    @Override
+    public TimeSeriesCore calculate(List<TimeSeriesDatasetDouble> items)
+    {
+      long[] times = new long[]
+      {1000, 2000, 3000};
+      double[] values = new double[]
+      {44, 55, 66};
+
```
IanMayo: @theanuradha - I've got stuck on the following line. This plugin can't see **Dataset**.  Dataset is in the **january.jar** that is shared by debrief.legacy

PR #2285, opened 2017-06-01
2282 Show symbol if visible with parent

### LINKED ISSUE #2281: XY Plot - add Show Symbols to toolbar
Switching on/off symbols on an XY Plot is a frequent operation.

But, going into the Properties View is cumbersome.

Add a suitable toggle button to the view toolbar.

### LINKED ISSUE #2282: XY Plot - more options for "Show Symbols"
The options for show symbols are currently Yes/No

For a frequent track, this means there are so many symbols that it starts to look like a big fat line.

So, the aim was to turn Yes/No into Yes/No/Only if visible. 

But, this clashes with the idea to add a Show Symbols toggle on the toolbar. 

So, maybe it's better to change the behaviour of "Yes" to *show symbols if the relevant fix has a symbol".

IanMayo:
Fixes #2282


IanMayo:
Drop this, duplicate.

PR #2338, opened 2017-07-06
After a resample, we don't need to re-generate the points

### LINKED ISSUE #2330: Resample DR track results in tiny, tiny track
If we resample a DR track from the `Positions` item drop-down, it re-samples correctly.

But, if we resample the DR track from the track-level drop-down, it becomes really, really tiny.

I expect it's an issue related to us using microseconds instead of milliseconds.

IanMayo:
Fixes #2330


IanMayo:
Drop this, it doesn't include all fixes.  All fixes present in #2339

PR #2364, opened 2017-07-26
increment versions (dummy change)

IanMayo:
We've got a Limpet build problem.

I'm forcing a dummy re-build of Debrief to try to isolate if it's a shared problem.


IanMayo:
Close. Eclipse servers back up.

PR #2368, opened 2017-07-27
For release

IanMayo:


IanMayo:
Cancel release.  We need hotfix to be able to assign label as primary/secondary #2369

PR #2609, opened 2017-12-12
2601 select error dot

### LINKED ISSUE #2601: New practice for selecting Stacked Dots items
We've introduced the ability to click on a Stacked Dot item, to select it in the Outline View.

We currently allow the user to click on the sensor cut or the TMA fix, and we select the relevant item.

But, we've established a tidier way of doing it.  If the analyst clicks on the line plot, select the TMA fix. If the analyst clicks on the error plot, select the sensor cut.

IanMayo:
Fixed #2601


IanMayo:
Cancel this. It's already in PR #2607

PR #2719, opened 2018-02-15
Eclipse Oxygen.2

IanMayo:
changes to allow build under Eclipse Oxygen.2


IanMayo:
Works ok with Windows too.  But, it needs the SWT_LIB variable to be assigned, pointing to the org.eclipse.swt.windows[or cocoa].jar

PR #2764, opened 2018-02-22
introduce pref for setting defaults

IanMayo:
Second try at sensor transparentcy


IanMayo:
drop, failed to identify issue

PR #2766, opened 2018-02-22
Feature/fourth transparent cuts

IanMayo:


IanMayo:
Hasn't isolated problem. Dropping

Issue #2933, opened 2018-06-26
D1.1 New export wizard


IanMayo:
As an **analyst** I need to **provide missing metadata** in order to **populate data fields with values not currently present in Debrief**.

We're going to be using a Wizard to do this.  This task covers the generation of that wizard.

Extra detail on the issue is present in the #2912 parent epic.

**Update 4/7/18** - we have some outstanding issues on this, including a couple of missing fields.

theanuradha:
@IanMayo https://github.com/debrief/debrief/pull/2941

IanMayo:
Hi there @theanuradha - I demo'd the new feature to the users this morning.  I got the following feedback:
1). They would like free-text boxes for "Semi-Major Axis" and "Semi-Minor Axis" (_I was mistaken about hard-coding them_)
2) The Classification should allow editing, as well as the drop-down values 
3) The Flag field should allow free editing, in addition to the drop-down vlaues
4) The Type box will be free-text, without any checks.

Note: I've tried to reflect the above in the wizard wireframes:
https://docs.google.com/presentation/d/1jNJ8TqeodjLW7oWzj3JLuqCrfwvT0zOYU_AAWSJshLE/edit#slide=id.g3c02f236c2_0_0

theanuradha:
@IanMayo https://github.com/debrief/debrief/pull/2976/commits/260603074410c71036f55e756c271779c15eb2a2

PR #2941, opened 2018-06-27
2912 new export wizard

theanuradha:
https://github.com/debrief/debrief/issues/2933


IanMayo:
Ok, I've been running through the wizard, and have some changes/feedback:

* I've changed provenance and unit - they now come from the data (if possible)
* We need to check the folder exists, and is writeable before we allow the wizard to complete. Users frequently have read-only folders, so we need to handle that with a warning message.
* Could we make the the dialog remember the previous values for all fields?  (but, controls for unit and provenance would use supplied values, if provided).  This persistence should be across runtimes, so would be stored in the Eclipse prefs, I guess. There's an accessor for the debrief prefs in CorePlugin.getPreferenceStore();

PR #2945, opened 2018-06-28
Added video player view to debrief perspective and on top view

### LINKED ISSUE #2940: M19.3 Video Player feedback
I've identified the following feature requests for the video player:

* include the video player (and image folder) into the Debrief Category of Show/View
* allow more than one instance of video player to be displayed
* provide new icons, in Debrief colors (via our artist)

ayeshare:
Fixes #2940

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/CorePlugin.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/CorePlugin.java
@@ -163,6 +163,8 @@ public class CorePlugin extends AbstractUIPlugin implements ClipboardOwner,
 
   private static final String EARTH_MODEL_PROVIDER =
       "org.mwc.cmap.core.EarthModelProvider";
+  
+  public static final String VIDEO_PLAYER_VIEW="org.mwc.cmap.media.views.VideoPlayerView";
 
   /**
    * preference for detault font size
--- a/org.mwc.cmap.media/plugin.xml
+++ b/org.mwc.cmap.media/plugin.xml
@@ -19,7 +19,7 @@
       <view
             name="Video Player"
             icon="icons/player.gif"
-            category="org.mwc.cmap.media"
+            category="org.mwc.debrief.core.DebriefCategory"
             class="org.mwc.cmap.media.views.VideoPlayerView"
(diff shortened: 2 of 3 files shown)


IanMayo:
Hey - good work so far, @ayeshare - good, clean code.

org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotEditor.java
```diff
@@ -1612,8 +1660,16 @@ protected void filesDropped(final String[] fileNames)
     // ok, iterate through the files
     for (int i = 0; i < fileNames.length; i++)
     {
+
```
IanMayo: I'd like to suggest one change here, please @ayeshare 

After this loop, Debrief updates the plot and resizes it to show all data.  This will give the effect that, if the user has zoomed in an area, then drags in a video, then the plot will zoom out to show the full plot.

I more user-friendly logic would be something like:
* if there's a single file, and it's a video file - send it to the video player
else
* run through the rest of this method, as normal: loop through files, load them, resize plot.

ayeshare: Sure, I will make that change. Thanks

org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotEditor.java
```diff
@@ -1643,14 +1645,32 @@ private boolean isVideoFile(final String fileName)
 
   private void openVideoPlayer(final String fileName)
   {
-    // CorePlugin.VIDEO_PLAYER_VIEW
     //#2940 #6
-    //now open a dialog box and get the start time
-    
+    //if we cannot get the start time from filename open the dialog
```
IanMayo: @ayeshare - this is quite subjective, but it is it possible to keep this understanding of PlaentMayo formats inside the media_player plugin?   I guess an alternate showVideoPlayer that omitted the start-time value could try to parse it from the fileName parameter.

Ideally, I'd also like the VideoPlayerStartTimeDialog to remain private within the media_player plugin.

Issue #2946, opened 2018-06-28
M1.2 Create task to build Debrief-Lite


IanMayo:
As a **Maintainer** I wish to **have an up-to-date release of Debrief-Lite** in order that **I can test it myself, or pass to others for testing**.

We're going to start giving copies of Debrief-Lite to people to play around with/review.

It would be useful if it could be quickly built.  (**DebriefApp.java** in org.mwc.debrief.legacy)

We should change our MVN build, so that we include a release of Debrief-Lite.  This file should also be pushed into the releases folder.

If a team member sorts out building Debrief-Lite (DebriefApp.java) then Ian can probably handle the travis bit.

Note: I expect I've just got to add a new entry to the releases block here:
https://github.com/debrief/debrief/blob/develop/.travis.yml#L30

IanMayo:
Hello there @theanuradha - this one's a natural one for you, please.  You're good at build.  Plus, you work on MACOS - so I hope you can get to the bottom of the MACOS run issues.

IanMayo:
(task renamed - to remove specific Maven directive)

PR #3024, opened 2018-07-18
Maven build for debrief-lite

ayeshare:
Supports #2946


IanMayo:
Closing - we're going down Ant route, instead.

PR #3029, opened 2018-07-19
incorporated splash handler

### LINKED ISSUE #2987: D16 Display build date & version number in splash screen
_As an_ **Analyst** _I would like to_ **see the Debrief version as it loads** _so_ **I have a feel for how new/old this version is, and avoid mistakenly opening a legacy version**

We have a Splash screen that is shown as Debrief loads:
image

It would help users to see the build date & version in this screen, particularly when the two Ian's are doing acceptance testing.

The build date & version should _somehow_ be extracted from the `Debriefng.product` file (or other runtime source), and displays as black text at the bottom-left of the Splash screen.

IanMayo:
Fixes #2987

### DIFF
--- a/org.mwc.debrief.core/plugin.xml
+++ b/org.mwc.debrief.core/plugin.xml
@@ -6,6 +6,18 @@
         <category id="org.mwc.debrief.core.DebriefCategory" name="Debrief"/>
     </extension>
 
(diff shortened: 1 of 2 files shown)


IanMayo:
Close. No developer activity.

PR #3048, opened 2018-07-27
Standardize to use self closing tags

jbampton:
A clean up of the XSLT code I once wrote.  

Hoping to finally get my name down as a contributor on GitHub :)


IanMayo:
Sorry John we moved away from the DITA documents a few years ago - since it was such a challenge to include basic things (like images) in them.

This PR breaks the build - so I can't merge it in this form.

Thanks for your support,
Ian

jbampton:
Hi @IanMayo that is fine.  Are you using XSL in other places that might need review ?

IanMayo:
Thanks for your contribution, but this legacy code is no longer being maintained.

PR #3052, opened 2018-07-31
Feature/3004 two way doppler

### LINKED ISSUE #3004: D18.4 Introduce two-way Doppler for freq residuals
Our frequency residuals plot calculates the doppler that should be received at the sensor platform, using the f-nought of the contact.

But, in multi-statics, it's the f-nought of the TX that is of relevance.

So, the Freq Residuals plot needs a combo-box that lists Debrief tracks that have a sensor on board, together with an _single-path_ default entry.

When the user selects something other than _single-path_, Debrief will first calculate the shift from TX to the contact, then the shift from the contact to the RX.

IanMayo:
Fixes #3004


IanMayo:
Partial implementation integrated,  to prevent it diverging from master while we wait for algorithm review.

PR #3056, opened 2018-08-01
Feature/3025 record button pptexport2

### LINKED ISSUE #3025: M3.7.1 UI for export to PPT
This task is to cover the UI elements of M3.7, Export to PPT.  (#2906)

It is to implement the UX that is defined here:  https://xd.adobe.com/view/c59ed2e7-857f-4f71-7f27-6ff23da4d285-b3af/

Significantly, the Time Controller will gain a record button.   This will trigger a listener class to start listening to time steps, and then start Playing the scenario.

When record is pushed again, recording will stop, and the export wizard will open.

Once the export wizard has opened, it will trigger the PPT export.

PPT export is covered in a separate issue.

IanMayo:
Fixes #3025


org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/AnimatedGif.java
```diff
@@ -0,0 +1,165 @@
...
+                  imageData.width,
+                  imageData.height);
+
+              /* Now loop through the images, creating and drawing each one
+               * on the off-screen image before drawing it on the shell. */
+              int repeatCount = loader.repeatCount;
+              while (loader.repeatCount == 0 || repeatCount > 0) {
+                switch (imageData.disposalMethod) {
+                  case SWT.DM_FILL_BACKGROUND:
+                    /* Fill with the background color before drawing. */
```
IanMayo: @ayeshare - the static checker is flagging an issue in this block. At line 92 we check if bgColor is non-null. But, in the next line we assign it to null, then don't change it.

So, at line 92 it can't be non-null.  I can't remember where this animation code came from, but can you check that we haven't introduced an issue?

Thanks.

Issue #2960, opened 2018-07-02
D11 - Offer to trim narrative to current time period


IanMayo:
_As an_ **analyst** _I wish to_ **control how much narrative content is loaded** _so that_ **the entries will already be there if I load more track data in the future**.

## Background
In addition to the track data, analysts get a lot of value from diaries kept during the exercise, such as the traditional _Captain's Log_.  This data is called **Narrative data**.  The data importer is called `ImportNarrativeDocument`.

Debrief loads this data, and shows the diary entry nearest to the current time on the Time Controller.

Some narratives can last for several months, so only entries that span the time for the currently loaded tracks are entered - all others are ignored.

But, some analysts find this a pain. They may be looking at a 24 hour period of data. When they load tracks for the next 24 hours, Debrief hasn't got the narrative data for the new period (since it dropped those entries).  If Debrief had stored the whole narrative - the data will still be there.

## Strategy

We don't know which of the above the analyst wants - so we have to find out via a dialog box.  We should offer a popup dialog that lets the analyst choose between:
* _Only load entries within the current loaded time period_
* _Load all narrative entries_

There should be a checkbox to remember to apply the current setting automatically next time.  There should be a preference field (in the **Maritime Analysis** preferences pane) to clear this _remember previous_ flag, titled "Ask user whether to include all narrative entries".

## UX

![d11 optional trim narrative](https://docs.google.com/drawings/d/e/2PACX-1vTLjqq84gDDbGEkPR0_Kl35A2v9uQFx9udzLWZcWU-47ou3_yjK44ri7fgIH4p1350emhAqyBWBsF8z/pub?w=459&h=374)

## Coding

We have a challenge in implementing this, the importer is in the debrief.legacy package, which doesn't contain any SWT/RCP code.  So, we'll have to give it a helper class that can use the above UI to get the answer from the user.  The `ImportNarrativeDocument` already has a QuestionHelper object that gets injected.  (I looked, and it doesn't seem right to extend that interface to include this "special" question.  So, we can can introduce a new interface that will include findWhatToImport - it will return a enum in one of three states: TRIMMED_DATA, ALL_DATA, CANCEL.  The instance will be created and injected from `DebriefPlugin` just like we do for QuestionHelper.

[...]

PR #3072, opened 2018-08-13
Implemented 2960, dialog for narrative entries in word docs

ayeshare:
supports #2960


org.mwc.debrief.core/src/org/mwc/debrief/core/ui/ImportNarrativeDialog.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Dialog popped up from {@link ImportNarrativeHelper}
+ * @author Ayesha
+ *
+ */
+public class ImportNarrativeDialog extends Dialog
+{
+  private Button _btnLoadedTracks;
+  private Button _btnAllData;
```
IanMayo: @ayeshare - the code checker is indicating that _btnAllData could have been a variable defined in the createDialogArea method - since it is only referred to in that method.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Word/ImportNarrativeDocument.java
```diff
@@ -699,6 +699,32 @@ public NarrEntry(final String entry) throws ParseException
...
+    private String name;
+    ImportNarrativeEnum(String string){
+      this.name = string;
+    }
+    public String getName() {
+      return this.name;
+    }
+    public static ImportNarrativeEnum getByName(String name) {
+      switch(name) {
+        case "trimmed-data":return TRIMMED_DATA;
```
IanMayo: @ayeshare - you're going beyond my ability with enums, here - so I'm keen to learn.   But, wouldn't it be better to replace these duplicated strings (`"trimmed-data"`) with final variables?

org.mwc.debrief.core/src/org/mwc/debrief/core/ui/ImportNarrativeHelper.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class ImportNarrativeHelper implements TrimNarrativeHelper
+{
+
+  private static final String PREF_DEF_NARRATIVE_CHOICE = "defaultNarrativeEntryChoice";
+  @Override
+  public ImportNarrativeEnum findWhatToImport()
+  {
+    final Display targetDisplay;
+    String defaultChoice = CorePlugin.getDefault().getPreference(PREF_DEF_NARRATIVE_CHOICE);
```
IanMayo: @ayeshare - users also need a way to clear the "re-use next time" choice.  

So, we need a checkbox in _Debrief specific settings_.  It should be titled "Re-use existing choice for trimming imported narratives".  Once the user has selected a preference, the box should get ticked.  But, when the box gets cleared by the user, the popup dialog should start appearing again.

I guess the value of the checkbox is controlled by the presence of the default choice preference value. If one is there, the box gets ticked.  If one isn't assigned, the box is clear.  

Unticking the box would clear the default choice.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Word/ImportNarrativeDocument.java
ayeshare: I can replace it with a constant but in enum the definition of Enum should be the very first statement.  So it will still be duplicated.

org.mwc.debrief.core/src/org/mwc/debrief/core/ui/ImportNarrativeHelper.java
ayeshare: implemented the preference

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Word/ImportNarrativeDocument.java
IanMayo: Ok, workaround applied:
image

PR #3099, opened 2018-09-04
2884_svg_icons

### LINKED ISSUE #2884: M27 - Icon update
Original requirement: `Updates to look and feel (e.g. Icons).`

As an ***Analyst*** I wish to ***use a wider set of icons*** so that ***my plots carry a more effective message***.

## General
The use of vehicle type-specific icons aids the viewers’ understanding of what is happening in a Debrief scenario.  Debrief contains a limited set of icons, but in general a wider set of standard icons would be beneficial to both analysis and the presentation of analysed product 

The shapes of the existing icons are hard-coded into Java, but we wish to introduce a tidier architecture, where icons are defined in SVG files.  This will make it easier to make changes to the icons, and to introduce user-generated icons in the future.

## Assumptions
The following capabilities are required:
* Introduce a large number (~60) of new icons
* Ability to use specify these new icons at both REP import and from within Debrief.
* A customer-provided document represents all new icons required.
* The customer will provide names to use for this icon set.
* We will support the following sub-set of SVG features:
SVG element types
* We won't support the full set of SVG formatting attributes
* We will scale the SVG icon to 1.0 by 1.0, centred on 0,0. Then we will apply SymbolSize property to this.
* If an icon just uses black, we will allow user to override color.  If an icon uses shades other than black, we will use those shades
* Icon files will need to include _some kind of_ boolean flag, to indicate if they should be rotated to match the subject's course.

It is expected that this task will create a new file format that will be utilised Task M12 (user defined icons, #2872 )

## Strategy

* [x] Get an artist to produce the set of SVG files, scaled to 1.0 by 1.0, centred at 0,0.
* [x] Store the files in a suitable location (prob org.mwc.cmap.legacy)
* [x] Create new SVGShape class, that extends PlainSymbol
* [x] Extend `SymbolFactory` to allow the inclusion of SVGShapes
* [x] SVGShape will take a filename in constructor
* [x] The first time `SVGShape.paint()` is called, we'll load the SVG from file, parse the XML, and translate it into a cached series of circle/rectangle/polygon/polyline paint elements/operations
* [x] We'll also determine the bounds of the shape, and the origin for the shape (TBD)
* [x] SVGShape will render SVG data in paint() method
* [x] We'll define new shapes in SymbolFactory.  The `SymbolPropertyEditor` will need to be extended so it supports the full range of icons
* [ ] This set of standard icons will be stored in a specific location in org.mwc.cmap.core plugin folder.  (_effectively_, an analyst can edit/overwrite an SVG image by replacing it in this folder)

## Outstanding items
* [ ] Consider if we want to support stroke & stroke-width, based on how the Debrief versions look compared to the originals

## Testing
* we need to handle missing SVG files
* we need to handle SVG features that we aren't able to render (_path_).
* we should add all of the symbols to the `shapes.rep` test file
* test that all-black icons can have color over-written
* test that multi-color icons retain their colors
* test that large icon (10 by 10 units) is scaled down to 1 by 1 units - and reflects Symbol Size parameter

## UX/UI 
None

## Training needs
**Tutorial**:  No, none required.
**User guide**:   No, none required.  (**Note:** we'll have to include reference once we support user-controlled icons).
**Reference Guide**: document assumptions in how SVG icons are detailed (origin, color, rotation)

## Graphics
* The full set of icons, in .SVG format
* Test icons, to include multi-color & oversized.

## Media
* It's not all that exciting.  We could provide an image of the folder of icons, plus a image of the icons loaded into Debrief.  An image containing the full set of images would also be useful for the reference manual.

## Assets
* Yes - the full set of icons.
* A sample `.svg` file is attached below

### LINKED ISSUE #3109: M27.2 Select new icons from REP file symbology
We're introducing a large block of new icons.

Users can select them from a drop-down list in the property editor.  But, we also need to be able to assign the new symbols from a REP file symbology:
http://www.debrief.info/tutorial/index.html#reference.html#replay_symbology

There are a few unused letters/numbers/symbols remaining.  We can assign those. Beyond that, we'll need to introduce a symbology parameter to indicate the symbol name.

So,
* [ ] Decide/agree on which icons should receive dedicated symbols
* [ ] Update the user guide to assign the new symbols to new letters
* [ ] Update the user guide to describe the new parameters
* [ ] Modify ImportReply to support the above features
* [ ] Modify SymbolFactory to include the relevant IDs

````
951212 050700.000 NELSON @C[SYMBOL=Coastguard] 22 11 10.08 N 21 43 20.34 W 270.5   2.0      0 
951212 050701.000 COLLINGWOOD @C[SYMBOL=Coastguard,LAYER=Support] 22 11 10.08 N 21 43 20.34 W 270.5   2.0      0 
````

### PR DESCRIPTION
**Pull request by saulhidalgoaular**
Fixes #2884 
Fixes #3109

### CODE REVIEW

File: org.mwc.cmap.legacy/src/MWC/GUI/Shapes/Symbols/SVG/SVGRectangle.java [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+  public SVGRectangle(Element dom)
+  {
+    super(dom);
+  }
+
+  @Override
+  public void render(CanvasType dest, double sym_size,
+      java.awt.Point origin_coords, double rotation_degs)
+  {
+    final double x = Double.parseDouble(get_dom().attr("x")) * sym_size;
```
IanMayo: Minor point. From your implementation, I see that SVGRectangle only gets created once we're already in a paint process `SVGShape.paint()`.

So, I believe we should extract x, y, width, height in the constructor for SVGRectangle.  Then we know they'll just be called once.

org.mwc.cmap.legacy/src/MWC/GUI/Shapes/Symbols/SVG/SVGElementFactory.java
```diff
@@ -0,0 +1,62 @@
...
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. 
+ */
+package MWC.GUI.Shapes.Symbols.SVG;
+
+import org.jsoup.nodes.Element;
+
+public class SVGElementFactory
+{
+  private Element _svgElement;
+
```
IanMayo: In a _normal_ factory pattern implementation, we have a single factory instance, and that instance returns relevant new child instances from a `getXXX()` method:
https://www.tutorialspoint.com/design_pattern/factory_pattern.htm

So, I think this should have an empty constructor, and should not have the internal `_svgElement` member.  All it needs is for `getInstance()` to take a `svgElement` as a parameter.

org.mwc.cmap.legacy/src/MWC/GUI/Shapes/Symbols/SVG/SVGRectangle.java
saulhidalgoaular: I totally agree with you!! I am doing the changes mentioned

PR #3103, opened 2018-09-06
initial revision of US-M24, insert sensor arc

ayeshare:
supports user story M24

### DIFF
--- a/org.mwc.cmap.media/META-INF/MANIFEST.MF
+++ b/org.mwc.cmap.media/META-INF/MANIFEST.MF
@@ -24,6 +24,7 @@ Bundle-ClassPath: lib/commons-cli.jar,
  lib/jni_helper.jar,
  lib/commons-io-2.1.jar,
  lib/xuggle-xuggler.jar
-Export-Package: org.mwc.cmap.media.time,
+Export-Package: org.mwc.cmap.media,
+ org.mwc.cmap.media.time,
  org.mwc.cmap.media.utility,
  org.mwc.cmap.media.views
--- a/org.mwc.debrief.core/META-INF/MANIFEST.MF
+++ b/org.mwc.debrief.core/META-INF/MANIFEST.MF
@@ -28,6 +28,7 @@ Export-Package: com.topografix.gpx.v10,
  org.mwc.debrief.core.contenttype,
  org.mwc.debrief.core.creators.chartFeatures,
  org.mwc.debrief.core.creators.shapes,
+ org.mwc.debrief.core.creators.shapes.dynamic,
  org.mwc.debrief.core.editors,
  org.mwc.debrief.core.editors.painters,
  org.mwc.debrief.core.editors.painters.highlighters,
@@ -64,5 +65,7 @@ Require-Bundle: org.mwc.cmap.core,
  org.eclipse.ui.intro,
  org.eclipse.swt,
  org.eclipse.jface,
- org.eclipse.ui.workbench
+ org.eclipse.ui.workbench,
+ org.eclipse.nebula.jface.cdatetime;bundle-version="0.14.0",
+ org.eclipse.nebula.widgets.cdatetime;bundle-version="0.14.0"
 Bundle-ActivationPolicy: lazy
--- a/org.mwc.debrief.core/plugin.xml
+++ b/org.mwc.debrief.core/plugin.xml
@@ -608,6 +608,26 @@
                     <reference definitionId="org.mwc.debrief.activePlotEditor"> </reference>
                 </visibleWhen>
             </command>
+            <separator
+                  name="org.mwc.debrief.core.separator1"
+                  visible="true">
+            </separator>
+            <menu
+                  id="dynamicShapes"
+                  label="Dynamic Shapes">
+               <command
+                     commandId="org.mwc.debrief.core.creators.shapes.InsertSensorArc"
+                     icon="icons/16/sensor.png"
+                     id="org.mwc.debrief.core.creators.shapes.InsertSensorArc"
+                     style="push">
+                  <visibleWhen
+                        checkEnabled="false">
+                     <reference
+                           definitionId="org.mwc.debrief.activePlotEditor">
+                     </reference>
+                  </visibleWhen>
+               </command>
+            </menu>
         </menuContribution>
 
         <menuContribution locationURI="toolbar:org.mwc.debrief.core.EditorShapes?after=shapes">
@@ -890,6 +910,11 @@
 
         <command defaultHandler="org.mwc.debrief.core.operations.ExportGPX"
             id="org.mwc.debrief.core.operations.ExportGPX" name="Export GPX"> </command>
+        <command
+              defaultHandler="org.mwc.debrief.core.creators.shapes.dynamic.InsertSensorArc"
+              id="org.mwc.debrief.core.creators.shapes.InsertSensorArc"
+              name="&amp;Sensor Arc">
+        </command>
     </extension>
 
     <extension point="org.eclipse.ui.newWizards">
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GenerateNewInsertSensorArcAction.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GenerateNewInsertSensorArcAction.java
@@ -1,0 +1,239 @@
+/**
+ * 
+ */
+package org.mwc.debrief.core.ContextOperations;
+
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.core.commands.operations.IUndoableOperation;
+import org.eclipse.core.runtime.IAdaptable;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jface.action.Action;
+import org.eclipse.jface.action.IMenuManager;
+import org.eclipse.jface.action.Separator;
+import org.eclipse.jface.dialogs.TrayDialog;
+import org.eclipse.jface.wizard.WizardDialog;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.PlatformUI;
+import org.mwc.cmap.core.CorePlugin;
+import org.mwc.cmap.core.DataTypes.Temporal.TimeProvider;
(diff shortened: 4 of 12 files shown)


IanMayo:
@ayeshare - in the current implementations, we know the target track before the wizard opens, so we no longer need to request it in the Wizard:
screen shot 2018-09-06 at 11 37 20

IanMayo:
Could you give the text boxes a default width?  It would be neater if they were all the same width. The width should be slightly wider than the current `1000` box, wide enough to accommodate 5 digits.
screen shot 2018-09-06 at 11 40 35

IanMayo:
Since we know the target track before the wizard opens, I've realised we can use the Track's start and end times, rather than the start/end time for the whole plot, via `track.getStartDTG()` and `track.getEndDTG()`
screen shot 2018-09-06 at 11 42 23

IanMayo:
Some more feedback for @ayeshare 
* [ ] Need validation on the location fields (centre of circle, corners of rectangle)
* [ ] Need to robustly handle text being entered for integer (radius)
* [ ] I've separately asked that the menu items be named `Dynamic Rectangle` instead of just `Rectangle`.
* [ ] `InsertDynamicXXX` classes would benefit from refactoring, to pull out common code
* [ ] The Codacy static checking plugin has suggested some improvements (see _Some checks were not successful_ on this page)
* [ ] After talking to a user, I've realised it would make more sense for the dynamic shapes to have optional Start _AND_ End times.  So, I've modified the [UI slide](https://docs.google.com/presentation/d/1px8b2r6bvBzkPmB4OwkhvLEUMeWo2VspoXfMnPNVDBI/edit#slide=id.g3ef6d0c7af_0_0). You may be able to re-use the `SensorArcTimingsWizardPage` code.

IanMayo:
@ayeshare - we also need to tidy the formatting on the Circle Bounds page:
* [ ] Widen lat & long fields so we can see whole of location field
image

org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicCircleWizard.java
```diff
@@ -56,11 +56,19 @@ public void addPages()
   public boolean performFinish()
   {
     Date startTime = _shapeTimingsPage.getStartTime();
```
IanMayo: @ayeshare - we've got lots of code duplication in these three blocks of code for the DynamicShapes.

Is it an option to refactor, to pull out common elements?

I'd _typically_ handle it by the core class having `abstract protected PlainShape getShape()` that gets called from the core code.

org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java
```diff
@@ -0,0 +1,134 @@
...
+  private String pageName;
+  private static Font font;
+  private static Font regularFont;
+  
+
+  protected DynamicShapeBaseWizardPage(String pageName)
+  {
+    super(pageName);
+    this.pageName = pageName;
+    System.out.println("Current pagename:"+pageName);
```
IanMayo: @ayeshare - I don't think we need the previous debug line.

org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * @author Ayesha <ayesha.ma@gmail.com>
+ *
+ */
+public abstract class DynamicShapeBaseWizardPage extends WizardPage
+{
+  public static final String TIMINGS_PAGE = "Timings";
+  public static final String BOUNDS_PAGE = "Bounds";
+  public static final String STYLING_PAGE = "Styling";
+  private String pageName;
```
IanMayo: @ayeshare - pageName can be made final.  We assign it once (in the constructor), then never modify it.

org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java
```diff
@@ -0,0 +1,134 @@
...
+  private static Font font;
```
IanMayo: @ayeshare - I'm pretty sure (though I may be mistaken) that `font` and `regularFont` should not be made static.   In each `createBaseControl` we over-write the previous static instance.  Then when we call `dispose()` we may only be releasing the _newest_ font, not the overwritten one.

Fortunately, we can probably only have one dialog open at a time, so we haven't had any exceptions thrown, but I think we should correct it.

org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java
ayeshare: done

IanMayo:
@ayeshare - I may have missed this in the sensor arcs PR, but I see we need the license statement at the top of all new files:
````
/*
 *    Debrief - the Open Source Maritime Analysis Application
 *    http://debrief.info
 *
 *    (C) 2000-2018, Deep Blue C Technology Ltd
 *
 *    This library is free software; you can redistribute it and/or
 *    modify it under the terms of the Eclipse Public License v1.0
 *    (http://www.eclipse.org/legal/epl-v10.html)
 *
 *    This library is distributed in the hope that it will be useful,
 *    but WITHOUT ANY WARRANTY; without even the implied warranty of
 *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 */
````

File: org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java (hunk 4 above)
Comment (ayeshare) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java]: done

File: org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java (hunk 3 above)
Comment (IanMayo) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeBaseWizardPage.java]: @ayeshare - correction, we don't need the pageName field at all.  I see it gets passed to the parent, so we can retrieve it via `getName`;

File: org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/dynshapes/DynamicShapeStylingPage.java [hunk 5]
```diff
@@ -0,0 +1,116 @@
...
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Text;
+
+/**
+ * @author Ayesha
+ *
+ */
+public class DynamicShapeStylingPage extends DynamicShapeBaseWizardPage
+{
+  private String _type;
```
IanMayo: this is another field that can be `final` - we assign it once, but never change it.

IanMayo:
I've got an outstanding issue. I've been debugging it, but it won't help either of us if I fix it, and then you run out of work :-)

We need to validate the location that has been entered.  It's possible to have spaces in the location latitude/longitude control.  When that happens the control returns null, and our following logic fails. So, we need to validate that control before we move on.  We can do it this way:

````
public boolean isPageComplete()
  {
    return (!_txtRadius.getText().isEmpty() && isValidRadius(_txtRadius
        .getText()) && (_txtCentre.getValue() != null));
  }
````
But, we need to fire that event when location changes, not just when the radius changes. That's another issue. The CWorldLocation control isn't telling it's modified listeners that the value has changed.  That needs debugging.

We should also do the above location verification for the Rectangle bounds page.

_Ideally_ our wizard would be able to inform the user that a location is corrupt.

IanMayo:
Aah, and another one.  If we tick the box to provide start (or end), and then uncheck the box, the value in the control still gets used.

We need to check the status of the checkbox before we return the date value:
````
  public Date getStartTime()
  {
    return _chkStartTime.getSelection() ? _cStartTime.getSelection() : null;
  }
````

IanMayo:
Oops, and another.  This may be MACOS specific, but the lat/long controls need to be about 10px wider.

image

ayeshare:
> 
> 
> I've got an outstanding issue. I've been debugging it, but it won't help either of us if I fix it, and then you run out of work :-)
> 
> We need to validate the location that has been entered. It's possible to have spaces in the location latitude/longitude control. When that happens the control returns null, and our following logic fails. So, we need to validate that control before we move on. We can do it this way:
> 
> ```
> public boolean isPageComplete()
>   {
>     return (!_txtRadius.getText().isEmpty() && isValidRadius(_txtRadius
>         .getText()) && (_txtCentre.getValue() != null));
>   }
> ```
> 
> But, we need to fire that event when location changes, not just when the radius changes. That's another issue. The CWorldLocation control isn't telling it's modified listeners that the value has changed. That needs debugging.
> 
> We should also do the above location verification for the Rectangle bounds page.
> 
> _Ideally_ our wizard would be able to inform the user that a location is corrupt.

I had seen this issue, but I missed it for the circle wizard.  Another issue is that error message is not displayed.

ayeshare:
> 
> 
> Oops, and another. This may be MACOS specific, but the lat/long controls need to be about 10px wider.
> 
> image

this one looks like is specific to mac

IanMayo:
@ayeshare - have pushed couple of minor changes, including the one that was probably Macos-specific.

We still need to fire the modified event when the lat/long fields are changed, since we currently only validate on radius-value change.

IanMayo:
@ayeshare - you'll need to `pull` before you start, because I've just pushed some refactoring into the branch.

I've just realised how much code is duplicated between the `DynamicXXXXWizard` classes. Would you mind refactoring them too?

IanMayo:
@ayeshare - we also need a bit more robust processing of Circle radius.

If I put letters into the box, an exception is silently thrown, but no feedback is displayed to user.

Example is that user may try to type `500m` (for meters).  I don't need us to parse the metres, displaying the radius error message is sufficient.

IanMayo:
@ayeshare - I think we also need to make the default size of the polygon wizard (or all of them), so we can see all the intro text.  The sample coords are missing for polygon, until I resize the wizard:
image

IanMayo:
@ayeshare - I've switched to catching ModifiedEvent, rather than SegmentEvent.  It now works on MACOS. Could you try it under MS Windows?

PR #3110, opened 2018-09-10
Feature/m11 unit centric view

### LINKED ISSUE #2871: M11 - Unit centric view
Original requirement: `Plot tracks as displacement from a given track (unit centric view)`

As an ***Analyst*** I wish to ***view how other tracks relate to my primary track*** so that I can ***track their relative disposition over time***.

## General
This feature is a special plotting mode that shows the position of non-primary tracks relative to the primary track, at the point in time for the non-primary track measurement.

By definition, the primary track is represented as a single point in the centre of the screen, with other tracks represented as lines on the plot.  For a non-primary track that maintains a constant station at, say, 1km distance on a bearing of 135 degrees from primary, this will be displayed as a single point at that range/distance.

## Assumptions
* Measurements are frequent enough that we can use interpolation to find synch'd positions 
* Background (reference) datasets have no meaning in this context, so should not be plotted
* Annotations (shapes) _may_ have meaning if we're working in Snail Mode - since we can calculate their relative location at that point in time.

## Strategy
* Introduce new View, titled `Unit Centric View`
* Introduce new `RelativeProjectionParent` object that returns heading/location at requested time, rather than at current `TimeController` time.
* Produce test-case that gives predicable results.

## Testing
* verify that unit tests give expected results
* verify that sample.dpf gives expected results.  A plot of range between tracks can be used to verify that the range from a non-primary to the centre of the plot should match that on the plot. A relative bearing plot should also prove useful

## UX/UI 
None

## Training needs
**Tutorial**:  No, too nich
**User guide**:   Yes, brief introduction within TimeController content. Maybe screenshot
**Reference Guide**: 

## Graphics
* Yes, we need two existing plotting modes, plus the new one.

## Media
* A video should prove useful in explaining what's happening, particularly involving switching between modes.

## Assets
* Nil

IanMayo:
Fixes #2871

### DIFF
--- a/org.mwc.asset.NetAsset2/src/org/mwc/asset/netasset2/plot/VPlot.java
+++ b/org.mwc.asset.NetAsset2/src/org/mwc/asset/netasset2/plot/VPlot.java
@@ -147,7 +147,7 @@ public class VPlot extends Composite implements IVPartMovement, IVTime,
 				{
 
 					@Override
-					public double getHeading()
+					public double getHeading(final HiResDate dtg)
 					{
 						double res = 0;
 						if (_recentStatus != null)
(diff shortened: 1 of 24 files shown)


IanMayo:
Drop this branch.  We're implementing it via #3111 instead.

PR #3129, opened 2018-09-19
M22 Paste rep from clipboard

### LINKED ISSUE #2880: M22 - Paste REP from clipboard
Original requirement: `Paste a rep file from clipboard`

As an **scientist** I wish to be able to **paste REP data from clipboard**  in order to **quickly load data from MS NotePad if I'm either learning about a new REP format, or loading data that I have manipulate by hand**

## Strategy
* [ ] Paste into existing plot, by pasting into empty area of Outline View
* [ ] ~Paste into selected folder (or track) in Outline View~ . (No, we let the importer decide where they go)

A `Paste REP from clipboard` action generator will be defined, and added to the right-click menu.  When the user right-clicks on an item, we'll:
* See if there's a single selected item
* See if the item is a Track, or a BaseLayer
* See if there is plain text on the clipboard
* See if there's a valid importer (we'll use a new ImportReplay method to try `getImporterFor` for each line of the clipboard. If we can process every line, then we're happy to continue)
* Verify we can paste this data into the target (see below)
* Create "Paste REP from Clipboard" action.

### Logic for acceptable paste 

Since starting the work, team conversations have lead the logic for acceptable paste to avoid the concept of where data can be pasted into.

Letting the user dictate (via right-click) where to paste introduces the problem of: do we follow the attributes in the rep file, or do we paste into the selected location.

It is better to ignore where the user right-clicked, and just paste into the top level.  So, don't offer paste from REP unless nothing has been selected in the Outline view.

(in the future users may determine special cases where data should be pasted into specified folder,  but we'll wait until that arises)

### Logic for acceptable paste - DEPRECATED

The following item types can accept the indicated data types:
* **SensorWrapper** : Sensor
* **TrackWrapper**: Position, Sensor, TUA, Sensor_Arcs
* **BaseLayer**: all annotation types (positions go into lightweight track)
* **Empty space (no selection)**: Positions as new track, annotations into (_ask user_) folder.

The code for processing the import is in ImportReplay.importThis().  But, we may have to make some API adjustments to account for ImportReplay expecting to process a file, rather than a text stream.

**Note:** there's an example of reading data into a Layers object in `DataFeed.insertData()`.

If the paste is applying to the plot, then we just dump the data into the plot's _layers object.  If we're pasting into a Layer, we need to dump the data into a temporary Layers object, then copy/move it's contents to the specified layer.  If the new Layers object just contains a single layer of bits, then the contents of this layer (not the layer itself) can be copied into the target layer.

## Testing
* [ ] Need to check we tidily handle corrupt (invalid) REP lines
* [ ] Need to check we can paste into folder/track, and warn user if we're trying to paste into invalid target
    * [ ] If tracks are pasted into a folder, they should be turned into LightweightTracks
    * [ ] only track (or sensor) items can be pasted into a Track

## UX/UI 
None

## Training needs
**Tutorial**: Quite a niche task. But, if it isn't introduced there, users may not know about it.  So, maybe just a comment describing it.
**User guide**:  No, don't think so.

## Graphics
None

## Media
Yes, would make a good demo.  Put some REP data into Notepad, representing single position.  Copy into Debrief via Clipboard. Then make modification to entry (change time & location), copy/paste again.

## Assets
None.

ayeshare:
Fixes #2880


org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -0,0 +1,216 @@
...
+    StringTokenizer tokens = new StringTokenizer(content,"\\r?\\n");
+    int lineCount = 0;
+    while(tokens.hasMoreTokens() && lineCount<=6) {
+      String line = tokens.nextToken().trim();
+      if(line.startsWith(";") && !line.startsWith(";;")) {
+        StringTokenizer lineTokens = new StringTokenizer(line);
+        if(lineTokens.hasMoreTokens()) {
+          String firstWord = lineTokens.nextToken();
+          switch(firstWord) {
+            case ";CIRCLE:":
```
IanMayo: @ayeshare - it will be a burden to maintain this list as we add new items.  Could you create a fancy regexp that looks for `;`, then up to _40?_ non white-space chars, then `:` ?

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -0,0 +1,215 @@
...
+              proceed = true;break;
+            default:proceed=false;break;
+          }
+        }
+      }
+      else {
+        StringTokenizer lineTokens = new StringTokenizer(line);
+        if(lineTokens.hasMoreTokens()) {
+          String firstWord = lineTokens.nextToken();
+          if(firstWord.matches("\\d{6}+")) {
```
IanMayo: @ayeshare - we do also support an 8 digit date, as described here:
http://www.debrief.info/tutorial/reference.html

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
ayeshare: this is done

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
ayeshare: done

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -128,7 +128,7 @@ private boolean isContentImportable(final String content) {
         StringTokenizer lineTokens = new StringTokenizer(line);
         if(lineTokens.hasMoreTokens()) {
           String firstWord = lineTokens.nextToken();
-          String regex = "^;[A-Z_]{3,40}+:$";
```
IanMayo: hey, well spotted :-)

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -541,6 +541,36 @@ public final void testParseSymbologyColor()
...
+";POLY: @GA30 21.9 0 0 N 21.5 0 0 W 22 0 0 N 21.8 0 0 W 22.1 0 0 N 21.5 0 0 W test poly\r\n"+
+";POLYLINE: @C 21.1 0 0 N 21.5 0 0 W 21.2 0 0 N 21.8 0 0 W 21.3 0 0 N 21.5 0 0 W test polyline\r\n"+
+";NARRATIVE:  951212 050200 NEL_STYLE comment 3\r\n";
+      ImportReplay testImporter = new ImportReplay();
+      Layers tmpLayers = new Layers();
+      Layers dest = new Layers();
+      testImporter.setLayers(tmpLayers);
+      testImporter.importThis(textToPaste);
+      testImporter.injectContent(tmpLayers, dest, true);
+      assertTrue(((Layer)tmpLayers.findLayer(ANNOTATION_LAYER)).elements().hasMoreElements());
```
IanMayo: @ayeshare - could you also check that the layer contains the correct number of elements?

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -541,6 +541,36 @@ public final void testParseSymbologyColor()
       assertEquals("sym type correct", "Hidar", label2.getSymbolType());
       assertEquals("color correct", DebriefColors.PURPLE, label2.getColor());
     }
+    
+    
+    public final void testPasteRep() {
+      String textToPaste =
```
IanMayo: @ayeshare - I think it would be sensible to add a few track positions (you could take a few from the start of `boat1.rep`).  You should then check that the new Track is in present in the `Layers` object, and that it contains the correct number of points.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1255,32 +1285,52 @@ final private String getThisSymProperty(final String symbology,
...
+      Layer existingLayer = destination.findLayer(l.getName());
       if(existingLayer==null) {
         Layer layerToAdd = (Layer)l;
-        existingLayers.addThisLayer(layerToAdd);
-        elementsAdded.add(layerToAdd);
+        if(performAdd) {
+          destination.addThisLayer(layerToAdd);
+        }
+        else {
+          destination.removeThisEditable(null, l);
```
IanMayo: @ayeshare - I'm pretty sure that `destination.removeThisLayer(layerToAdd);` would be a better way of deleting hte layer.

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -118,57 +120,36 @@ public IStatus undo(IProgressMonitor monitor, IAdaptable info)
 
   private boolean isContentImportable(final String content) {
```
IanMayo: it's probably worth introducing a unit test, to verify this passes true/false for the right types of string.

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -118,57 +120,36 @@ public IStatus undo(IProgressMonitor monitor, IAdaptable info)
...
           String firstWord = lineTokens.nextToken();
-          if(firstWord.matches("\\d{6}+")) {
-            proceed=true;
+          if(!(firstWord.matches("\\d{6}+") || firstWord.matches("\\d{8}+"))) {
+            proceed=false;
           }
         }
+        
+      }
+      //we do it only for 6 lines
```
IanMayo: this is really minor, but...  I think it may be tidier to use a `while` loop.  It will put all the _loop control_ logic in the same place, at the start.

I guess max_lines could Math.min(6, lines.length), then `while (i < max_lines && proceed)`

org.mwc.debrief.data_feed/src/org/mwc/debrief/data_feed/views/DataFeed.java
```diff
@@ -340,8 +362,8 @@ public void setFocus()
 	public void insertData(final String data)
```
IanMayo: psst @ayeshare, why has this file been changed?

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1242,6 +1271,66 @@ final private String getThisSymProperty(final String symbology,
     }
     return null;
   }
+
```
IanMayo: we don't use the object returned by this method. We no longer have to keep a list of what has been added.  This logic can be removed.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1242,6 +1271,66 @@ final private String getThisSymProperty(final String symbology,
...
+    Enumeration<Editable> tempElements = ((Layer)l).elements();
+    while(tempElements.hasMoreElements()) {
+      Editable elem = tempElements.nextElement();
+      layerToAddTo.add(elem);
+      elementsAdded.add(elem);
+    }
+    return elementsAdded;
+  }
+  
+  public final void importThis(final String text) {
```
IanMayo: there's already an `importThis` method in this class.  It does lots of other _tidying up_ steps while doing the import.  

I _think_ it would be safer for us to re-use that method, since users will _probably_ expect that tidying up to happen, I think we should re-use that method.

To do that, I think we would extract the code to that puts the InputStream into a list of lines, then make a call to a common method that does all the other processing. Clearly this second method is the one that our PasteRepAction will call.

_Tip: since this is refactoring, we'll help ourselves by doing it after the new/updated unit testing is implemented. These tests will give us confidence that our refactoring is safe_.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1242,6 +1271,66 @@ final private String getThisSymProperty(final String symbology,
     }
     return null;
   }
+  
+  private List<Editable> addElementsToExistingLayer(Layer layerToAddTo,Editable l)
```
IanMayo: This method, and `injectContent` can both be made static.  

The advantage of making them static is we can be sure they don't rely on member state (member fields).

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -0,0 +1,156 @@
...
+        //See if there is plain text on the clipboard
+        if(isContentImportable(clipBoardContent)) {
+          parent.add(createAction(theLayers, clipBoardContent));
+        }
+
+      }
+    }
+  }
+  
+  private Action createAction(final Layers theLayers, final String clipboardContent) {
```
IanMayo: The Action title's a bit terse.  Could you make the more verbose tool-tip text the action title, too?

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -0,0 +1,156 @@
...
+    final Action doPasteAction = new Action("Paste REP")
+    {
+      @Override
+      public void run()
+      {
+        PasteRepOperation operation = new PasteRepOperation("Paste from clipboard",theLayers, clipboardContent );
+        CorePlugin.run(operation);
+      }
+    };
+    //doPasteAction.setImageDescriptor();
```
IanMayo: there's an icon we can use in `icons\16\paste.png`.  You can get it using `CorePlugin.getImageDescriptor...`

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
IanMayo: the `createAction()` method can be made static

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
ayeshare: done

org.mwc.debrief.data_feed/src/org/mwc/debrief/data_feed/views/DataFeed.java
ayeshare: because I removed the throws IOException, because exception was never thrown

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
ayeshare: done

org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /* (non-Javadoc)
+   * @see org.mwc.cmap.core.property_support.RightClickSupport.RightClickContextItemGenerator#generate(org.eclipse.jface.action.IMenuManager, MWC.GUI.Layers, MWC.GUI.Layer[], MWC.GUI.Editable[])
+   */
+  @Override
+  public void generate(IMenuManager parent, Layers theLayers,
+      Layer[] parentLayers, Editable[] subjects)
+  {
+    //see if there is nothing selected 
+    if(subjects.length==0) {
```
IanMayo: Oh dear, Ian's being picky again.  I think your eclipse is configured with the wrong braces format.  We have braces on a newline.

Here are steps to import the project styles:
https://github.com/debrief/debrief/wiki/On-boarding#project-settings

IanMayo:
One more request, please @ayeshare . When you're on top of your updates to this, could you work through some of the issues that Codacy has spotted?  It's quite likely that you're tidying code that you're not responsible for, but it all goes towards a more healthy codebase.

Don't worry about handling _all_ of the complaints, refactoring long methods can sometimes be really time-consuming, and may introduce fragility.  But a lot of the feedback is quick & simple to implement.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1275,72 +1395,128 @@ final private String getThisSymProperty(final String symbology,
...
-
-    final int numLines = countLinesFor(fName);
-
-    final Reader reader = new InputStreamReader(is);
-    final BufferedReader br = new ReaderMonitor(reader, numLines, fName);
+    final Reader reader;
+    BufferedReader br = null;
     String thisLine = null;
     try
     {
```
IanMayo: @ayeshare - in the following block, we duplicate the ReaderMonitor constructor.  This introduces a _slight_ fragility in having to track that we're correctly using reader and numLines, if the API changes at all. Just having the single use of the constructor keeps things simpler.

So, it would be _slightly_ better code if we just had the one constructor:
````
        final String nameToUse;
        if(fName != null)
        {
          nameToUse = fName;
        }
        else
        {
          nameToUse = "Pasted REP content";
        }
        br = new ReaderMonitor(reader, numLines, nameToUse);
````

A more terse version would be:
````
        final String nameToUse = fName != null ? fName : "Pasted REP content";
        br = new ReaderMonitor(reader, numLines, nameToUse);
````

Comment (IanMayo) [reply org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java]: Note: I've also expanded the text we use in the filename param, when we're pasting from REP.

File: org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java (hunk 7 above)
Comment (ayeshare) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java]: done

File: org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java (hunk 8 above)
Comment (ayeshare) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java]: done

File: org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java (hunk 14 above)
Comment (ayeshare) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java]: done

File: org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java (hunk 13 above)
Comment (ayeshare) [reply org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java]: done

File: org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GeneratePasteRepClipboard.java [hunk 17]
```diff
@@ -95,12 +95,10 @@ public IStatus undo(final IProgressMonitor monitor, final IAdaptable info)
     }
   }
 
-  private static Action doPasteAction;
```
IanMayo: @ayeshare - we didn't need to keep a copy of the action, all we need to do is create the action, and add it to the parent.

Issue #3127, opened 2018-09-18
M3.15 Export to PPTX should have play and record buttons


IanMayo:
The record mode toggle caused some confusion today.

This was partially a MS Windows UI issue, where the toggle state wasn't apparent.

But, the analyst in question was quite confident that separate (adjacent) play and record buttons would be easier to understand.

The mockup that details this new UI is here: https://xd.adobe.com/view/8cfc1dce-a683-4d0e-5c40-f7c8917a8dde-facd/

IanMayo:
Fixed, in #3161

PR #3161, opened 2018-09-27
M3.15 modified to use have both play and record buttons, pending issue

ayeshare:
supports #3127


IanMayo:
There are still some debug lines to be removed:

image

ayeshare:
> 
> 
> There are still some debug lines to be removed:
> 
> image

removed the debug lines

PR #3187, opened 2018-09-28
Bugfix/3186 icons missing

### LINKED ISSUE #3186: M27.4 Icons not in distribution
Analyst downloaded build dated 24th Sept.

SVG icons were present in `Symbol Type` drop-down list. But, when we selected one, the icon wasn't shown.

We looked at the error log, and got `FileNotFoundException`, thrown from `SVGShape.getDocument()#253`.

I'll check, but I guess there's a chance the folder isn't included in the build tab of ``MANIFEST.MF`

Yes, that's what it is:

image

`svg_symbols` needs to be ticked in the binary build.

IanMayo:
Fixes #3186

### DIFF
--- a/org.mwc.cmap.legacy/build.properties
+++ b/org.mwc.cmap.legacy/build.properties
@@ -1,7 +1,8 @@
 output.cmap_legacy.jar = bin/
 bin.includes = cmap_legacy.jar,\
                META-INF/,\
-               libs/
+               libs/,\
+               svg_symbols/
 jars.compile.order = cmap_legacy.jar
 source.cmap_legacy.jar = src/
 src.includes = libs/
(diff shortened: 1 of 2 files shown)


IanMayo:
Hello there @saulhidalgoaular - I'd like you have a look at this, please - when you want a break from PowerPoint.

The icons wouldn't load at the customer demo on Wednesday.  My first suspicion was that they weren't included in the dist, but they still didn't work on MS Windows.

I've investigated, and it appears that MS Windows needs the `svg_symbols` folder to be specified in the path.

To be honest, I don't really know how osx/linux managed to find the svg file when it didn't know the parent folder name :-)

Anyway, I'd appreciate you having a think about this issue, based on your understanding of how it was originally implemented.

I've introduced a workaround, but hopefully we can find a strategy that doesn't need the workaround.

PR #3206, opened 2018-10-04
hotfix - copy/paste TMA

IanMayo:
We have an issue.  When a TMA ellipse is copy/pasted into the same parent layer, it doesn't get prepended with `copy of`.

It should do.


IanMayo:
Closing. This issue involves opening a can of worms that it would be better to defer, until the issue causes a tangible problem.

PR #3207, opened 2018-10-04
3186 svg path issue

theanuradha:


IanMayo:
Hmm @theanuradha,  we don't close the input resource, do we?

IanMayo:
Fixed #3186

PR #3209, opened 2018-10-04
replaced the menu in view with dropdown actions (Dont Merge)

### LINKED ISSUE #3116: M11.1 Tidy UI for Unit Centric View
The adjustments for the unit-centric view are currently hidden in the drop-down menu:
image

It would be more intuitive to users that these options were available if each relevant toggle button had a drop-down menu.
image

The snail length would go below the snail-mode toggle. Note we still need the toggle status between "normal" and "snail" mode.

ayeshare:
Fixes #3116


IanMayo:
Dropping this, we had conflicts we couldn't resolve. Superceded by new PR #3211

PR #3251, opened 2018-10-08
3149 Starting Time Offset for PPTX Export

saulhidalgoaular:


IanMayo:
@saulhidalgoaular - there's an outstanding issue, with the times.  We collect the series of formatted time strings in the _times array.  But, we create a new formatted time in PlotTracks:
https://github.com/debrief/debrief/pull/3251/files#diff-7b6a47a331c93330c45a7268df50cb47R730

We should store the list of times in the TrackData object, and inject the list into the exported XML file.  This list should then be used in PlotTracks.

We're using the formatted time from the TimeController view.  This is so that the analyst can choose how the time is formatted (including whether to include full date, or seconds, or milliseconds...)

IanMayo:
Aah, and another thing @saulhidalgoaular 

If it's possible, I'd like us to defer the display of the callouts until the relevant track starts.

This may mean animating the appearance of all of the callouts, but with most having a delay of zero.

saulhidalgoaular:
I have done some manual tests, and it looks OK. Please, confirm just in case. Thanks.

IanMayo:
Just one minor change, it tripped over when the TimeController date format was `yy/mm/dd hh:mm`, since it tried to create several file paths, due to the `\` separators in the filename.

The time strings are working properly too :-)

Good piece of work @saulhidalgoaular

PR #3289, opened 2018-10-16
Increment versions, for release

IanMayo:


IanMayo:
Drop, we needed to increment develop.

Issue #3237, opened 2018-10-08
M24-UI for dyn’ shapes - No validation for invalid inputs while creating Dynamic shapes


helenayele:
### Steps to Reproduce:
1. Create Dynamic poly
2. Input some values for the coordinate, make some characters like on this SS - https://gyazo.com/21f71d0251e50affede923403f943661
3. Click Next and give name and select color , before reaching this step though Debrief silently fails and logs this 
java.lang.NumberFormatException: For input string: "-sfsdfswe"
        at sun.misc.FloatingDecimal.readJavaFormatString(Unknown Source)
        at sun.misc.FloatingDecimal.parseDouble(Unknown Source)
        at java.lang.Double.parseDouble(Unknown Source)
        at java.lang.Double.valueOf(Unknown Source)
        at org.mwc.debrief.core.wizards.dynshapes.DynamicPolygonBoundsPage.isCoordinatesValid(DynamicPolygonBoundsPage.java:129)
        at org.mwc.debrief.core.wizards.dynshapes.DynamicPolygonBoundsPage.isPageComplete(DynamicPolygonBoundsPage.java:108)
        at org.mwc.debrief.core.wizards.dynshapes.DynamicPolygonBoundsPage$1.modifyText(DynamicPolygonBoundsPage.java:77)
        at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:183)
        at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
        at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4145)
        at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1055)

### Expected Result:
It should validate the user input, and shouldn't accept invalid values like the above

helenayele:
when trying to delete characters using backspace, it throw Unhandled event loop exception - https://gyazo.com/2ceb809f080a57e79fc5112f9e57e177

PR #3300, opened 2018-10-16
fixed validation

ayeshare:
supports #3237


IanMayo:
@ayeshare - I only thought of this after I'd merged it.

A better *engineering practice* for this would to have started off with adding a (failing) unit test to the class, declared like this one:
https://github.com/debrief/debrief/blob/develop/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportDynamicRectangle.java#L277

The test would try to import an invalid line of REP data.  It would fail, then you'd fix the `isCoordinatesValid()` method call, and the test would pass.

No, the same problem _probably_ won't arise again, but we'd be increasing our test coverage, which brings long term benefits.

Next time :-)

Issue #3305, opened 2018-10-17
M19 - Video/Audio sync- starting player from Timecontroller doesn't change the play icon in media player view


helenayele:
When I start playing from time controller after dragging video file to editor, even if the media player plays, it doesn't change its play icon.
### Steps to Reproduce
1. Open sample.dpf file
2. Drag 19951212_103415.avi file on to the opened plot from other_format folder
3. Open Time controller and start playing
See that media player also starts but its icon is still not changed  - https://gyazo.com/fb51de08b67b892bea95d3391f00829b
### Expected Result
Its icon should be changed to 'playing' mode

ayeshare:
fixed, but cant create a pull request for this. branch is 3305_videoplayer_play_icon

IanMayo:
Have had a play with the PR.  We've got some inconsistencies in the UI response - switching off from one player doesn't always make the other switch.

But, while playing, I have reflected on the UI, and think it's right that playing is started and stopped from the same controller.

Being able to use either controller to control play will cause implementation difficulties. This is because Playing from the Media Player will always play forward at real time for that video.  But, playing from the TimeController will move forward at the pace & step-size configured for that scenario.

So, I'm going to close this.

PR #3324, opened 2018-10-22
explicitly update the play and pause icons in videoplayer

ayeshare:
Supports #3305


IanMayo:
Closing.  The recommended UI approach will introduce implementation complexity.

PR #3331, opened 2018-10-22
2888 m7 macro trials

IanMayo:
working trial


IanMayo:
Close this. The real work is in another branch.

Issue #3318, opened 2018-10-19
Narrative Viewer- Unable to modify the properties of recorded  narrative data


helenayele:
Trying to modify the recorded narrative data in properties view is not possible. 
### Steps to Reproduce:
1. Open boat1.rep file
2. Drag narrative.rep file to boat1.rep plot
3. you will get Narratives(19 items) added in outline view 
4. Open Narrative Viewer view 
5. In the Narrative Viewer , scroll down to the 121037 narrative entry, you will see TARGET ZIG DETECTED in the Entry field.
6.. Double-click on that row and its properties will show in the Properties view.
7. In the Properties view, change the color properties using custom color of  R = 29  Green = 56 and Blue - 96 and press enter
8. See that the color of that selected row  is Not changed in the narrative viewer , But instead it logs NullpointerException - 
[NPE_Narrative_Viewer.txt](https://github.com/debrief/debrief/files/2496152/NPE_Narrative_Viewer.txt)
9. After restarting Debrief I can't open Narrative viewer, it just crashed - 
https://gyazo.com/89f5b519318cca836aad78feb2d0916d

### Expected Result
It should change the color of the selected narrative data row after changing its color in the properties view. and No exception/error should be logged

ayeshare:
See issue #3071, they seem related.
I see one more issue: The narrative viewer has to be opened **_after_** loading narratives to display the narratives.

IanMayo:
Yes, let's remove the `Color` property from NarrativeEntryInfo, please.

IanMayo:
Fixed by @ayeshare  in a3c5cc4

PR #3332, opened 2018-10-23
3318 props modify exceptions

ayeshare:
supports #3318


org.mwc.cmap.legacy/src/MWC/TacticalData/NarrativeEntry.java
```diff
@@ -411,6 +410,7 @@ public final double rangeFrom(final MWC.GenericData.WorldLocation other)
   public void setColor(final Color color)
   {
     _color = color;
+    clearHash();
```
IanMayo: hey, that was clever :-)

PR #3340, opened 2018-10-26
3339 coordinate unit tests

### LINKED ISSUE #3339: Missing Unit Tests for CoordinateRecorder
I think it’s important that we have scenarios where they start at the same time, primary starts first, secondary starts first, etc.  In that case, it’s probably easier to generate the tracks by hand.
For the test, I think it’s better to do it headless.  So, we put some data into a Layers object.  Then we pass it to `CoordinateRecorder`, then we call `startStepping` once, then `newTime` a few times, then `stopStepping`, then we examine the contents of `_tracks` (or any other relevant data items).

saulhidalgoaular:
Fixes #3339


org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+    {
+      CoordinateRecorder recorder = getRecorder();
+
+      MWC.GenericData.HiResDate currentTime = new MWC.GenericData.HiResDate(
+          818748540000L);
+
+      final int AMOUNT_OF_STEPS = 3;
+      doIteration(recorder, currentTime, AMOUNT_OF_STEPS);
+
+      final java.util.Map<String, Debrief.ReaderWriter.powerPoint.model.Track> track =
```
IanMayo: @saulhidalgoaular - I'm thinking out loud here.  I would prefer you to retrieve the first segment, then to perform a series of:
`assertEquals("correct course", 12.0f, segment.getCourse(), 0,0001`

This has the advantage that it will find the specific attribute that doesn't match, and it also removes the dependency on unrelated code : track-point generation, and the TrackPoint.equals() implementation.

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+          recorder._tracks;
+      checkTrackSize(track);
+      assertTrue(track.get(collingwood).getStepsToSkip() == 1);
+      assertTrue(track.get(collingwood).getSegments().size() == 2);
+      assertTrue("0064bd".equals(track.get(collingwood).getColorAsString()));
+      checkCollingwoodSegments(track);
+      assertTrue(track.get(nelson).getSegments().size() == 3);
```
IanMayo: Aah, there's an issue from an earlier release here. A `segment` is a set of track points.  I'm pretty sure the above method should be `track.get(nelson).getPoints().size() == 3`.  So, the method (probably getter and setter) needs to be renamed to refer to points, not segments.

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+      assertTrue("e01c3e".equals(track.get(nelson).getColorAsString()));
+      final Debrief.ReaderWriter.powerPoint.model.TrackPoint firstSegmentNelson =
+          new Debrief.ReaderWriter.powerPoint.model.TrackPoint();
+      firstSegmentNelson.setCourse(2.6267204f);
+      firstSegmentNelson.setLatitude(20.0f);
+      firstSegmentNelson.setLongitude(245.0f);
+      firstSegmentNelson.setSpeed(1.1252066f);
```
IanMayo: Another fix from a previous release that I missed. We never call `getSpeed()` or `getCourse()`.  We only use the location and time. So, I think we should remove `course` and `speed` from TrackPoint, please.

If at some point in the future we require them, then we can add them back.

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+    public CoordinateRecorder getRecorder()
+    {
+      final File testFile = new File(
+          "../org.mwc.cmap.combined.feature/root_installs/sample_data/offset_times.rep");
+
+      // ok, now try to read it in
+      final Layers _theLayers = new Layers();
+
+      // add the REP importer
+      MWC.Utilities.ReaderWriter.ImportManager.addImporter(
```
IanMayo: I _think_ we can start off further down the call chain here.  I _think_ we can call ImportReplay directly to load the data into the `Layers` object.  That would avoid the thread-related code that is necessary in the `BaseImportCaller` implementation.

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+      checkTrackSize(track);
+      assertTrue(track.get(collingwood).getStepsToSkip() == 1);
+      assertTrue(track.get(collingwood).getSegments().size() == 2);
+      assertTrue("0064bd".equals(track.get(collingwood).getColorAsString()));
+      checkCollingwoodSegments(track);
+      assertTrue(track.get(nelson).getSegments().size() == 3);
+      assertTrue("e01c3e".equals(track.get(nelson).getColorAsString()));
+      final Debrief.ReaderWriter.powerPoint.model.TrackPoint firstSegmentNelson =
+          new Debrief.ReaderWriter.powerPoint.model.TrackPoint();
+      firstSegmentNelson.setCourse(2.6267204f);
```
IanMayo: since we never modify the member fields in TrackPoint, I think it would be "neater" in engineering terms if we had them as final fields, and set them once in the constructor.

Note: In a separate comment I've asked that we move away from creating a modified TrackPoint to use in this test, I'm just adding the comment against an instance of `TrackSegment.setCourse()`

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
```diff
@@ -144,14 +149,297 @@ protected void showMessageDialog(final String message)
...
+      firstSegmentNelson.setSpeed(1.1252066f);
+      firstSegmentNelson.setTime(new Date(818748540000L));
+      firstSegmentNelson.setFormattedTime("120609");
+      assertEquals(firstSegmentNelson, track.get(nelson).getSegments().get(0));
+      final Debrief.ReaderWriter.powerPoint.model.TrackPoint secondSegmentNelson =
+          new Debrief.ReaderWriter.powerPoint.model.TrackPoint();
+      secondSegmentNelson.setCourse(2.6162486f);
+      secondSegmentNelson.setLatitude(14.0f);
+      secondSegmentNelson.setLongitude(250.0f);
+      secondSegmentNelson.setSpeed(1.1252066f);
```
IanMayo: @saulhidalgoaular - I'm not a fan of all of the magic numbers in the testing, and am worried that they may become a maintenance burden.  

Mind you, you could argue that if these tests start to fail then it could be an early warning of an issue in some unrelated area.   I don't know which way to swing on that, happy to discuss :-)

Is there a simpler approach?  For this test, where we're validating that we handle tracks starting at different times, I _guess_ that we'll be confident things are working if the tracks have the correct number of children.

org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/recorders/CoordinateRecorder.java
IanMayo: @saulhidalgoaular - sorry this is dragging on.  There's a subtle improvement to the way you've structured your tests.

If you use:
````
assertEquals("correct number of skipped steps", 1, track.get(collingwood),getStepsToSkip());
````
Then, when the test fails we get an explanation of what went wrong, plus it says "expected to receive `1`, but received `5`".  This makes it slightly quicker to diagnose what went wrong.

saulhidalgoaular: oh Sir. I am sorry. I thought you meant about the attributes of the Track. I ignored it because you said below to remove it. I am doing now the changes you mentioned,

IanMayo:
Hmm, the tests are failing for me, on my desktop:
image

saulhidalgoaular:
> Hmm, the tests are failing for me, on my desktop:
> image
> 
> Aah, note: that's when I run just the test. It runs ok from AllTests in `org.mwc.debrief.test2`
> 
> Would you mind investigating?

oh ... I see. ImportThis needed to be initialized!!!!

saulhidalgoaular:
> Hmm, the tests are failing for me, on my desktop:
> image
> 
> Aah, note: that's when I run just the test. It runs ok from AllTests in `org.mwc.debrief.test2`
> 
> Would you mind investigating?

Hmmm, now that I think, maybe it would be good to add a method that checks if there is any available importer. I am currently adding a new importing, even if it is already available. It doesn't add too much overhead, because it takes the first importer that can read the file. 

https://github.com/debrief/debrief/blob/028a2a07def0380122ef64c77e4c7f441af17b05/org.mwc.cmap.legacy/src/MWC/Utilities/ReaderWriter/ImportManager.java#L337

IanMayo:
@saulhidalgoaular - tests now working fine.  Ok for me to merge?

saulhidalgoaular:
> @saulhidalgoaular - tests now working fine. Ok for me to merge?

Yes. It is ready to be merged!!

Issue #3137, opened 2018-09-20
M28.1 Investigate P2 repo build


IanMayo:
A customer wants to trial using P2 updates.

I _think_ our Maven build includes a P2 repo.

We need to:
* run local build `mvn clean verify`
* navigate into `org.mwc.debrief.product\target`
* take copy of `repository`
* download an older version of Debrief (say, 3 versions ago)
* open Debrief, register repository folder as update site
* see if it will update to include the new versions

If all the above works, then we should include a maven step to rename the repository.zip, then a Travis step to copy the repository to the release.

ayeshare:
these instructions didnt work.

ayeshare:
It turns out my workspace had the same version as an older release version I tried.

When I used the release 4.0.431 and checked for updates with the same steps above, I could install and restart debrief.  On checking about version it had upgraded to 4.0.440 which was the version in my workspace

ayeshare:
I made the following changes:

- added org.mwc.site/pom.xml to /pom.xml
-   to org.mwc.site/pom.xml, added step to archive repository.

Under org.mwc.site/target and file site-0.0.1-Snapshot.zip is created.

IanMayo:
Ok, the only remaining step is to move the repo to the release

IanMayo:
@ayeshare - could you keep this on your ToDo list?  We need an `mvn` command to rename `site-0.0.1-Snapshot.zip` to `Debrief_UpdateSite.zip`

I'm pretty sure there's an example of renaming (well, moving) a file in the `pom.xml` in `org.mwc.product`.  But, I don't know how to structure it correctly in Maven-speak.

IanMayo:
Aah, once we've got p2 updating, we should modify the build to push the p2 repository to the `gh-pages` branch.  We can host an update site there, and Debrief testers/enthusiasts can update their local copies live from that one.

IanMayo:
Aah, something else that would be useful to learn:
**Can we specify the p2 update site from the command line**.  A user organisation could then modify the startup shortcut to include the P2 update site path. This will prevent users from having to manually add the site.

PR #3354, opened 2018-10-31
M28 1 p2updatesite

ayeshare:
supports #3137

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ deploy:
     tags: true
     all_branches: true
 - provider: pages
+  api_key:
+    secure: $GITHUB_TOKEN
   local-dir: "/home/travis/build/debrief/debrief/org.mwc.debrief.site/target/update-site"
   verbose: true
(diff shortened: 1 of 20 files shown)


IanMayo:
@ayeshare - I don't think the Travis deploy is right.   You've added a 'local-dir' parameter to the 'releases' deployment.

I'm pretty sure we have to add a 'pages' deployment to the `releases` provider.  

We need to switch it to being a multiple deployment like this:
https://docs.travis-ci.com/user/deployment/#deploying-to-multiple-providers

We also need to _Add_ the new `pages` deployment, like this:
https://docs.travis-ci.com/user/deployment/pages/

IanMayo:
No, the provider for the update-site should be a `pages` provider, as detailed in the link above.

It looks like you've successfully followed the structure for _multiple deploy providers_, but you also need to read the link on _GitHub Pages provider_.

IanMayo:
Closing. Superceded by other P2 work

PR #3374, opened 2018-11-05
Bug/3368 update os slicer

### LINKED ISSUE #3368: Bug - Ownship leg graph not being updated
At the foot of the Stacked Dots plot we have the Leg Slicing graphs.

The plot fails to update after:
* deleting cuts during O/S turn
* resolving ambiguity
* switching an ambiguous leg to the other side
* switching to a new TMA track

We know the underlying data is being updated, since the stacked dot plots get updated. It's just the leg plot that fails to update.

IanMayo:
Fixes #3368


IanMayo:
Close this. We're going to apply the fix to Master, then pull it into Develop.

PR #3376, opened 2018-11-05
3368 master

### LINKED ISSUE #3368: Bug - Ownship leg graph not being updated

IanMayo:
Fixes #3368


IanMayo:
No, close. We don't want to pull all these changes into Master.

Issue #3359, opened 2018-10-31
PPT Export issues when recording reaches end of time 


helenayele:
Video file that shows the issue : 
[PPT_Export_issue.zip](https://github.com/debrief/debrief/files/2533920/PPT_Export_issue.zip)

While recording the movement of tracks and exporting in PPT format, if we don't pause recording and it reaches its end time, export dialog will be shown. So sometimes, it  does show the dialog that prompts the user to choose the export location 2 times. This happens usually when it uses invalid name for the file name  and it logs the below exception  - java.nio.file.InvalidPathException: Illegal char <:> at index 72: C:\Users\helena\Downloads\debrief_video_Export\DebriefExport-95\12\12 11:16_17.PPTX
	at sun.nio.fs.WindowsPathParser.normalize(Unknown Source)
	at sun.nio.fs.WindowsPathParser.parse(Unknown Source)
	at sun.nio.fs.WindowsPathParser.parse(Unknown Source)
	at sun.nio.fs.WindowsPath.parse(Unknown Source)
	at sun.nio.fs.WindowsFileSystem.getPath(Unknown Source)
	at java.io.File.toPath(Unknown Source)
	at Debrief.ReaderWriter.powerPoint.PackPresentation.pack(PackPresentation.java:63)
	at Debrief.ReaderWriter.powerPoint.PlotTracks.createPptxFromTrackData(PlotTracks.java:685)
	at Debrief.ReaderWriter.powerPoint.PlotTracks.export(PlotTracks.java:872)
	at Debrief.GUI.Views.CoreCoordinateRecorder.exportFile(CoreCoordinateRecorder.java:337)
	at Debrief.GUI.Views.CoreCoordinateRecorder.stopStepping(CoreCoordinateRecorder.java:522)
	at org.mwc.cmap.TimeController.views.TimeController.stopRecording(TimeController.java:3160)
	at org.mwc.cmap.TimeController.views.TimeController.processClick(TimeController.java:2260)
	at org.mwc.cmap.TimeController.views.TimeController.onTime(TimeController.java:1990)
	at MWC.Utilities.Timer.Timer.fireTimerEvent(Timer.java:141)
	at MWC.Utilities.Timer.Timer$TimerThread.run(Timer.java:163)

IanMayo:
@Helen - we now have a fix for this in the `develop` branch. Would you mind re-testing, to see if you can still reproduce the issue?

[...]

PR #3382, opened 2018-11-06
3359 PPT Export issues when recording reaches end of time

saulhidalgoaular:
Fixes https://github.com/debrief/debrief/issues/3359


saulhidalgoaular:
Honestly, I don't see how
`final String tidyName = tidyString(startTime);`
still returns invalid characters. Bug is very hard to reproduce.

On this commit I am also checking that the filename given by the user is valid.

IanMayo:
@saulhidalgoaular - it may be a combination of time format & operating system.  In @helenayele's video the time format is something like `yyyy/mm/dd hh:mm:ss`.  `:` is illegal in MS-Windows, but it may be acceptable (or get silently fixed) on Linux.

Issue #3189, opened 2018-09-30
M3.21 Support single point tracks


IanMayo:
We need to check that the PowerPoint export handles single point tracks.

I guess the implementation would defer the appearance of the footprint until the track starts, then just leave the footprint & callout) at that position

saulhidalgoaular:
I have attached the pptx generated from the growing_track branch of ppt_renderer project. I _think_ it is Ok. I delayed the footprint animation a bit (interval duration), so, currently the marker moves to the position and then the footprint appears.

If you consider it better to have the marker is in initial position of the track, I can place it there.

[scenario_long_range.txt](https://github.com/debrief/debrief/files/2432162/scenario_long_range.txt)
[scenario_long_range_tem.pptx](https://github.com/debrief/debrief/files/2432163/scenario_long_range_tem.pptx)

Thanks.
Greetings.

IanMayo:
Good work @saulhidalgoaular 

Ok, I see a flaw in our data handling.  Whether we have tracks with a single point, or tracks with multiple points, we need to handle where one or more tracks isn't "live" at the start time.

The data structure needs to be modified to include a "delay" parameter (in milliseconds) for each track.  If this value is non-zero, we need to delay the start of that animation for that period.

This delay is the _real world_ time before the animation starts, not the elapsed time in _model time_.

The `delay` parameter will be assigned in the constructor for `powerPoint.model.Track`.  The value will be calculated using the number of time steps.  So, it will be `_times.size()` multiplied by `_worldModelIntervalMillis`.

saulhidalgoaular:
Good afternoon. I am checking old bugs/new features, and I see that it is very related with https://github.com/debrief/debrief/issues/3149 . But I still think the exporter must be modified too `org.mwc.debrief.core.loaders.JaxbGpxHelper` .

PR #3384, opened 2018-11-07
3189 Support single point tracks

saulhidalgoaular:
Fixes https://github.com/debrief/debrief/issues/3189


IanMayo:
This doesn't work for single-point tracks, but it works perfectly for delayed tracks.

Don't worry though - single point tracks currently get _special processing_ in Debrief, where they're treated as being permanently visible.

PR #3391, opened 2018-11-08
Don't show labels for all infills.

### LINKED ISSUE #3370: Bug - All labels showing for dynamic infill
When we generate a dynamic infill segment, we copy the formatting from the last position on the previous leg.

But, when we do this, we should not copy the `Label Visible` attribute.  We should not show labels by default for Dynamic Infills.

IanMayo:
Don't show labels for infills, even if the last point of the previous leg has a label showing

Fixes #3370


IanMayo:
Ignoring CodeFactor review.  Don't want to do heavy refactoring until our UI-Test coverage has increased.

Issue #3366, opened 2018-11-02
D28 Import BRT file


IanMayo:
As an ***analyst*** I wish to ***import BRT data-files*** in order to ***reduce the steps in reconstruction, and potentially overcome a third-party error in generating ambiguous bearings***.

As a by-product of another process, and analyst has a bearing file in BRT format.  If we allow this to be imported, we remove the analysis step of "Convert BRT file to Debrief format".  Sadly, in the current implementation of the third party tool that does this, some of the bearings get messed up.

So, adding this feature will both save time, and result in better quality data.

## Acceptance criteria
1. Drag/drop .BRT file into Debrief, wizard opens.  User answers wizard questions. Check those answers have been used in the import.
2. Drag/drop both Towed Array and Hull-Mounted datasets.  Observe how they're separately handled in Debrief. The Towed Array data should be ambiguous, with an array offset distance. The non towed-array data shouldn't have either.  Learn more about _Ambiguous Bearings_ in the [Debrief Glossary ](https://github.com/debrief/debrief/wiki/glossary)
3. Verify that towed array cuts get converted correctly to ambiguous cuts during the turn
4. Wizard correctly handles single track in plot, multiple tracks with one Blue, multiple tracks with non blue, multiple tracks with more than one blue.

## Data format
This format is CSV of time and relative bearing, like this:
````
1514444259.00000, 113.3
````
The time value is seconds since 1970, and the bearing is degrees.

The time for the value is believed to be 2017/12/12 06:57:39  (excuse typo if that's slightly out).

## Implementation
For Debrief, we need to convert the relative bearing to absolute, using ownship heading.  We also don't know if the data is ambiguous or not.  See discussion on relative bearing here: https://github.com/debrief/debrief/wiki/glossary

So, when the file is dropped into a plot, we do the following:
1. Find the track to add it to:
    * If there is only one track loaded, we use that one
    * if there is more than one track loaded, but there is a single one colored DebriefColors.Blue, we use that one
    * otherwise we popup a dialog of tracks, and ask which one to add it to.
2. Determine what kind of sensor it is:
    * Ask if it's from a TowedArray
    * If it is, then ask for the towed array offset (re-using the previously used value [* see **Note 1**])
3. Choose the color for the data (re-use previous value)
4. Choose the default sensor cut length (re-use previous value)
5. Use the filename as the sensor name
6. Add the sensor cuts to the sensor, using the course at the relevant time to determine actual (and, if necessary, ambiguous) bearing.

[...]

PR #3393, opened 2018-11-09
3366 BRT Import

saulhidalgoaular:
Supports #3366

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/BRT/BRTImporter.java
+++ b/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/BRT/BRTImporter.java
@@ -16,23 +16,23 @@ package Debrief.ReaderWriter.BRT;
 
 import java.awt.Color;
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
 
-import org.junit.Assert;
-
+import Debrief.Wrappers.SensorContactWrapper;
 import Debrief.Wrappers.SensorWrapper;
 import Debrief.Wrappers.TrackWrapper;
 import MWC.GUI.Editable;
 import MWC.GUI.Layer;
 import MWC.GUI.Layers;
 import MWC.GUI.Properties.DebriefColors;
+import MWC.GUI.Properties.LineStylePropertyEditor;
 import MWC.GUI.Tools.Action;
+import MWC.GenericData.HiResDate;
+import MWC.GenericData.WorldDistance;
 import MWC.TacticalData.NarrativeWrapper;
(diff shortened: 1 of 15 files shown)


org.mwc.debrief.legacy/src/Debrief/ReaderWriter/BRT/BRTImporter.java
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  private TrackWrapper findTrack()
+  {
+    // TODO
+    return null;
+  }
+
+  public static class ImportBRTAction implements Action
+  {
```
IanMayo: @saulhidalgoaular - I'm being picky here, but...

For these two member fields, we know they aren't going to change.  So we should mark them as final, and then the compiler will _look after us_ by ensuring the never get changed.  It's the same for their declaration as parameters in the constructor.

We do have the Debrief code-cleanup settings.  With these loaded, when you run **Source > Clean up**, it will apply final when it's obvious they don't get changed.

But, in my opinion doing it early means the compile is _looking after you_ while you're still developing it.

(more on the settings here: https://github.com/debrief/debrief/wiki/On-boarding#project-settings )

saulhidalgoaular: oh, I see. I just changed it!

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/BRT/BRTImporter.java
```diff
@@ -0,0 +1,324 @@
...
+import junit.framework.TestCase;
+
+public class BRTImporter
+{
+
+  public static class BRTImporterTest extends TestCase
+  {
+
+    static public final String TEST_ALL_TEST_TYPE = "UNIT";
+
```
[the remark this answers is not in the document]
saulhidalgoaular: oh, ok ok. Changing

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/BRT/BRTImporter.java
```diff
@@ -221,13 +273,14 @@ public static TrackWrapper findTrack(final TrackWrapper[] allTracks)
...
       return allTracks[0];
     }
-    int amountOfBlueTracks = 0, indexOfBlueTrack = 0;
+    int amountOfBlueTracks = 0;
+    int indexOfBlueTrack = 0;
     for (int i = 0; i < allTracks.length; i++)
     {
       if (DebriefColors.BLUE.equals(allTracks[i].getTrackColor()))
       {
         ++amountOfBlueTracks;
```
IanMayo: hey, @saulhidalgoaular - I was pleased to spot this next typo.  The tests were all passing, but that was because of the sequence of tracks in the test data 😃

saulhidalgoaular: Thanks for spotting it :) I will check it more carefully next time.

PR #3437, opened 2018-11-15
Fixed the animation issue

### LINKED ISSUE #3402: Multiple recording icons are displayed  while recording track positions
Issue  - 
https://gyazo.com/64fdf5e69fe835d11ed10654e73dfdd6
https://gyazo.com/c72284f658d811bba098c5a38a367026

I got those duplicate icons that shows recording state while recording.

### Steps to reproduce:
1. Open sample.dpf
2. select record button - then pause recording and cancel exporting
3. repeat step 2
4. Select Record  button.
5. While recording resize - expand the width of the timecontroller by dragging to right. this time you will see multiple recording state's icon

Then you will get those multiple recording states' icon.

Once you see those multiple icons, you can reproduce it always unless you restart debrief.

ayeshare:
Fixes #3402

### DIFF
--- a/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/AnimatedGif.java
+++ b/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/AnimatedGif.java
@@ -31,6 +31,7 @@ public class AnimatedGif
   private ImageData[] imageDataArray;
   private Label lbl;
   private boolean cancel;
+  private Rectangle bounds;
   public AnimatedGif(Label lbl,String filename) {
     this.lbl = lbl;
     try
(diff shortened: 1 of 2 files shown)


IanMayo:
@ayeshare - just to help my Java knowledge, could you explain (in very loose terms) what the problem was, and how you've fixed it?  (my threading knowledge is quite poor...)

ayeshare:
It took a long time for me to figure out the solution.  I figured out that every time there is a stop and start there were multiple animation threads running.  There was no way to stop these threads. (Thread stop works but it is deprecated).  Another issue was since so many threads were getting created, it could run out of resources.  
The solution was to pause the thread on stop recording and resume the thread on start recording.  We have just one thread doing the animation.  When you resize the bounds will change, so while resuming the thread, it should resume with the new bounds.  So passed in the bounds to the resume method.
animate() is called only the first time, then we just make calls to resume and cancel

IanMayo:
(Ian nods politely, as if he fully understands what @ayeshare was saying)

I'm pleased we've got you on the team :-)

Issue #190, opened 2014-05-19
Display of solution course/speed still unclear


pecko:
_[Migrated from Assembla ticket [190](https://www.assembla.com/spaces/Debrief/tickets/190) - reported by @IanMayo on 2009-12-10 09:46:37]_

The display of the solution course/speed is still difficult to read when a track is being shifted.

Experiment with:
1. Moving the label to the outside of the last point on the track
2. Making course/speed properties in the the properties editor update in real-time.  There should be existing functionality to support this.

PR #3454, opened 2018-11-19
3453 Sensor Name in BRT Import

### LINKED ISSUE #3453: Import BRT file Sensor Data wizard misses sensor name entry page
Currently in the Import BRT Sensor Data wizard we didn't have an option to enter sensor name,  we’re using the filename as a default value. But we do have this option while importing sensor data in our previous implementation.

saulhidalgoaular:
Fixes #3453

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/loaders/BRTLoader.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/loaders/BRTLoader.java
@@ -30,8 +30,8 @@ import org.eclipse.jface.window.Window;
 import org.eclipse.jface.wizard.WizardDialog;
 import org.eclipse.swt.widgets.Display;
 import org.mwc.cmap.core.CorePlugin;
-import org.mwc.cmap.core.wizards.ImportBRTDialog;
 import org.mwc.debrief.core.DebriefPlugin;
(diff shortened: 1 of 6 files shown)


org.mwc.debrief.core/src/org/mwc/debrief/core/loaders/BRTLoader.java
```diff
@@ -155,8 +156,17 @@ public void run(final IProgressMonitor pm)
...
+          
+
+          // create sensor
+          String fileName = new File(_fileName).getName();
+          if (fileName.lastIndexOf('.') > 0)
+          {
+            fileName = fileName.substring(0, fileName.lastIndexOf('.'));
+          }
+          
           final ImportBRTDialog wizard = new ImportBRTDialog(theTrack,
```
IanMayo: @saulhidalgoaular - we've got two very similar variable names here, and I suspect that on line 190 you meant to refer to `_fileName`.

I suggest that `fileName` change to `defaultSensorName`.  Then it will be clear which is to be used in #190..

saulhidalgoaular: Ok. Changing now!!

PR #3463, opened 2018-11-20
fixed nullpointer exception by validating preferences and adding error

### LINKED ISSUE #3395: NullPointerException silently logged while using 'Export Track to CSV Text Format' ctx menu
### Steps to reproduce:

1. Open preference page
2. Select CSV Export Wizard ,and select 'Include 'Export Track to CSV Text Format...' check box
3. And don't select the path where those CSV files reside , leave it blank like this - [Image from Gyazo](https://gyazo.com/03d6244b87006d8f9b9dc84b9f701e52)
4. Open Sample.dpf file 
5. Right click on Nelson track in the editor > Select 'Export Track to CSV Text Format'  ctx menu
See that the below NPE is logged silently 
[NPE_While_exporting_CSV_export_location.txt](https://github.com/debrief/debrief/files/2565123/NPE_While_exporting_CSV_export_location.txt)
. It didn't bring any message to the user. And the user has no any idea what is happening here.

### Expected Result:
Debrief should tell for its users what is happening here. Need to inform the user he/she didn't select any path for the export location instead of failing silently.

ayeshare:
Fixes #3395

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/CSVExportPage3.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/CSVExportPage3.java
@@ -37,7 +37,11 @@ public class CSVExportPage3 extends CustomWizardPage
   private static final String CSV_EXPORT_PURPOSE = "CSV_EXPORT_purpose";
 
   private static final String CSV_EXPORT_EXPORT_FOLDER = "CSV_EXPORT_exportFolder";
-
+  
+  private static final String PURPOSE = "PURPOSE";
+  
+  private static final String DISTRIBUTION = "DISTRIBUTION";
+  
   public static final String PAGE_ID = "3. Release";
 
   private final DropdownProvider provider;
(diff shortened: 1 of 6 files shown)


org.mwc.debrief.core/src/org/mwc/debrief/core/wizards/CSVExportPage3.java
```diff
@@ -114,8 +118,16 @@ public void widgetSelected(final SelectionEvent e)
 
   private void addPurposeField(final Composite contents)
   {
```
IanMayo: @ayeshare - the validation methods for Purpose and Distribution are almost identical.  Could you have a go at refactoring out a common method that they both call?

I'm happy show you how I'd do it, if it seems challenging.

ayeshare: done.

IanMayo: Yes, I like it.

Issue #3471, opened 2018-11-21
M1.6.1.0 Create application skeleton


IanMayo:
Build the application, to match the UX definition.

Link to the UX definition:
https://xd.adobe.com/view/01609336-41de-4dc8-54e1-8fdb125718f4-7751/

ayeshare:
@helenayele can start testing this.  We made some changes from the original UX and will undergo further changes.

PR #3480, opened 2018-11-22
Added app skeleton for debrief-lite app

ayeshare:
Supports #3471


IanMayo:
Hello @ayeshare - I think we could do some further implementation in here, so it's ready to take the next steps of functionality.  Specifically, it would be good to have a main menu & main toolbar.  Yes, they will just contain placeholder items.  

We've got icons for Save (in org.mwc.debrief.core/icons/16) - but I don't see any in this repo. Maybe they're pulled in as part of the Ant build.

The value in having menu items and toolbar items is that we can copy the implementation when we start adding them for real.

IanMayo:
@ayeshare - I can see there are a couple of outstanding static analysis items to fix.  I'm sorry that the turnaround time for these checks is so slow.

Separate to that though, I'd appreciate you putting one or more dummy items into the toolbar and the menu - just to verify that the containers are there, and working.

PR #3520, opened 2018-11-29
3345 msi without npm

### LINKED ISSUE #3345: M28.3 Produce MSI build
During some Internet research @ayeshare found an npm tool that can create MSI installers on macos & linux:
https://github.com/mmckegg/msi-packager

This looks an ideal way of creating MSI installers. While @IanMayo saw them as a chore that couldn't be automated, this tool looks like a way in which we could potentially produce an `.mis` for every release.

This task is to get the build working on a local machine, then to research how to implement it on the travis servers.  We also need to introduce steps to extract the current version number from the debriefng.product file, and inject it into the config file (or apply it as a call parameter)

A quick bit of research suggests that `before_deploy` in Travis could be correct stage to perform the packaging, then we can include the msi in the`deploy` stage.

**Update 27/11/18:** We now have a set of linux scripts to run the linux msitools app. While this still doesn't provide the full capabilities of the MS Windows wix application, it has fewer missing capabilities than the npm msi-packager application.  I've attached a .zip file that contains the scripts.

We should update our msi-build branch to include these scripts.
[MSI_Build_scripts.zip](https://github.com/debrief/debrief/files/2620578/MSI_Build_scripts.zip)

I guess this task will include:
* [ ] Get packager running from command line
* [ ] Introduce ability to update version numbers in command line
* [ ] Transform script so it runs as part of local maven build
* [ ] Integrate script into Travis headless build.

### LINKED ISSUE #3501: M28.3 Produce MSI Installer
This will resolve the issue that the users have with installing the Ap.

IanMayo:
Fixes #3345 
Fixes #3501


IanMayo:
Dev has verified the MSI works, and that the wxs files don't need any improvement.

Issue #3421, opened 2018-11-13
M7.2.2 Produce code to expose object models - First Bucket of 8


Waynearendse:
Call the constructor and Write the Wrappers
Write a unit test for each wrapper
30 mins per object - check no exceptions is thrown. Testing diff parameters
Add the line for code completion to the wrappers 
For one Buckets of 8 = 4hours

**To include investigating why new object didn't appear in Outline View**

Code review 1 hour

saulhidalgoaular:
Methods ready for testing

Shapes:
- createEllipse
- createFurthestOnCircleShape
- createPolygon
- createRangeRingShape
- createRectangleShape
- createWheelShape

Core
- createFont

Narratives
- createNarrative

helenayele:
@saulhidalgoaular  , to create createFurthestOnCircleShape , we need to expose WorldSpeed, and I can't see it in the module explorer.

PR #3522, opened 2018-11-30
3421 object expose

saulhidalgoaular:
Supports #3421


org.mwc.cmap.tests/.project
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
IanMayo: @saulhidalgoaular - we don't want this file in the repo. You should delete the project from your Eclipse workspace.

saulhidalgoaular: ohhh... ok. Sorry. I am deleting it!!

PR #3525, opened 2018-11-30
App freeze on switching to new plot

### LINKED ISSUE #3524: Bug - app freeze on loading new data
We've got a show-stopping bug.

If we've got the Stacked Dots (bearing residuals) view open, when we open a new .rep file, the application freezes:
https://youtu.be/DY-_Ltxwyrg

@helenayele - could you verify this also happens for you?

IanMayo:
Fixes #3524 

Put updating the stacked dots into an async display thread


IanMayo:
Investigations so far make it seem this issue is much more fragile on MACOS.  It does seem related to the PartMonitor listeners in `BaseStackedDotsView`.

I'm unable to reproduce it so far on MS Windows.

IanMayo:
Close. Issue fixed by @G in other branch.

PR #3531, opened 2018-12-03
Bug/3526 dtg in preamble

### LINKED ISSUE #3526: Failing to import narratives from PDF
Analysts have received a narrative.  It's unlike those normally received, in this way:
* it's a single 1000+ page document, containing several hundred pages of narrative in the middle
* it's in PDF format, and is missing the hidden fields we rely on to give context to the data

So, produce a working mockup of this, then modify Debrief so that it can load the document.

Note: for a working `.doc` narrative, Debrief is able to reconstruct context for entries when the hidden fields are missing.  We may just have to extend this logic - particularly to include capturing the month and year indicators.

IanMayo:
Fixes #3526


IanMayo:
Close, we were trying to merge back into the wrong branch.

PR #3534, opened 2018-12-04
debrief-lite changes for loading .rep file and display tracks

unnivm:
Hi @IanMayo ,

I have made changes to the following files. Please verify them and let me know.

I will get the projection from geomap using the way you suggested.

### DIFF
--- a/org.mwc.debrief.lite/.classpath
+++ b/org.mwc.debrief.lite/.classpath
@@ -23,11 +23,7 @@
 		</attributes>
 	</classpathentry>
 	<classpathentry combineaccessrules="false" kind="src" path="/org.geotools"/>
-	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.jfreechart"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.legacy"/>
-	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.NaturalEarth"/>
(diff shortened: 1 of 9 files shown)


org.mwc.debrief.lite/.classpath
```diff
@@ -28,5 +28,6 @@
 	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.NaturalEarth"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.openmap"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.debrief.legacy"/>
+	<classpathentry combineaccessrules="false" kind="src" path="/org.mwc.cmap.gt2Plot"/>
```
IanMayo: @unnivm - we have to remove this dependency - we can't introduce the dependency on eclipse libs in gt2plot.

IanMayo: @unnivm - could you also explain why we depend on NaturalEarth and OpenMap?

thx.

unnivm: @IanMayo ,

Looks these two dependencies(NaturalEarth and OpenMap) are existing. But they are not needed. I have removed from the classpath.

But gt2plot is needed as we are using GtProjection (until now we are using it). This entry is added automatically when we add the Gt2plot project dependency.

What I will remove entries from classpath for gt2plot and make one more commit. But at the time of importing the project, we need to resolve this dependency manually.

Please let me know.

IanMayo:
Close this branch. We're representing this work in #3540

PR #3536, opened 2018-12-04
added separate ant build for building lite and invoked from maven build

### LINKED ISSUE #3523: M1.6.2.0.3 Automate build of new Debrief-Lite
The new Debrief-Lite builds and runs from Eclipse, but for our Continuous Integration (CI) we need it to run from the command line.

The CI runs via Maven, but building Debrief-Lite via Maven proved a challenge (because of the local dependencies, I believe).  So, we're going to build it in Ant, but trigger that Ant build from Maven.

ayeshare:
Fixes #3523


org.mwc.debrief.core/plugin.xml
```diff
@@ -1304,9 +1304,5 @@
             class="org.mwc.debrief.core.splashHandlers.ExtensibleSplashHandler"
             id="org.mwc.debrief.core.splashHandlers.extensible">
       </splashHandler>
```
IanMayo: hmm, @ayeshare - isn't this removing the splash screen from the main Debrief app?

org.mwc.debrief.lite/.project
```diff
@@ -19,5 +29,6 @@
 	<natures>
 		<nature>org.eclipse.jdt.core.javanature</nature>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
```
IanMayo: Hey, @ayeshare - I don't really know anything about natures. Can you explain what adding the PluginNature does?

org.mwc.debrief.core/plugin.xml
ayeshare: oh i have accidentally committed this file.  let me undo it

org.mwc.debrief.lite/.project
ayeshare: This converts a regular java project to be a plugin project.  Doing this is required to be able to run this from the top level pom file.  The pom file will build the project as if it is an eclipse plugin and during the compile phase, the pom file also makes a call to the debrief.lite's ant script

org.mwc.debrief.lite/build.xml
```diff
@@ -1,109 +1,146 @@
...
+				<include name="images"/>
+		 	</fileset>
+		</copy>
+		<jar destfile="${temp.folder}/debrief_lite.jar" basedir="${temp.folder}/debrief_lite.jar.bin">
+			<manifest>
+				<attribute name="Main-Class" value="org.mwc.debrief.lite.DebriefLiteApp"/>
+				<attribute name="Implementation-Title" value="Debrief Lite"/>
+				<attribute name="Implementation-Version" value="${product.version}"/>
+				<attribute name="Implementation-Vendor" value="DeepBlue C Ltd"/>
+				<attribute name="Class-Path"
```
IanMayo: @ayeshare - we're pulling in eclipse SWT libraries in this classpath.  I'm really hoping we can do without them. 

Separate to that, if possible I'd like to avoid this "magic string" of jar names.  I presume we already define a .classpath file somewhere for Debrief-Lite. Can we pull in the contents of that file into this variable?

(I'm hoping that by managing the dependencies from something like .classpath, Eclipse will help us with compiler warnings and/or refactorings)

ayeshare: If you are referring to the libs in .classpath file: the .classpath file is only used by developers in the workbench.
I really wish we could remove the magic string of jar names.  One jar was doing that for us.  We can put that in a manifest file and include the manifest file, which is the other option.

IanMayo: We're speaking beyond what I really understand, but I like the sound of the manifest file. Could you try it?

IanMayo:
Hey, @ayeshare -  I'm pleased you remember how to do this legacy build stuff.

org.mwc.debrief.lite/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,5 +1,6 @@
 eclipse.preferences.version=1
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
```
IanMayo: @ayeshare - for my self-teaching, could you explain why we've changed this setting?

IanMayo: Aah, I see. I guess Eclipse just decided to insert it.

PR #3568, opened 2018-12-13
Hotfix/defer accessing editor registry

IanMayo:


IanMayo:
Forget about this, we need a more robust fix

PR #3576, opened 2018-12-17
Allow ellipse end date to be specified in REP format

### LINKED ISSUE #3575: Not possible to set Ellipse end-date in REP format
In the REP format, we can only specify the start date for Ellipses.

It's a frequent requirement to also need the end date.

We should introduce an Ellipse2 format, to allow this.

IanMayo:
Fixes #3575


IanMayo:
Can't merge this, it's failing because it times out trying to download RCPTT, for automated UI Tests.

PR #3584, opened 2018-12-19
remove property-editor dependency on chart object

IanMayo:
fairly hefty package of work to remove the Chart dependency from the property editor framework.


IanMayo:
@ayeshare - have you used Debrief legacy?

If you have, would you mind doing a bit of exploratory testing to verify that I haven't broken it?

I've had a play, and everything seems to work ok.

ayeshare:
I did some tests to check if property editor is working fine on the debrief app and it looks good.  On applying the changes to property editor, the chart is getting updated.

PR #3602, opened 2019-01-09
Debrief scripting cookbook

IanMayo:
Introduce Debrief Scripting Cookbook to reference section of user guide.


IanMayo:
@helenayele - Lakshi has moved the cookbook code into the master user manual. But, both she & I have made formatting changes.  So, the cookbook scripts could do with testing.  They're all in Chapter 17 of this pdf:
[ng_help.pdf](https://github.com/debrief/debrief/files/2932358/ng_help.pdf)

Issue #3603, opened 2019-01-10
Import more elements, update test (Import Replay)


saulhidalgoaular:
We now import more elements, update test 
Recovering this commit from the wrong branch.
https://github.com/debrief/debrief/commit/6b9ec4b84f018fc1032bed570d8c39befd182a9a

helenayele:
@saulhidalgoaular , Would you please some steps to test this feature?

PR #3604, opened 2019-01-10
3603 We now import more elements, update test

saulhidalgoaular:
Solves #3603


IanMayo:
Close. Fixed in https://github.com/debrief/debrief/pull/3606

Issue #3630, opened 2019-01-17
M1.6.4.6 Range Slider control


IanMayo:
In order to let the analyst filter the visible time period, we need a range slider control to use in Debrief-Lite

Full Debrief has it's own custom SWT control.  There are _some_ available in Swing, but we need to verify that:
* [ ] They render correctly in our Radiance UI theme
* [ ] We can insert them into our JRibbon control bar
* [ ] They safely handle the _UI simplification_ that JRibbon performs as the window is resized.

Here's one example:
https://ernienotes.wordpress.com/2010/12/27/creating-a-java-swing-range-slider/

Googling indicates that there is also one in the SwingX library from Sun, but I can't find where to download SwingX.

Aaah, there's another in the Open Source version of JIDE here:
http://www.jidesoft.com/products/oss.htm

IanMayo:
@saulhidalgoaular  has this working.

helenayele:
@saulhidalgoaular  , would you tell me how to see this feature on debrief-lite?

IanMayo:
The control is in the Time ribbon bar in Debrief-Lite.

PR #3637, opened 2019-01-19
3630 M1.6.4.6 Range Slider control

saulhidalgoaular:
Supports #3630 

It still needs to be integrated with the map, but it works now storing the max and min value as seconds from the epoch (time as UTC).

image
image


IanMayo:
Hey, I'm particularly pleased it works in the collapsed mode.

saulhidalgoaular:
> Hey, I'm particularly pleased it works in the collapsed mode.

Awesome. Thanks!!

PR #3642, opened 2019-01-20
3638 Refactoring File, Insert, View and Time Controller.

### LINKED ISSUE #3638: Refactor Debrief Ribbon into one class per tab
@ayeshare 's ribbon class currently builds the whole bar.  But, I know the Time controls are going to have lots of functionality.  

So, I think now's the time to separate the bar out into several classes - one per tab (File, View, Insert, Time).

@saulhidalgoaular , _if_ you work over the weekend, could you take on that task?

It will reduce code conflicts.

saulhidalgoaular:
Fix #3638


IanMayo:
@saulhidalgoaular - I think we can have some more refactoring to the DebriefRibbon class.

`addMenus` is called immediately after the `DebriefRibbon` constructor from `DebriefLiteApp`.  

image

So, it looks like the `addMenus` functionality could be pulled back into the constructor.  That would mean we wouldn't need the member variables in the `DebriefRibbon` class.  

Further to that, I think we could pass in the `JRibbon` element instead of the `JRibbonFrame`.

IanMayo:
@saulhidalgoaular - I'm going to merge this branch back into ribbon, to @ayeshare can carry on working in her menu tab.

I'd like you to resolve the above minor refactoring issue (`DebriefRibbon` constructor) in a separate branch/PR, please.

PR #3669, opened 2019-01-25
3504 splashscreen liteapp

### LINKED ISSUE #3504: M1.6.4.0 Splash Screen
## Development 
Copy the splash screen image used for normal Debrief.  Find a way (if possible) to include a build version.

Where do you get the version from? (tech challenge - research)

Build properties file

## Testing
Check the splash screen is shown while debrif-lite is launched
Check the version of debrief-lite shown in the splash screen
change the version in the POM file and see that the screen updates with the correct version number ? is this possible?

Note:
_This feature will add value to this Sprint, and reduce risk for future phases_

ayeshare:
Fixes #3504


IanMayo:
The splash icon has been added to `debrief.core`.  Shouldn't it go into the `debrief.lite` plugin?

org.mwc.debrief.lite/build.xml
```diff
@@ -101,6 +101,7 @@
 			<fileset dir="../org.geotools/libs" includes="*.jar"/>
 			<fileset dir="" includes="cmap_legacy.jar"/>
 			<fileset dir="../org.mwc.debrief.legacy" includes="debrief_legacy.jar"/>
+			<fileset dir="libs" includes="*.jar"/>
```
IanMayo: @ayeshare - were we missing out on some libs in the build? Or, should this PR include another jar?

IanMayo:
@ayeshare - this isn't running from the IDE, I get this NPE:
image

It's failing because `SplashScreen.getSplashScreen()` is returning null here:
image

ayeshare:
> The splash icon has been added to `debrief.core`. Shouldn't it go into the `debrief.lite` plugin?

That was already existing, it is an old one.  I used the same image and put it in debrief.lite images folder

org.mwc.debrief.lite/build.xml
ayeshare: We had not included the jars in the dist, it was already added to the build for compilation.
My change was to add the jars to the dist and also in the manifest classpath

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DisplaySplash.java
```diff
@@ -0,0 +1,138 @@
...
+    g.fillRect(0,140,300,80);
+    g.setPaintMode();
+    g.setColor(Color.RED);
+    g.drawString(message, 20, 170);
+  }
+  public DisplaySplash(int numTasks) {
+    super();
+    this.numTasks=numTasks;
+
+    splash = SplashScreen.getSplashScreen();
```
IanMayo: @ayeshare - I'm still unable to run this from the IDE.  This method returns null, so `splash` has null value.  We then get an NPE each time we refer to it.

IanMayo:
@ayeshare - this is still failing for me.   I've just written a long description, suggesting fixes.  But then I googled it, and now understand. `SplashScreen.getSplashScreen()` will return null if the image hasn't been provided on the command line or in the manifest.  I've done this in my Eclipse Run Confi - and it works fine.

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DisplaySplash.java
IanMayo: Here's how I passed the argument:
image

IanMayo:
@ayeshare - I'm keen to get this merged, so I'm fixing the static analysis checks

PR #3671, opened 2019-01-25
3651 long range satc

### LINKED ISSUE #3651: Improve SATC performance in long range encounter
An analyst & I were considering a dataset. 

The dataset had the following characteristics:
* 950 resolved TA cuts, working to 10 degree error
* about 36 hours long
* 50 miles long at tallest point
* routes like image below

We ran through it in SATC, and added these contributions
* all bearings
* 6 knots for whole period
* roughly sliced about 20 legs

We asked SATC to run calculate, with these shortcomings:
* took 4 minutes to run (though analyst didn't mind this delay)
* solution was unachievable
* for some legs, the direction solution was inverted vertically (SW instead of NW).  These solutions matched the bearings, but required extremely fast infills (70kts) that was just unachievable.  If these invalid infills had been blocked, the legs would probably have faced the right way.

Routes like this:
image

IanMayo:
Fixes #3651 

Track improvements to long range SATC


IanMayo:
`DO NOT MERGE THIS`

Superceded in #3677

PR #3675, opened 2019-01-30
changes to insert tab

### LINKED ISSUE #3668: D-Lite Insert ribbon layout
We have a UI for the insert menu, which includes larger icons for "Select layer" and "Label", since the first is important, and since the second is the most commonly used annotation type.

We also have layout priorities for when the application shrinks:
https://projects.invisionapp.com/share/M8PY9ZONZHC#/screens/342699156

## Update 5th Feb

Here is the current version of the `Insert` tab:

image

I believe we have the following outstanding changes:
* [ ] Icons for Absolute & Relative Time Markers
* [ ] Vertical separator between the time markers & the grid icon
* [ ] Ensure backend for all buttons is implemented (to actually create the shapes/features)
* [ ] _if possible_, make the icons for `Select target layer` and `Label` larger, since they're the most commonly used
* [ ] Remove the `Chart Lib` icon, it's not being implemented until a later phase.
* [ ] Use correct icons for `Coastline` and `Natural Earth`.

ayeshare:
fixes #3668

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/menu/DebriefRibbonInsert.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/menu/DebriefRibbonInsert.java
@@ -99,17 +99,17 @@ public class DebriefRibbonInsert
     final FlamingoCommand coastlineCmd = MenuUtils.addCommand("Coastline",
         "images/16/coast.png", new CreateCoast(_toolParent, _theProperties,
             decs, _theLayers, bounds), referenceDataMenu,
-        RibbonElementPriority.MEDIUM);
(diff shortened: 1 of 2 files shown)


IanMayo:
Thanks for that. I've got some more changes, please:
insertpanel

* 1 & 2 - could you use the Debrief icons for Chart Lib & Natural Earth?
* 3 - I don't think we need the `Core` label
* 4 - Could you replace this label with `Select target layer`
* 5 - Could you swap these two around, so that the toggle comes before the label

Oh, and if it's possible, could you remove the separator between the two larger icons (4&5) and the block of smaller icons?

IanMayo:
The changes for this have been included in #3693

PR #3699, opened 2019-02-06
enable rcpt tests on ease branch

theanuradha:


IanMayo:
that was a very clever fix @theanuradha  :-)

Do you want to use this PR to fix the tests?

(Merging it now will make all EASE_DEV PRs fail, because the tests will fail)

Issue #3488, opened 2018-11-22
D31 Produce Multistatic dataset


IanMayo:
An analyst is going to be doing some research into multi-static sensor measurements.

The analyst has requested that exercise participants follow a run-plan, to produce a sample dataset.  But, it may be some time before this opportunity arises.

Here is a such a run plan:
![larger multi-statics run trial](https://user-images.githubusercontent.com/1108513/48898582-03e08500-ee45-11e8-8a46-c30ee60dc04f.png)

So, they would benefit from a fictional sample dataset, from which they can start to consider the processing chain to give them their required end product.

The provision of fictional dataset of representative proportions would allow them to start making progress.

This task is to use a simulator to re-create the run-plan, then to output the data to text file, both as Debrief `.rep` files, and to `csv` data-files.  The simulator should produced measured bearing, frequency, and received signal strength at the buoys.  The signal strength should take into account a fictional aspect dependent acoustic absorption factor.

Once the data is produced, the analyst can be guided through how to use Debrief to view various geometric angles between the entities.

This baseline capability _may_ lead to the identification of Debrief software changes to either further reveal participant dynamics, or to perform some calculation(s) on the loaded elements.

The task is estimated as 3 days of effort.

IanMayo:
Demo of first implementation:
https://youtu.be/Sa9XCzDk528

PR #3711, opened 2019-02-06
D31 multistatics

IanMayo:
First phase of #3488


IanMayo:
Ok merge that changes, so they're safe

PR #3713, opened 2019-02-07
3694 save plot

### LINKED ISSUE #3694: M2.1 File Save for D-Lite
As an ***analyst*** I wish to be able to ***save the current plot*** in order to ***retain the changes and new features that I have added to the plot***

As an ***analyst*** I wish to be able to ***save the current plot with a different name*** in order to ***develop two different versions of a plot***

## Implementation
The processing of the legacy Debrief application should be suitable as a template for this, with the actual export being performed in `SavePlotAsXML`.  

(note: I've just had a look, and it looks like DebriefXMLReaderWriter doesn't currently have export functionality, it's commented out.  I don't know why this is. _Maybe_ we'll have to copy some `exportThis` functionality from `DebriefEclipseXMLReaderWriter`)

(second note: in order for `.dpf` files to be compatible across both apps, we may have to artificially create the `<plot>` and '<session>` XML elements that wrap the layers object.  I suggest this is done within the Debrief object model, rather than via XML manipulation - in order that we are confident we're using the same XML import/export across both apps).

A file dialog should force the user to choose a folder to export the .dpf file to.  For a default filename, we _could_ start with the filename prefix of the first file loaded in the current session.   (in that case, we'll have to store it).

The Save dialog should warn if the user tries to over-write an existing file, with the usual replace/cancel dialog.

We'll need both `Save` and `Save As` functionality.

### XML compatibility
* Load `shapes.rep` in full Debrief, save to `shapes_full.dpf`
* Load `shapes.rep` in D-Lite, save to `shapes_lite.dpf`
* Compare the resulting files. They won't be identical, since Debrief stores more. But, where both versions contains the same elements (the `Layers`) they should be identical.
* Load each .dpf file into the other app, they should load without error

### UI
* Perform `Save`, verify correct name is offered as default.  Complete save, verify filename is correct.
* Perform `Save`, verify correct name is offered as default.  Modify name. Complete save, verify filename is correctly modified.
* Re-do above steps, verifying it correctly handles filename already exists.
* Re-do above steps for `Save As` functionality.

IanMayo:
Fixes #3694


IanMayo:
@ayeshare - you're welcome to take over this branch/PR.  It's saving to a hard-coded filename, so we just need the UI logic.

IanMayo:
Hello @ayeshare - had a play with this PR:
* [ ] I loaded boat1.rep, then pressed "Save".  It saved the xml content into `boat1.rep` - it should force the file to be renamed as `boat1.dpf`.  This left me with a failing `boat1.rep` - the Replay importer fell over when it found all that XML code.
* [ ] when I tried `Save As`, it didn't force the file to take a `.dpf` suffix, so I did the save without the suffix.

ayeshare:
> Hello @ayeshare - had a play with this PR:
> 
> * [ ]  I loaded boat1.rep, then pressed "Save".  It saved the xml content into `boat1.rep` - it should force the file to be renamed as `boat1.dpf`.  This left me with a failing `boat1.rep` - the Replay importer fell over when it found all that XML code.
> * [ ]  when I tried `Save As`, it didn't force the file to take a `.dpf` suffix, so I did the save without the suffix.

Fixed both the above issues

Issue #3681, opened 2019-01-31
D-Lite tie in Time Controller


IanMayo:
We've got our Time Controller UI, we now need to tie the controls into the Debrief plot.

This is the first challenge.  The next challenge is to tie the Normal/Snail painters into the plot, too - #3682

IanMayo:
Still have to do:
* [ ] update time filter UI to match data
* [ ] implement filter to time
* [ ] correctly display the time label (Green on black)

IanMayo:
@saulhidalgoaular - I'm assigning this to you, it contains some small cosmetic changes to finish of the Time Ribbon.

saulhidalgoaular:
Ok. Ok. Working on it.

helenayele:
And also Clicking on properties button should open the property view

saulhidalgoaular:
I think we can close this issue. Branch with the features has been merged to Snail Painter. Final fix has been merged to Develop and verified by Ian.

helenayele:
Verified

PR #3690, opened 2019-02-04
3625 snail painter

IanMayo:
Fixes #3682

PR #3727, opened 2019-02-08
3681 - D-Lite tie in Time Controller

saulhidalgoaular:
Support #3681 
format Current Time label according to selected drop-down format.
It still needs to be updated when the dropdown changes


IanMayo:
@saulhidalgoaular  - we also need to reset the time period when we load from a `.dpf` file.

We can use the same `layers.getTimePeriod()` call that you use for `handleImportRep()`

Oh, and I've also fixed a merge conflict :-)

IanMayo:
@saulhidalgoaular - is this PR ready to be checked/merged?

saulhidalgoaular:
> @saulhidalgoaular - is this PR ready to be checked/merged?
> 
> Aah, hold on. It's been merged into #3690 , right?

Yes, I have been merging it to snail painters, because some of this changes are needed there.

PR #3752, opened 2019-02-12
Single point tma (to correct branch)

### LINKED ISSUE #3726: Stacked dots failing with single point leg
An analysis is slicing a manual TMA leg.

He identified a really short leg, containing just one sensor fix.

Once the solution has this one-point leg, and he drags a different leg, the stacked dots plot fails to update.  An array bounds exceeded exception is thrown.

Investigate handling of single-point solutions.  They're uncommon, but I'd like to support them.   The stacked dots plot should certainly continue to work,

IanMayo:
Fixes #3726


IanMayo:
Cancel this - it starts from the wrong branch

PR #3761, opened 2019-02-14
added error message if no layer name was provided

### LINKED ISSUE #3740: Don't allow a user to create a layer with blank name
Currently a user can create a layer without providing a name.
It looks like this:
[Image from Gyazo](https://gyazo.com/f56e5377a2f78471fa9badd776581817)
Even if it has no any effect on the application, creating a layer without a name has no any value for the user.
So prohibit creating a layer without a name.

Note:
In full Debrief, user can't create a layer without a name .

ayeshare:
Fixes #3740


IanMayo:
@helenayele - could you check this works as you expect?

PR #3769, opened 2019-02-18
extend progress api and remove swt swing dialogs mixup

theanuradha:


theanuradha:
@IanMayo I have removed Swing and SWT event thread mixup and extend progress api so we can show progress from SWT than popup a swing dialog

IanMayo:
I think we **SHOULD NOT MERGE** this PR, I think we should merge the one based upon 'Master': #3770

PR #3770, opened 2019-02-18
extend progress api and remove swt swing dialogs mixup

theanuradha:


theanuradha:
> Hmm, from looking at the code, the _tmpFrame variable never gets assigned. Am I missing something?

as I move finalise to resource close method it might get call before **_tmpFrame** init so needed NULL check

IanMayo:
No, I can't see any code that assigns to this private member variable. Shouldn't we just delete it?

theanuradha:
> No, I can't see any code that assigns to this private member variable. Shouldn't we just delete it?

Ah sorry yes I have clean that up

PR #3802, opened 2019-02-28
added null checks to fix

### LINKED ISSUE #3788: Need to provide dpf file extension while saving rep file
When we save rep file in D-Lite, if we miss to give .dpf extension, the file is saved as text file even if dpf file is selected in **Files of Type** combo. like this - [Image from Gyazo](https://gyazo.com/eb9d01c9ce27c203b085ced39e4982f2)
If I didn't provide the .dpf after the file name, it automatically saved as text file.

ayeshare:
Fixes #3788


IanMayo:
Works fine, sorry I didn't catch this yesterday.

PR #3805, opened 2019-03-01
implementation of new plot

### LINKED ISSUE #3803: Implement New action in the file ribbon tab
Clicking on the new action should close existing file that is open and reset all the existing variables that has been cached for the file that was already open.

ayeshare:
Fixes #3803

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
+++ b/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
@@ -1657,7 +1657,8 @@ public class ImportReplay extends PlainImporterBase
     }
   }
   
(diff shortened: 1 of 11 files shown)


org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportReplay.java
```diff
@@ -1663,7 +1656,24 @@ private final void importRep(final String fName, final InputStream is,
       }
     }
   }
-
+  
+  private void handleException(final Exception e, final int lineCount,
+      final String line, final String fName, final String message)
+  {
+    // produce the error message
+    SwingUtilities.invokeLater(new Runnable()
```
IanMayo: @ayeshare - I'm nervous about this, since we use ImportReplay in Full Debrief.  I don't thing SWT apps like calling SwingUtilities.

PR #3828, opened 2019-03-11
removed toggle button and replaced with select layer dropdown

### LINKED ISSUE #3820: implement new design for layers popup dialog
https://xd.adobe.com/view/b84612dd-e964-4322-699d-90574b10b452-ce77/screen/9e7e3619-b626-4a9b-a0aa-13956e655dbb/Add-shapes-layer-selector

The link above has the UX design for this requirement.
Select Layer drop down should list the existing layers. Two special options should be available viz, Active Layer and New Layer.
Selecting Active Layer should use the layer that is selected in the outline view
New layer should allow a new layer to be created.

When user clicks on the drop down, the select layer drop down should be populated.

ayeshare:
Fixes #3820

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
@@ -219,8 +219,8 @@ public final class CreateLabel extends CoreCreateShape
           // works for debrief-legacy
           // user cancelled.
           JOptionPane.showMessageDialog(null,
-              "A layer can only be created if a name is provided. "
-                  + "The shape has not been created", "Error",
+              "An item can only be created if a parent layer is specified. "
+                  + "The item has not been created", "Error",
               JOptionPane.ERROR_MESSAGE);
           res = null;          
         }
--- a/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
@@ -160,15 +160,6 @@ public final class CreateLabel extends CoreCreateShape
         if(layerToAddTo!=null) {
           theLayer = _theData.findLayer(layerToAddTo);
         }
-        if(theLayer!=null) {
-          if(theLayer != null) {
-            res =  new CreateLabelAction(_thePanel,
-                theLayer,
-                theWrapper,
-                _theData);
-
-          }
-        }
       }
       if(layerToAddTo!=null) {
         theLayer = _theData.findLayer(layerToAddTo);
@@ -190,11 +181,15 @@ public final class CreateLabel extends CoreCreateShape
           theLayer.setName("Misc");
           _theData.addThisLayer(theLayer);
         }
-        res = new CreateLabelAction(_thePanel,
+      }
+      if(theLayer!=null) {
+        res =  new CreateLabelAction(_thePanel,
             theLayer,
             theWrapper,
             _theData);
+
       }
+  
     }
     else
     {
--- a/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
@@ -79,6 +79,7 @@ import Debrief.Wrappers.LabelWrapper;
 import MWC.GUI.BaseLayer;
 import MWC.GUI.Layer;
 import MWC.GUI.Layers;
+import MWC.GUI.PlainWrapper;
 import MWC.GUI.ToolParent;
 import MWC.GUI.Properties.PropertiesPanel;
 import MWC.GUI.Tools.Action;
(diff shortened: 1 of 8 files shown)


IanMayo:
@ayeshare - I can't see a copy right of the design now, but I think two features are missing from the new target layer ribbon area:
* static image label showing stack of layers
* a label at the foot of the ribbon area, it may have been "Select Target Layer"?

helenayele:
Yes , Manually select layers button  is removed. I accept that. I think the design doc should be  modified.
But  the ribbon  has not "Active Layer"  title. 
The other issue is when the user selects 'User Selected Layer' , it automatically brings "Adding a new features dialog" to choose the layer. After I select it and I try to insert the shape. But the shape is not inserted in the selected layer. Neither it throws exception.
[Image from Gyazo](https://gyazo.com/0e30da1cb652f18cba114652b2883b9f)

ayeshare:
@helenayele did you get all the changes from the branch? It just works fine here.
I have fixed the ui as per design

helenayele:
Okay on the current fix, I can see that selecting User-Selected Layer option brings a dialog to choose the existing layers or to create new one. Creating new layer and adding shapes is working fine. But if I select the existing layers , the combo doesn't show the selected layer, it  shows "User-Selected" layers as selected layer instead . And when  I insert new shape, first it inserts the shape in Misc layer and when I try to insert other shape, it logs NPE.   
#### Stack Traces
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at Debrief.Tools.Palette.CreateLabel$CreateLabelAction.execute(CreateLabel.java:227)
	at MWC.GUI.Tools.PlainTool.execute(PlainTool.java:182)
	at MWC.GUI.Tools.PlainTool.actionPerformed(PlainTool.java:162)
	at org.pushingpixels.flamingo.api.common.AbstractCommandButton.fireActionPerformed(AbstractCommandButton.java:607)
	at org.pushingpixels.flamingo.api.common.AbstractCommandButton$ActionHandler.actionPerformed(AbstractCommandButton.java:559)
	at javax.swing.DefaultButtonModel.fireActionPerformed(Unknown Source)
	at ...
[User_Selected_layer_NPE.txt](https://github.com/debrief/debrief/files/2955629/User_Selected_layer_NPE.txt)

helenayele:
Thanks @ayeshare  , the above issue now gets fixed. But I have got one comment - it would be very nice if we can change the background color of the combo look like the design doc's -  Compare it
[Image from Gyazo](https://gyazo.com/7d9f093226e254cd19a363c9eba15b52)

helenayele:
We need to resize/expand the combo-dropdown on first click. Currently it gets expanded after we click inside it and select a layer - 
[Image from Gyazo](https://gyazo.com/ebe36b54e31f1f193040052b7443861c)

IanMayo:
@helenayele - I've fixed the combo box size issue. Hopefully I haven't broken something else in the porcess.

org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
```diff
@@ -133,28 +136,73 @@ public final Action getData()
...
+            final Layer newLayer = new BaseLayer();
+            newLayer.setName(layerToAddTo);
+
+            // add to layers object
+            _theData.addThisLayer(newLayer);
+          }
+        }
+        if(layerToAddTo!=null) {
+          theLayer = _theData.findLayer(layerToAddTo);
+        }
```
IanMayo: @ayeshare - the logic is confused in this method.  It's not necessarily down to you, but I'd appreciate you going through this refactoring exercise.

We're duplicating the `layer != null` test on the next two lines. Then we create a `res` value that I think gets over-written later on.

I know it's a chore, but could you re-work the logic in the whole of this method, to verify control flows as expected.  

Again, I know it's a chore, but I suggest you declare res as `final Action res;` - this will ensure it both gets assigned _somewhere_, and that it only gets assigned once.

IanMayo:
@ayeshare - I'm worried that the `CreateLabel.getData()` method may hurt us in the longer term. I've asked you to refactor it, with some tips, here: https://github.com/debrief/debrief/pull/3828/files#r265078066

helenayele:
@IanMayo the combobox size is still not changed specially when we have loaded tracks 
- [Image from Gyazo](https://gyazo.com/4fa764ddf3d5207a638d98ab4a9dcb93)

But if we don't have opened track, the size is correct.

IanMayo:
@helenayele - I made quite large changes in my last commit, it's worth re-testing the various insert-shape permutations. I'll investigate sorting out the combo size with a plot loaded.

helenayele:
I didn't find any issue except that "Adding a new drawing feature"dialog's Okay button is always enabled even the user doesn't select any layer.
[Image from Gyazo](https://gyazo.com/1117415eccbd08dce7735b01f6360f40)

IanMayo:
Ok, @helenayele - have just pushed a commit that ensures a layer is selected.

PR #3855, opened 2019-03-15
Use uniform dialog for adding an new layer

### LINKED ISSUE #3846: Modify the title of New Layer dialog
When we invoke Add Layer functionality from outline view, the dialog looks good, it has proper dialog and looks like this:
image

And when we try to Create Layer from the Layers Combo , it does not have proper **title** and looks like 
[Image from Gyazo](https://gyazo.com/2b7f2dcb91df9c00951421a3ff7378f2)

Please make them identical.

IanMayo:
Fixes #3846


IanMayo:
@helenayele - the dialogs should now be consistent.

helenayele:
It is now fixed. but I am not getting add review button on the above

PR #3856, opened 2019-03-15
3854 cant add grid

### LINKED ISSUE #3854: Unable to insert Grid and Scale after Resetting
Once we reset , the default "Chart Features" layer is removed, and after a user loads his rep/dpf files, when he tries to insert those two features , they will not be added, since they are only added to this "Chart Features" layer. Even if a user tries to create the layer with the same name, they are not going to be added. Neither they throw exception.
### Steps to reproduce:
1. Load boat1.rep
2. Select New
3. laod boat1.rep
4. Select [Add new Layer] from  layers combo by using "Chart Features" name
5. Try to add Scale or Grid, see that they are not added

[Image from Gyazo](https://gyazo.com/4b604d8cdec4942275bf8ff1d7e3e164)

IanMayo:
Fixes #3854


IanMayo:
@helenayele - hopefully this one is ok with you, too.

IanMayo:
Ok, @helenayele - this one should be fixed (clearing Chart Features layer on File/New)

PR #3861, opened 2019-03-15
3852 extra marker when time slider was outside range

### LINKED ISSUE #3852: Time player - shows no marker at the end of shorter track
Mentioned by @helenayele in slack.

saulhidalgoaular:
Fix #3852

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
@@ -81,6 +81,7 @@ import MWC.GUI.DragDrop.FileDropSupport;
 import MWC.GUI.DragDrop.FileDropSupport.FileDropListener;
 import MWC.GUI.LayerManager.Swing.SwingLayerManager;
 import MWC.GUI.Undo.UndoBuffer;
+import MWC.GenericData.HiResDate;
 import MWC.GenericData.TimePeriod;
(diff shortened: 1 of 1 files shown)


IanMayo:
works for me :-)

saulhidalgoaular:
> I see 3 issues ,
> 
>     1. When I play back manually , I see duplicate markers - [Image from Gyazo](https://gyazo.com/aa2d2a4a41494283fcafa4bbf5dc15cc)

it looks like frame-dropping, due cpu. I was unable to reproduce in my main computer. I have another computer, which I am going to test in a couple of minutes. Maybe it should be listed as a different issue, since it is not directly related with this fix.

> 
>     2. No starting mark is shown on shorter track - [Image from Gyazo](https://gyazo.com/738e6c13ca8d34205055b14b86800b04)

I consider this behavior is correct, because boat2.rep starts a bit later than boat1.rep. I re-tried it on full-debrief and we have the same behavior.

> 
>     3. When Shorter track finishes playing ,  end marker is seen but then hide again.
>        [Image from Gyazo](https://gyazo.com/0df813103f92ed6faaa136553d2a3c80)

Similar to number 2.

saulhidalgoaular:
I cannot reproduce commend number 1. What comes to my mind is to create another issue about it.

IanMayo:
Ok, merging this. @helenayele - if you still have issue 1, could you create it as a new issue, against D-Lite release 3?

helenayele:
Sure, I am still reproducing issue #1 and filed it at #3877

PR #3864, opened 2019-03-17
3822 Time format is now marked.

### LINKED ISSUE #3822: Mark the selected time format in Format Time control combo
Currently when a user selects a date format it is applied in the time filter and looks like this 
image
It would be very helpful if we mark the selected time format just like full Debrief do
[Image from Gyazo](https://gyazo.com/69f4079f949262e7583f51ff745b89b7)

saulhidalgoaular:
Fix #3822


IanMayo:
Ok @helenayele - I've just pushed a version that shows the current selection.

PR #3899, opened 2019-03-23
[WIP] 2937 import planning legs

### LINKED ISSUE #2937: M30 - Create Planning Legs from REP files
Planning legs can only currently be generated from editing XML files, or via the user interface.

A REP file format should be produced, to allow them to be created from REP files.

We'll need several formats, to allow for the different ways an analyst can specify the distance for a leg.

IanMayo:
Fixes #2937


IanMayo:
Not ready for merge:
* [ ] we need to investigate why the start point of each leg has its name showing.
* [ ] we should verify the correct number of legs in tests
* [ ] we should also add a closing leg example to the test dataset

IanMayo:
@helenayele - this is a new feature. In the `sample_data` folder for this branch, there's a file called `planning_tracks.rep`.   You can open this in a text editor, and view the two sets of planning tracks it contains.

When loaded into Full Debrief, you should see that the two tracks have loaded.

PR #3935, opened 2019-03-29
added globle error log verification support

theanuradha:


IanMayo:
Hello @theanuradha - I'm looking at the code. Is there a line that indicates "only break on errors" (rather than warnings or info?)

Issue #3811, opened 2019-03-04
Some of the functionalities in the Step Control properties window are not working


helenayele:
Some of the step control's properties like
[![Image from Gyazo](https://i.gyazo.com/027e0edb9f697e20b33caf3176a80948.png)](https://gyazo.com/027e0edb9f697e20b33caf3176a80948)

1. Edit Highlighter - throws InvocationTargetException
2. Edit DisplayMode - throws InvocationTargetException
3.  DateFormat - no exception is thrown
4. . Font Size - throws InvocationTargetException
5. Highligther - Throws InvocationTargetException
Full stacktrace
[java.lang.reflect.InvocationTargetException.txt](https://github.com/debrief/debrief/files/2926187/java.lang.reflect.InvocationTargetException.txt)

IanMayo:
@saulhidalgoaular - once you've finished your current work, could you take this on, please?

I guess that for the `edit` buttons, we'll need to popup another Properties Window.

helenayele:
Verified.

PR #3940, opened 2019-03-30
3811 step control properties window

saulhidalgoaular:
Supports #3811 
Not finished yet

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Properties/DateFormatPropertyEditor.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Properties/DateFormatPropertyEditor.java
@@ -161,7 +161,7 @@ public class DateFormatPropertyEditor extends PropertyEditorSupport
    * the currently selected format
    * 
    */
-  protected int _myFormat = 4;
+  protected int _myFormat = 6;
 
   /**
    * provide easy access to the 'first cut of the day' format
@@ -174,6 +174,11 @@ public class DateFormatPropertyEditor extends PropertyEditorSupport
    * 
    */
   public static final String TIME_FORMAT = "HHmm";
+  
+  /**
+   * default format for time label
+   */
+  public static final String DEFAULT_DATETIME_FORMAT = "yy/MM/dd hh:mm:ss";
 
   /**
    * retrieve/initialise our list of date formats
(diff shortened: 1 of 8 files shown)


saulhidalgoaular:
> Good morning @IanMayo . Regarding this commit, I am prompting the properties editor for Display Mode, despite it doesn't have properties to edit. Do we keep this behavior? I have checked Debrief Legacy and it does the same.

oh. Nevermind. I just realized there are properties for the snail painter.

saulhidalgoaular:
There is a weird behavior with options 'T+ 'SSS and 'T+ 'MM:SS' . It just doesn't assign the selected format. I checked and the same happens with Debrief Legacy. Is it expected?
image

IanMayo:
Ah yes, that is a strange format.

It's for when the displayed time is meant be relative to some absolute time.   Imagine a simulated torpedo hitting a vessel.  The time of impact could be "T 00:00".  One minute before that would be "T -01:00" and one minute after would be "T +01:00". 

We no longer use the time label for this purpose though.  So, let's remove those formats from the listing.

Note: we now use the `Time Display (Relative)` UI element to do this time-zero style display of information:
image

saulhidalgoaular:
Good mornigng @helenayele 

**Number 1:** Changes done.
**Number 2:** Removed unneeded formats.
**Number 3:** I think is depends mostly on the width of the application. For example, here I have a it with Font at 30.

Debrief Lite (22 Mar 17 20:24) - boat1 rep -_032

And here I have it with font size 5.

Debrief Lite (22 Mar 17 20:24) - boat1 rep -_033

I remember I saw an issue that defined the specific behavior when resizing for Debrief Lite.
I have changed the range as requested tho. Now the font size can be at most 20.

**Number 4:** I have seen this same behavior in Debrief Legacy. It shows the marker for other files tho (like sample.dpf). I can confirm if it is the desired behavior to check it deeper.

All this changes affect Debrief Legacy.

saulhidalgoaular:
Hi @helenayele 
**Number 1** Removed Range option.
**Number 2** I cannot reproduce that. Do you have the steps to arrive to that state?

saulhidalgoaular:
> Now small step is applied in the player's stepper.
> But we need to use uniform date format in our system. The edit properties use the below selected date format by default
> [Image from Gyazo](https://gyazo.com/92e9721cbb79e3d6ccdf400b7ac68bd6).

Hi @helenayele . I have added the default format to the properties editor and it is now selected by default.

PR #3957, opened 2019-04-03
eliminate the need for ctrl+click

### LINKED ISSUE #3953: Full-Debrief - can't paste into top level
A user had some trouble this morning.

It's an issue that goes back a long way.

With data loaded, the analyst wanted to make a copy of a track.  He right-clicked on the track, and then right-clicked into empty space at the bottom of the Outline view.

The popup menu that was shown, was for the original track - not the top level.

From this popup menu, he found `Paste` and pasted the track - expecting it to be a new track at the top level. Instead it was pasted into itself, with unpredictable consequences.

The only way to paste into the top level is to use ctrl-click to de-select the original track, then right-click in the empty space.

I think we (you) fixed a similar issue in D-Lite.

When the user right-clicks in the Outline, and the click target is below the present items, the right-click menu should be for the top level, not the currently selected item.

Here is the menu that was offered after copying a track, then right-clicking into empty space:
image

This is what is provided after using ctrl-click to de-select the existing track:
image

ayeshare:
Fixes #3953


org.mwc.debrief.core/src/org/mwc/debrief/core/editors/PlotOutlinePage.java
```diff
@@ -797,7 +797,7 @@ private void contributeToActionBars()
   public void createControl(final Composite parent)
   {
 
-    _treeViewer = new MyTreeViewer(parent, SWT.MULTI | SWT.H_SCROLL
+    _treeViewer = new MyTreeViewer(parent, SWT.SINGLE | SWT.H_SCROLL
```
IanMayo: @ayeshare - I think we need to be able to do multi-select, partly to request an XY graph of multiple tracks.  Is it essential to have SWT.SINGLE for this fix to work?

IanMayo: Aah, I've switched it back to MULTI, and it works ok.  

If there is a problem, we'll have to come back & fix it.

Issue #3886, opened 2019-03-21
Layers combo should refresh its size immedietly


helenayele:
After resetting, layers combo does left extra spaces. It should refresh its size as soon as it gets reset.

### Steps
1. Load Shapes.rep
2. Open Layers combo from insert menu and see its size
3. On File menu , select New > Don't save it
4. Now select Insert menu and open layers combo - it has some extra spaces left seen.
[![Image from Gyazo](https://i.gyazo.com/9009fff083b5527ad5425c06e6e65a49.png)](https://gyazo.com/9009fff083b5527ad5425c06e6e65a49)
5 . Click inside the combo again and see that now it gets refreshed and no more extra spaces.
Video
[space_in_layers_combo.zip](https://github.com/debrief/debrief/files/2991147/space_in_layers_combo.zip)

IanMayo:
Hello @theanuradha - you haven't had much to do with Debrief-Lite.  But, I'd like your support in tidying up some UI issues, please.

For this one, you'll branch off `develop`. 

Hmm, I think there's at least one other issue associated with this drop-down.  I'll try to assign them to you, too.

IanMayo:
@helenayele - this should be fixed now.

helenayele:
yes, now it is resized according to the size of the layers. Verified.

PR #3959, opened 2019-04-04
fix combo repain issues on dlite

theanuradha:


theanuradha:
#3886

IanMayo:
@theanuradha - we've got some static analysis warnings for the new code (see Codacy review above).

I just tried to fix them, but couldn't checkout the branch. When I did, I couldn't push back to your repo.

Would you mind going in to fix the compiler warnings for the above code please?

theanuradha:
@IanMayo should be fine now

IanMayo:
@theanuradha - I think this PR introduced a bug for the combo:

> there is also issue in layers combo selection - it always default to User Selected layer 
> option - even if I select like the other options , when I try to insert shapes  - 
> it comes back the user selected layer option - https://gyazo.com/4e06ab68d7dbee6a6c4390d0314465e2
> See in the video, after I select Chart features when I try to insert shape, the combo 
> selects the user Selected layer by itself

PR #3962, opened 2019-04-04
build changes to include shp file in build

### LINKED ISSUE #3956: Deploy Debrief-Lite with coastline shapefile already present
Having a coastline available in Debrief-Lite makes a great improvement to the initial view, and is always able to give some context for where in the world tracks are.

We have a folder containing the shapefile (and supporting files) of country borders that would be good for this `ne_10m_admin0_countries_89S`

Here's the master copy:
https://github.com/debrief/NaturalEarth/tree/master/ne_10m_admin0_countries_89S

We _could_ put this folder into the top level of the distribution and expect users to navigate to it.  

But, my preference would be to have it as a resource inside a Debrief-Lite .jar file.

The Debrief-Lite build process would pull in the folder, and package it with the app.

ayeshare:
Fixes #3956

### DIFF
--- a/org.mwc.debrief.lite/build.xml
+++ b/org.mwc.debrief.lite/build.xml
@@ -104,6 +104,7 @@
 			<fileset dir="libs" includes="*.jar"/>
 		</copy>
 		<zip destfile="${dist.dir}/debrief-lite-1.0.377.zip">
+			<fileset dir="." includes="data/**/*"/>
 			<fileset dir="${debrief-lite-dist.dir}" includes="**/*"/>
 		</zip>
 	</target>
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/map/GeoToolMapRenderer.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/map/GeoToolMapRenderer.java
@@ -67,7 +67,7 @@ public class GeoToolMapRenderer implements BaseMap
   {
     _myRenderers.add(renderer);
   }
-  
+
   private static class CustomMapPane extends JMapPane
   {
 
@@ -75,7 +75,7 @@ public class GeoToolMapRenderer implements BaseMap
      *
      */
     private static final long serialVersionUID = 1L;
-    
+
     private final MouseDragLine dragLine;
 
     private final GeoToolMapRenderer _renderer;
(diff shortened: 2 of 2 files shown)


IanMayo:
@ayeshare - at run-time, on a sea-riders laptop, they won't have the `org.cmap.NaturalEarth` plugin present. We need a copy in this plugin. (or I may be mistaken)

ayeshare:
> riders laptop, they won't have the `org.cmap.NaturalEarth` plugin present. We need a copy in this plu

They wont need natural earth.  We are referring to natural earth plugin in dev mode.  I can try removing reference to that and directly use it from data folder itself

IanMayo:
Aah, sorry @ayeshare - I was mistaken.  It's quite acceptable to do that for in dev mode.

You are welcome to re-instate it, I forgot we had that dev workaround.

I think we need to highlight/comment on what's happening more clearly, particularly for next time I get confused on it :-)

We'll also need to run test-builds, to verify that D-Lite is able to get the data out of the `.jar`.

ayeshare:
> Aah, sorry @ayeshare - I was mistaken. It's quite acceptable to do that for in dev mode.
> 
> You are welcome to re-instate it, I forgot we had that dev workaround.
>

We dont have to re-instate it, the same code will work for dev because the data folder is in current directory.
 
> I think we need to highlight/comment on what's happening more clearly, particularly for next time I get confused on it :-)

The build copies the data folder under org.mwc.debrief.lite under dist and the released zip will include the data folder.
The zip file will have debrief-lite.jar, data(folder containing shp file) and libs(folder containing all jars).
The shp file is loaded from the DebriefLiteApp.java when app is launched.  In case the file is not found, for any reason, user can point to the directory.
For dev mode:
DebriefLiteApp.java will load the shp file from the data folder under current directory.
 
> We'll also need to run test-builds, to verify that D-Lite is able to get the data out of the `.jar`.

I can create a jar file and send it to @helenayele to test it.  Is there any other way she can get it before merging to develop?

IanMayo:
> I can create a jar file and send it to @helenayele to test it. Is there any other way she can get it before merging to develop?

No, your manually built way is the only way I can think of.

IanMayo:
Hmm @ayeshare , I'm running `ant clean dist` and I get a compile error (attached).

No compiler error in Eclipse.

Any suggestions?

[res.txt](https://github.com/debrief/debrief/files/3053641/res.txt)

helenayele:
@IanMayo  would you tell me how am going to review this please? in the morning Ayesha uploaded the jar/zip file and I try to run from the command line and able to test it.

IanMayo:
@helenayele - and you saw the coastline file?  That's enough to verify it.

If you zoom in the UK it should be a crisp coastline. Let me find an example.

PR #3964, opened 2019-04-04
3960 Properties editor not opening for Tracks

### LINKED ISSUE #3960: Properties editor not opening for Tracks
Ooh, this is a scary one.

I open Debrief. I drag in `boat1.rep` and `boat2.rep`.  

If I double-click on one of the tracks `Track:NELSON`, the property editor only shows with one item:
image

I suspect we either have to step through the `getInfo()` method in TrackWrapper, or use `git bisect` (hopefully from a graphical git client) to find where the problem was introduced.

saulhidalgoaular:
Fix #3960 
This reverts commit aa529ca05d7b57803b43219ed3f2082975a28ace.


IanMayo:
Oh wow, I got that logic quite wrong :-(

PR #3974, opened 2019-04-06
3966 Update name for tracks

### LINKED ISSUE #3966: Renamed layer named is not applied in Edit Properties tab
See the below Screenshot:
[Image from Gyazo](https://gyazo.com/a0861aac4464e4b694771d92065dd9b1)

See that I renamed the layer RED_FORCES to "RED_FORCES_Renamed" but in the Edit properties , the tab "RED_FORCES" is still retained the old layer name. 
RE-opening the edit properties window doesn't help also.

saulhidalgoaular:
Fix #3966

### DIFF
--- a/org.mwc.asset.legacy/src/ASSET/GUI/Editors/Decisions/CompositeEditor.java
+++ b/org.mwc.asset.legacy/src/ASSET/GUI/Editors/Decisions/CompositeEditor.java
@@ -10,7 +10,7 @@
  *
  *    This library is distributed in the hope that it will be useful,
  *    but WITHOUT ANY WARRANTY; without even the implied warranty of
- *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. 
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  */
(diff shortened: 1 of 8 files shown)


IanMayo:
@saulhidalgoaular - are you using the Eclipse code cleanup settings from the `contribs` folder?

https://github.com/debrief/debrief/blob/develop/contribs/code_cleanup_settings.xml

saulhidalgoaular:
> @saulhidalgoaular - are you using the Eclipse code cleanup settings from the `contribs` folder?
> 
> https://github.com/debrief/debrief/blob/develop/contribs/code_cleanup_settings.xml

I re-formatted it with the file given just in case. I was using the format settings that we used in the ppt_renderer project (I think it was the same).

Well, I re-pushed it.

IanMayo:
@saulhidalgoaular - the suggestions from `Codacy` all seem valid in the above review.  Could you implement them?

(I'm assuming you can see them. If you can't, I'm happy to do them).

IanMayo:
Yes, @saulhidalgoaular  - let's ignore the `really long class` warning for SwingPropertyEditor. It's legacy code that we may break by refactoring.

PR #3976, opened 2019-04-07
2996 Adding layers to png

### LINKED ISSUE #2996: M1.6.3.1 Export .png
As an **analyst** I wish to **export a copy of the plot as a `.png` file** so that I can help my audience understand what happened in the exercise.

## Status
We now have a bare-bones app that looks _something_ like an analysis application.  To test the full _cradle to grave_ analysis flow, we need to output some data.  So we need to add a button to the interface (or ctrl-c key listener) that will place a bitmapped copy of the current plot on the clipboard.

## General strategy
The Full Debrief method to do the above is in `CorePlotEditor` action called `copyClipboardAction`. Elements of that can be re-used.  Ideally, the non-Swing elements of the above would be moved to a class in `org.mwc.debrief.legacy` in order that the same code be available to Debrief-Lite and full Debrief.

## Testing
* open Debrief-Lite
* load shapefile coastline
* load track data
* zoom in on some portion of the track
* copy plot to clipboard
* paste to MS Word - verify image matches that in Debrief.

## Implementation

* Create image same dimensions as map
* Tell map to render to image
* Integrate ctrl-C shortcut*
* Add menu option to copy to clipboard
* Put PNG file onto clipboard
*  Acceptance: perform copy operation, Paste into MS Word. Verify it matches the displayed map

saulhidalgoaular:
Fix #2996

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/menu/DebriefRibbonFile.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/menu/DebriefRibbonFile.java
@@ -11,7 +11,6 @@ import java.awt.datatransfer.Transferable;
 import java.awt.datatransfer.UnsupportedFlavorException;
 import java.awt.event.ActionEvent;
 import java.awt.image.BufferedImage;
-import java.awt.image.RenderedImage;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


IanMayo:
Great, well done @saulhidalgoaular

PR #3978, opened 2019-04-08
remembered the last save and last open location

### LINKED ISSUE #3827: Make Open Dialog smart enough to remember previous selected location
Our  Open dialog needs to be  a little smart just like window open dialog. When I open file say from D://Debrief location for the first time, being in the same session of D-Lite , when I try to open file, can it remember my previous selected location (D://Debrief) so that it will save my time.

Currently it uses the users home directory as root location, and it doesn't create good experience specially for windows user.

ayeshare:
Fixes #3827


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
```diff
@@ -641,7 +641,16 @@ public void exit()
         }
         else
         {
-          outputFileName = DoSaveAs.showSaveDialog(null, "DebriefPlot");
+          final File directory;
```
IanMayo: Good use of `final` keyword :-)

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
```diff
@@ -973,4 +982,19 @@ protected void update(final Layers theData, final Plottable newItem,
   {
     _instance.getLayerManager().updateData((Layer) theLayer, newItem);
   }
+  public static String getLastFileLocation()
+  {
+    return _instance.app.getProperty("last_file_location");
```
IanMayo: @ayeshare - this is a valid solution, but I don't like these "magic strings".  I'll replace them with constants.  But, I'll also do some refactoring, to try to avoid these static methods.   

The app has the concept of preferences already.  I'll move knowledge of the String back to the `SaveAs` (or relevant class), and then just use the existing preferences.

IanMayo:
@helenayele - I've changed how Debrief remembers the file locations.  Would you mind checking it still works, please?

PR #3994, opened 2019-04-09
Move DynamicShapes out of core, support them in D-Lite

### LINKED ISSUE #3683: Dynamic Shape Handler needs to move plugin
`DynamicShapeLayerHandler` and `DynamicShapeHandler` are currently in org.mwc.debrief.core

That means they can't be referred to from `Lite`.

More them to Debrief.legacy.

In full Debrief they're loaded by extension in the `plugin.xml` - we should explicitly register them in `DebriefLiteApp`

IanMayo:
Fixes #3683


IanMayo:
This is a code change, not a UI change.  The UI change will be applied to the `develop` branch in a separate PR.

PR #4000, opened 2019-04-09
Hotfix npe in insert chart feature

IanMayo:
We were getting NPE if the user pressed `Cancel` on the `Select target layer` dialog for `Insert Chart Feature`.


IanMayo:
Wrong commit got pulled into this PR.

PR #4001, opened 2019-04-09
Move dynamic shapes into legacy

IanMayo:
We wish to access dynamic shapes from legacy, so we need to remove them from the `.core` package.


IanMayo:
Close. Re-doing in smaller steps, to track where UI_Test failure is introduced

Issue #3697, opened 2019-02-05
When dragging and dropping non-rep file on the D-Lite, it silently fails


helenayele:
When dragging and dropping non-rep file on the D-Lite, it fails silently, no proper warning dialog is shown . Sometimes it  shows this error dialog 
[![Image from Gyazo](https://i.gyazo.com/9fec1152e326d662d88cb500787f0f47.png)](https://gyazo.com/9fec1152e326d662d88cb500787f0f47) and logs  this 
[DynamicLayer handler not found.txt](https://github.com/debrief/debrief/files/2832784/DynamicLayer.handler.not.found.txt)

### Expected Result
It should show warning Message ,that informs the user that he/she should drag-drop only rep files.

helenayele:
And also D-Lite should support Both xml and dpf file types in addition to rep files.

IanMayo:
@helenayele - could you re-validate this, please?

helenayele:
Okay, now , when I try to load some problematic dpf files like sample.dpf, boat1t.dpf files, it silently logs the below exception and D-Lite becomes busy for long period of time. I can only terminate it from eclipse.
java.lang.IllegalArgumentException: not implemented (yet)
	at org.mwc.debrief.lite.LiteParent.logError(LiteParent.java:13)
	at MWC.Utilities.Errors.Trace.trace(Trace.java:222)
	at MWC.Utilities.Errors.Trace.trace(Trace.java:155)
	at org.mwc.debrief.lite.DebriefLiteApp.FilesReceived(DebriefLiteApp.java:527)
	at MWC.GUI.DragDrop.FileDropSupport.fileReceived(FileDropSupport.java:145)
	at MWC.GUI.DragDrop.FileDropSupport.drop(FileDropSupport.java:179)
	at java.awt.dnd.DropTarget.drop(Unknown Source)
	at sun.awt.dnd.SunDropTargetContextPeer.processDropMessage(Unknown Source)
	at sun.awt.dnd.SunDropTargetContextPeer$EventDispatcher.dispatchDropEvent(Unknown Source)
	at sun.awt.dnd.SunDropTargetContextPeer$EventDispatcher.dispatchEvent(Unknown Source)
	at sun.awt.dnd.SunDropTargetEvent.dispatch(Unknown Source)
	at java.awt.Component.dispatchEventImpl(Unknown Source)
	at java.awt.Container.dispatchEventImpl(Unknown Source)
[...]

PR #4007, opened 2019-04-10
3683 loading dynamic layers in d lite

### LINKED ISSUE #3683: Dynamic Shape Handler needs to move plugin

IanMayo:
Fixes #3683 

This fix allows boat1t.dpf to be loaded.  Once loaded, you will see the time displayed in two places on the plot.

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Canvas/ExtendedCanvasAdapter.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Canvas/ExtendedCanvasAdapter.java
@@ -1,0 +1,112 @@
+package MWC.GUI.Canvas;
+
+import java.awt.AlphaComposite;
+import java.awt.Color;
+import java.awt.Composite;
+import java.awt.Graphics;
+import java.awt.Graphics2D;
+import java.awt.Shape;
+
+import MWC.Algorithms.PlainProjection;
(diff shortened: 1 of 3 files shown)


IanMayo:
@helenayele - while this PR prevents the warnings when loading those files, it doesn't correctly plot the time label in `boat1t.dpf`.

That has been captured in a separate issue - #4009

helenayele:
@IanMayo  does it support to open dynamic shapes like arc, rectangle,etc

IanMayo:
@helenayele - you gave a list of failing files here:
https://github.com/debrief/debrief/issues/3697#issuecomment-480796373

IanMayo:
@helenayele - yes, it can load `boat_1_2_arcs.rep`.

The dynamic shapes load (we can verify it from the Outline). But, the rendering still needs to be fixed.

helenayele:
Yes from outline view the dynamic shapes are fully loaded but they are not loaded in the map. sure will wait the fix

IanMayo:
@helenayele - I'm deferring that fix. It has external dependencies.

PR #4022, opened 2019-04-12
Fixed exception handling to propagate back to the caller.

### LINKED ISSUE #3826: Provide user friendly Info when user tries to open invalid files
When I try to open Rep files using **Open Plot** action , it displays these dialogs in ascending order .

1. [Image from Gyazo](https://gyazo.com/7c4e31bb65319acc17746379580f6fd4)

2. [Image from Gyazo](https://gyazo.com/117462fc0890310827aa902cf31d1fb2)

For some of the Rep files the first dialog may not be shown but the 2 one is common for all. 
### Expected Result:

1. No multiple error dialog should be shown for the user -  A Single error dialog should be seen
2. The file name should not be loaded in the title bar
3. Should not silently logged exception it should show the problems in dialog 
4. It would be very cool if we  provide user friendly message to our users.

ayeshare:
Fixes #3826


IanMayo:
@helenayele - I'm going to create a new PR that re-structures this one (for weird admin reasons).

But, could you please review @ayeshare's solution?

IanMayo:
@helenayele - so has the problem gone away in the `develop` branch?

helenayele:
yes, I don't get such exception while opening rep/dpf files - all problematic dpf files are handled

IanMayo:
Correctly fixing the UI issue involved changes in lots of files, together with changes to a fundamental part of Debrief (xml import) - which has a chance of affecting application reliability.

Since the UI issue has gone away, we no longer have to accept that potential problem.

So, sadly I think the correct option is to close this PR.

IanMayo:
I'm closing this PR, since I found an alternate strategy that affected fewer files. 

Lots of calls to Import come from tests.  These were all modified to `catch` the `ParseException`.  But, by making `ParseException` extend `RuntimeException` (which is unchecked), no changes were necessary in the unit tests.

We just had to catch it in our UI code.

PR #4024, opened 2019-04-12
3826 open plot error 3

IanMayo:
Attempt minimal set of steps necessary to correctly handle D-Lite import failure.


IanMayo:
@helenayele - does this branch contain the same dialog-related problems as #4025?

helenayele:
No it doesn't contain such dialog related issues.

IanMayo:
@helenayele - can you remember which bad data file silently opened?

helenayele:
sure this is the dpf file that opened silently - 
[boat2_eee.zip](https://github.com/debrief/debrief/files/3079967/boat2_eee.zip)

IanMayo:
Aaah, @helenayele - this is the `master` branch, which is deliberately behind `develop` when it comes to DebriefLite.

You only need to test Full Debrief from this branch.

helenayele:
So @IanMayo  what are the changes  in the full debrief due to this PR ?

IanMayo:
Morning @helenayele  :-)  The changes will affect loading `.rep` and `.dpf` files - both by drag and drop or using the `File` menu.

helenayele:
But @IanMayo  can we stop loading the file on full debrief if it is invalid file. Currently it loads the file even if it is invalid - [Image from Gyazo](https://gyazo.com/e1a53fb5306d09d7c16835dc8f1fb684) ?

IanMayo:
@helenayele - I think it loaded the first 6 points - since the problem was only on line 7.  I _think_ It's the same for a similarly broken file in D-Lite.

helenayele:
Yes , D-Lite also tries to load - it behaves the same as of  Full Debrief.

Issue #4020, opened 2019-04-11
Copy/Paste cloning in tests failing on Travis


IanMayo:
I wanted to use Copy/Paste to test something in Travis.

The test failed, because Eclipse couldn't find the class definitions:
```
java.lang.ClassNotFoundException: Debrief.Wrappers.Track.RelativeTMASegment
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:501)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:421)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:412)
	at org.eclipse.osgi.internal.baseadaptor.DefaultClassLoader.loadClass(DefaultClassLoader.java:107)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at java.io.ObjectInputStream.resolveClass(ObjectInputStream.java:683)
```
Here is the raw output:
https://travis-ci.org/debrief/debrief/builds/518810509#L4277

PR #4028, opened 2019-04-15
introduce copy/paste test

IanMayo:
@theanuradha has identified a fix to but #4020 

This PR attempts to apply it to the base `master` branch.


IanMayo:
No, we won't apply this set of changes - we need G's classpath changes, too.

PR #4031, opened 2019-04-16
handled import of dsf files

### LINKED ISSUE #3905: D-Lite not loading sensor data (.dsf)
When we drag in `.dsf` file, the file gets processed, but the sensors don't appear in the Outline.

This is because of the Full Debrief processing where loading the sensors is deferred, to allow:
* if only one set of sensor data, let the analyst format it
* if no host platform is found, ask user to select one.

We don't need the above capabilities for D-Lite.

So, we should probably just perform `ImportReplay.StorePendingSensors()`

ayeshare:
Fixes #3905


IanMayo:
@helenayele - is this ready for merge?

PR #4034, opened 2019-04-16
set selection to null after removing element

### LINKED ISSUE #4014: Cut action doesn't remove an item after pasting
Currently after I cut a shape from one layer and paste it into another layer, the shaped doesn't get removed from the original layer.
[Image from Gyazo](https://gyazo.com/42530fabb654859dd0c63c545df022e5)

This happens for both shapes and positions.

ayeshare:
Fixes #4014


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/outline/OutlinePanelView.java
```diff
@@ -323,9 +323,9 @@ public void valueChanged(TreeSelectionEvent e)
       @Override
       public void actionPerformed(ActionEvent e)
       {
-        if(_cutContents!=null) {
```
IanMayo: @ayeshare - it's not good practice to leave the line commented out like this - the future Ian (or even the current Ian) won't understand what's happening.

If we don't need the code we should delete it.  Or, add a comment to explain why it's commented out, please.

IanMayo:
@helenayele  - yes, I get the same behaviour. Very strange 😆 

@ayeshare - I did just have a look to see if there was a quick fix. But, sadly not.  I'll have to leave it to you.

IanMayo:
I'm merging this. The error message text has been corrected in master, which is now in develop.  So, it's effectively already fixed.

PR #4042, opened 2019-04-17
add chartfeatures layer if deleted

### LINKED ISSUE #4004: User not able to add Grid and Scale again after deleting them
Currently a user doesn't have a chance to re-insert Grid and Scale  to his/her map once they deleted them.

### Steps:
1. Open boat1.rep
2. Add scale and grid shapes, they are added in the chart features by default
3. Delete chart features layer
4. try to add now Scale and Grid - they will not be added - no exception/error or message is seen

### Expected Result:
User may want to re-add those shapes so our app should be able to allow this.

ayeshare:
Fixes #4004


ayeshare:
> Fixes #4004

@IanMayo there is probably still a case where an old replay file is imported and it doesnt have chart features layer.  Let me know if this case exists.
Current fix is to disable delete action if user selects chart features layer

IanMayo:
@helenayele - I made quite a few changes to this, I'd appreciate you re-testing it, please.

PR #4048, opened 2019-04-18
Pull plain create changes back into develop

IanMayo:
#4042 needs some changes deep in old code - so I'm making the change first in master, then pulling it to develop


IanMayo:
Pushing changes through. Will fix coding issues once in `develop`

PR #4058, opened 2019-04-19
D3 new slicing algorithm

IanMayo:
Integrate updated slicing algorithm(s) into SATC


IanMayo:
Introduce new slicing algorithm.  Refactored some SATC code to allow easier integration, not that it isn't expected to run standalone.

PR #4070, opened 2019-04-24
by default disable all buttons that need a selection.

### LINKED ISSUE #4059: Enable Add Layer button after loading files in D-Lite
Currently it is not possible to add layer after loading files in D-Lite.
See the following screenshots:
[Image from Gyazo](https://gyazo.com/e7d4a882c2a4ea4525f49a96230ed3ab)
when we have only a single layer, the Add Layer button is not enabled.

[Image from Gyazo](https://gyazo.com/f2ec484691c24bbd1921563658ed218a)

ayeshare:
Fixes #4059


IanMayo:
@ayeshare  - it looks like @helenayele has identified an outstanding issue with this PR (edit butting enabled with nothing selected)

PR #4089, opened 2019-04-26
4053 reset xyplot

### LINKED ISSUE #4053: Need to empty XY Plot on File > New
Saul, when we do File > New and we reset the plot, we should clear the XY Plot, and it's menus.

saulhidalgoaular:
Fixes #4053


IanMayo:
@helenayele - could you check this too, please?

Issue #3865, opened 2019-03-18
M1.6.4.10 Support for Undo in D-Lite


IanMayo:
We need to introduce support for Undo in Debrief-Lite
----
#### Update 
Ayesha & I just chatted.
1. ensure undo/redo works for:
    a. insert decorations (scale/grid)
    b. insert shapes & reference data
    c. pan/zoom/zoom-out/fit-to-window
    d. property editing
    e. support for cut/paste/delete/add-layer in Outline View
2. add UI elements for undo/redo
3. connect UI to UndoBuffer

The logical sequence for these is to break them down into these PRs:
a.  2, 3, 1a . **Done**
b. 1b . **Done**
c. 1c . _defer until after GeoTools upgrade_
d. 1d . **Done**
e. 1e

The split is so we can review/discuss a working implementation before going on to support the other actions.  The other steps are broken down into separate PRs, since they represent quite different implementations of Undo.

----

It will need a place on the toolbar (or _maybe_ in the application-level drop-down, shown as an _Orb_ in the flamingo tutorial).

We will also need a command history class, into which changes are stored, so we can move backwards & forwards through undo/redo.

We will then have to wrap a number of UI changes in to `Action` objects, that have execute, undo, redo, etc.

Hey, here's an article from TWENTY ONE YEARS AGO on implementing a Swing undo history:
https://www.javaworld.com/article/2076698/add-an-undo-redo-function-to-your-java-apps-with-swing.html

And here's a slightly more modern one:
https://www.forward.com.au/javaProgramming/javaGuiTips/undoRedo.html

IanMayo:
@ayeshare - I'm assigning this meaty one to you.

We've got an `UndoBuffer` object in D-Lite.  I'd like you to work through the ribbons, integrating the undo functionality.  It's probably worth starting with `Insert` and `View` menus.  Hopefully you can copy the functionality from either Debrief Classic or Full Debrief.  You will also have to add `undo` and `redo` buttons to the title bar of the app.

Happy to support you as required.

helenayele:
Functional tests:
1. After user adds a layer , s/he can undo the action so the new layer should be removed
2. After a user deletes a layer. s/he can undo the action so the removed layer should be re-appear
3. After a user removes a layer or shape in a layer, or position from track layer, redo can make the object to appear in the previous location. The same for adding action.
4. If user tries to cut /copy and paste an object, and wants to redo - the pasted object should be removed from current position and be available  in original location.

[...]

PR #4102, opened 2019-05-01
implemented undo/redo on outline view actions

ayeshare:
Supports #3865


IanMayo:
@helenayele - I'd appreciate you testing these outstanding items, please.

helenayele:
and when I undo and redo the visibility of the shapes/layers , it is a bit slow and it doesn't seem it is working  - [Image from Gyazo](https://gyazo.com/fb80aeb28e5563e8e11c859e45bffcfa)

helenayele:
Getting NPE while redoing edit properties action:
### Steps:
1. Add Grid
2. Change its name by opening its property window
3. Undo 
4. Redo
When it reaches retrieving the name of the Grid ,it logs NPE.
[NPE_While_Redo.txt](https://github.com/debrief/debrief/files/3137224/NPE_While_Redo.txt)

IanMayo:
Ok, @ayeshare - I've fixed the undo/redo for clicking on the icons in the Outline View.

But, I've introduced a problem. After I `redo`, the text for the item disappears:
image

(the text for `Misc` layer from `sample.dpf` when I use undo/redo to undo the change visibility icon change)

I'm not familiar with how we render the items in the OutlinePanelView.  Would you mind having a look?

helenayele:
I still see some issues
1 .  after I add grid and scale 
[Image from Gyazo](https://gyazo.com/f15bb4108d7acd09ddf73b70f5079af2)
See the console message in the video - after adding different shapes I click undo and finally I add Grid and Scale, When I add scale and grid Chart features layer is added, and when I click undo - both grid and scale are removed but the Chart features layer is not removed.

2. The exception  mentioned in the above comment still reproducible.

3. When duplicate shapes are added in a layer, when undo, they will get removed at once - is this expected? [Image from Gyazo](https://gyazo.com/21109fbba9681bb7c7054977cfeeb54b)

org.mwc.debrief.legacy/src/Debrief/Tools/Palette/CreateLabel.java
```diff
@@ -224,6 +230,9 @@ public final void undo()
      */
     public final void execute()
     {
+      if(addedLayer!=null) {
```
IanMayo: @ayeshare - _I think_ we need to check if the layer is already present before adding it.  For `redo` I know it will have been deleted already, and we have to add it.  But, for the initial `execute` I'm pretty sure we've already added the new layer to the `Layers` object.

ayeshare: Thanks @IanMayo

IanMayo:
@helenayele - yes, nicely judged.

PR #4109, opened 2019-05-02
4062 Time Filtering for XY Graph

### LINKED ISSUE #4062: Implementing Time Filtering for XY Plot
Debrief Lite need to have Time Filtering for XY Plot using the Time Controller tab in the Ribbon.

saulhidalgoaular:
Fix #4062

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
@@ -509,6 +509,14 @@ public class GraphPanelToolbar extends JPanel
     }
   }
 
+  /**
+   * It extracts the TrackWrapper objects, then it tries to assign it. If it contains the same
+   * values, it is not assigned and returns false. True when it is assigned.
+   * 
+   * @param layers
+   *          Layers of the session. We are extracting Tracks from it.
+   * @return true if it was actually assigned. If they are the same, they are not assigned.
+   */
   private boolean assignTracks(final Layers layers)
   {
     final Enumeration<Editable> elem = layers.elements();
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
@@ -222,27 +222,33 @@ public class GraphPanelToolbar extends JPanel
     });
 
     if (_stepControl != null && _stepControl.getLayers() != null)
-    { 
+    {
       final DataListener trackChangeListener = new DataListener()
       {
 
         @Override
         public void dataExtended(final Layers theData)
         {
-          assignTracks(theData);
+          if (!assignTracks(theData))
+          {
+            updateXYPlot(operationComboBox);
+          }
         }
 
         @Override
         public void dataModified(final Layers theData, final Layer changedLayer)
         {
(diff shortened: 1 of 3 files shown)


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
```diff
@@ -503,10 +509,9 @@ private void updateSymbolButton(final String symbolOn, final String symbolOff,
     }
   }
 
-  private void assignTracks(final Layers layers)
+  private boolean assignTracks(final Layers layers)
```
IanMayo: @saulhidalgoaular - could you include a javadoc to explain what the boolean status means?

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JSelectTrackModel.java
```diff
@@ -140,14 +141,29 @@ public void setPrimaryTrack(final TrackWrapper newPrimary)
   }
 
   @Override
-  public void setTracks(final List<TrackWrapper> tracks)
+  public boolean setTracks(final List<TrackWrapper> tracks)
```
IanMayo: Aaah @saulhidalgoaular - could you indicate the meaning of the returned value here, too.  Well, if it's not documented in the parent class.

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JSelectTrackModel.java
```diff
@@ -140,14 +141,29 @@ public void setPrimaryTrack(final TrackWrapper newPrimary)
...
+    }
     for (final TrackWrapper track : tracks)
     {
-      this._tracks.add(new TrackWrapperSelect(track, false));
+      newTracks.add(new TrackWrapperSelect(track, false));
+      isDifferent |= !oldTracksSet.contains(track);
+    }
+    if (isDifferent)
+    {
+      this._tracks = newTracks;
```
IanMayo: Hmm, with a slight change, we can make _tracks a final field.

So, instead of:
````
this._tracks = newTracks;
````
we would use:
````
 _tracks.clear();
 _tracks.addAll(newTracks);
````

There isn't a strong argument for the change, it's just the slight benefit of the compiler ensuring _tracks never gets deleted/replaced.

@saulhidalgoaular - what do you think?

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/graph/GraphPanelToolbar.java
saulhidalgoaular: Good morning. Sure.

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JSelectTrackModel.java
saulhidalgoaular: I agree. I am changing it too.

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JSelectTrackModel.java
saulhidalgoaular: Good morning. Yes, javadoc has been added to the interface. I am adding it here too to make the code more readable.

PR #4126, opened 2019-05-03
4103 jre in msi

### LINKED ISSUE #4103: D32.1 Package JRE in MSI
We have scripts to produce 32bit and 64bit MRI installer files.

One user community wish to switch to distributing a JRE with Debrief, to avoid trouble with apps that require different JRE versions.

In support of that, I wish us to make these changes:
* stop producing the 32 bit installer (by commenting out the line(s) in `.travis.yml`)
* change the script so that a JRE 8 zip is downloaded from AdoptOpenJDK.net . (I just timed it, it took about 20 seconds) [1]
* change the MSI install script so that it pulls the JRE into the MSI install
* run the MSI script, and verify the created `.MSI` file.

[1] Here's the URL for the current JRE8.  We'll soon be moving to JRE11, so for now it's ok to hard-code this URL: https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u212-b03/OpenJDK8U-jre_x64_windows_hotspot_8u212b03.zip

saulhidalgoaular:
Fix #4103

### DIFF
--- a/contribs/msi/make_x64msi.sh
+++ b/contribs/msi/make_x64msi.sh
@@ -24,7 +24,7 @@ echo "Done."
 
 echo "Downloading JRE"
 wget -O jre.zip ${JRE_URL}
-unzip jre.zip -d ${SOURCEDIR}
+unzip -q jre.zip -d ${SOURCEDIR}
 mv ${SOURCEDIR}jdk* ${SOURCEDIR}jre
 rm jre.zip
 echo "Done."
--- a/contribs/msi/make_x64msi.sh
+++ b/contribs/msi/make_x64msi.sh
@@ -9,6 +9,8 @@ SOURCEDIR=contribs/msi/src/
 RESOURCESDIR=contribs/msi/resources/
 #Path to the wixl working folder
 WORKDIR=contribs/msi/
+#URL of the jre to use
+JRE_URL=https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u212-b03/OpenJDK8U-jre_x64_windows_hotspot_8u212b03.zip
 
 echo "Clearing the sources folder."
 rm -rf ${SOURCEDIR}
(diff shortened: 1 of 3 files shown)


saulhidalgoaular:
Good afternoon. It looks OK, but I will test the script again later

contribs/msi/make_x64msi.sh
```diff
@@ -20,6 +22,13 @@ version=$(grep "product.*version"  org.mwc.debrief.product/debriefng.product  |
 sed -i "s/versionReplacement/$version/g" ${WORKDIR}Debrief64.wxs
 echo "Done."
 
+echo "Downloading JRE"
+wget -O jre.zip ${JRE_URL}
+unzip jre.zip -d ${SOURCEDIR}
+mv ${SOURCEDIR}jdk* ${SOURCEDIR}jre
```
IanMayo: Hmm, @saulhidalgoaular  - this looks surprising.  Why is the folder called `jdk`?

IanMayo:
@saulhidalgoaular - have performed test release, to verify that Travis is happy with the new steps:
https://github.com/debrief/debrief/releases/tag/20190504_trial_JRE_in_MRI_2

Could you download & verify the MSI please?

Aah, it's probably worth skimming through the Travis build log, to verify it's happy with the steps (though I expect it would have reported any problem if they had occurred):
https://travis-ci.org/debrief/debrief/builds/528189087

I've just had a look - I think it downloaded the JRE in 0.6 seconds!

It's _probably_ worth removing the lines that list the contents of the JRE, now that we know it works.

thx.

contribs/msi/make_x64msi.sh
saulhidalgoaular: Good evening. Yes, I wondered the same. They have named it as `jdk8u212-b03-jre` . but it is actually a `jre`.

saulhidalgoaular:
Good evening @IanMayo .

I have downloaded the msi from the release link you have provided, and it gives this error. 
image

However, I have generated the .msi by myself exactly as travis does, and it runs perfectly. 

image

Only difference I see is the msitool version. Travis has 0.98 and I had 0.99. I am updating it just in case. I checked in detail the travis log and it looked all good.

IanMayo:
The new version (0.99) still doesn't work for me.  The last version that worked for me was `20190329_3_0_473` - from 29th March.  I see that `msi` files stopped being included in the build after that date.  I _think_ the problem there was that the tutorial PDF had changed, but we hadn't updated our builder.

IanMayo:
Travis is still building broken .msi files, but ones built manually on Linux are valid.

PR #4141, opened 2019-05-08
4140 empty default folder for wmf export

### LINKED ISSUE #4140: Default Folder for WMF in Full Debrief
When I run the WMF export in the XY Graph section for Full Debrief, It shows the following error.

image

It suggests to check out the properties section, but there is no option about WMF Directory.

saulhidalgoaular:
Fix #4140


saulhidalgoaular:
I am just checking if the folder path is not empty, to avoid the error mentioned in the issue description. Maybe it would be good to have a Directory Chooser. I can add it if you consider it needed.

IanMayo:
@saulhidalgoaular - yes, we do need to introduce an `Options` dialog for D-Lite.

Amongst other things, this will contain the WMF export folder.

PR #4142, opened 2019-05-08
4117 Successful message for WMF export

### LINKED ISSUE #4117: Show message after producing WMF file of a graph
It would be very helpful if we show a message for the user after producing WMF file of his/her graph. Currently after selecting this "Produce WMF file of a graph" button , a dialog to select the location of the file will be shown and after that nothing is displayed.
Due to this  user has no any information whether the file is produced successfully or not unless the user checks the location.

saulhidalgoaular:
Fix #4117


IanMayo:
@helenayele - no.  `Successful` is the status we use in other operations.

Issue #4150, opened 2019-05-10
NullPointerException while resetting positions' property


helenayele:
After changing position's property and applying, trying to reset it will log NPE with this stack trace ;
[NPE_while_Resetting_position.txt](https://github.com/debrief/debrief/files/3165413/NPE_while_Resetting_position.txt)
### Steps ot reproduce:
1. Open boat1.rep
2. Open the first position's (120500) property 
3. Change its label from 120500 to 121500 and Apply 
4. Then before closing the property window - select Reset, this time it will throw NPE. And also the value will not be reset.
[![Image from Gyazo](https://i.gyazo.com/40b9f9ff4317324fbd9193135f1d6e37.gif)](https://gyazo.com/40b9f9ff4317324fbd9193135f1d6e37)

helenayele:
Verified

PR #4151, opened 2019-05-10
3762 popup editor

### LINKED ISSUE #3762: Properties window is not available for some layers in D-Lite
In D-Lite, we don't show any properties for some layers see the screenshots below
[Image from Gyazo](https://gyazo.com/12ba5a386f6fc284493c884dfc74f689)

### Steps to Reproduce
1. Open BULK_BLUE_TRACKS.rep on D-Lite
2. Expand Track:HVU  and double click 121213.14(97 positions) layer to open its property window
See that nothing is shown.
But in full debrief , its property is shown like this:
image

IanMayo:
Fixes #3762

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Properties/TimeFrequencyPropertyEditor.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Properties/TimeFrequencyPropertyEditor.java
@@ -163,7 +163,16 @@ public class TimeFrequencyPropertyEditor extends PropertyEditorSupport
 
   public Object getValue()
   {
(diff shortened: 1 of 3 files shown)


IanMayo:
@helenayele - could you check to see if i've missed out any Outline types that don't provide a popup editor?

IanMayo:
@helenayele - hopefully fixed

helenayele:
And also I tried to check by opening most of the rep and dpf file from sample_data and  I get all items's property window opened.

helenayele:
Yes , now able to approve my changes without any issue. But after approving my changes when I select reset it logs exception but this is filed as separate issue on #4150

PR #4161, opened 2019-05-12
4147 Changing Flow Layout to GridLayout

### LINKED ISSUE #4147: Change default Property window width
The default property window size is not enough to show all the available actions. 
For example when I open the property of a position from Nelson_A:0500 layer, it looks like this :
[Image from Gyazo](https://gyazo.com/3497bbc61251086c20f4857777aae4ba)

And when I increase the width , I can see one more hidden action 'Export Shape' . So the user will not have an idea about this action, and he/she can miss it.

[Image from Gyazo](https://gyazo.com/9d5afabb7a369d9da0db5cb7bca42113)

saulhidalgoaular:
Fix #4147 
Alternative 1. I have added this alternative, because might look neater. Bad thing is that sometimes buttons are too small to read their captions.


IanMayo:
Thanks for having a go, @saulhidalgoaular 

It's certainly more efficient use of space - but the button legibility does seem an issue.

Again - thanks for having a go :-)

PR #4162, opened 2019-05-12
4147 Changing Flow Layout to GridLayout

### LINKED ISSUE #4147: Change default Property window width

saulhidalgoaular:
Fix #4147 

This is the exact behavior @IanMayo described in the Issue.


IanMayo:
Let's wait & see what @helenayele  thinks :-)

PR #4182, opened 2019-05-20
specifically refreshed the changed layer in case of narratives

### LINKED ISSUE #4134: Refresh issues with large Narrative files
For very large narrative entries (> 1000) , after editing any of the entries and applying, D-Lite will take so much time to refresh and show the new updated value.

ayeshare:
Fixes #4134

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
@@ -896,83 +896,76 @@ public class SwingLayerManager extends SwingCustomEditor implements
     return null;
   }
   
-  private void updateData(Layer changedLayer)
+  private void updateThisLayer(final Layer changedLayer)
   {
-    TreeNode treeNode = getTreeNode(null,changedLayer.getName(),changedLayer);
-    if(treeNode != null) {
-      ((DefaultTreeModel)_myTree.getModel()).reload(treeNode);
+    final TreeNode treeNode = getTreeNode(null, changedLayer.getName(),
+        changedLayer);
+    if (treeNode != null)
+    {
+      ((DefaultTreeModel) _myTree.getModel()).reload(treeNode);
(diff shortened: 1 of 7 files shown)


IanMayo:
@ayeshare - I've done a bit of refactoring, so please pull my changes before you start work on @helenayele 's issue(s).

helenayele:
@IanMayo with your change now,  I am not able to reproduce the above issue 
[Image from Gyazo](https://gyazo.com/06688302063d3d97c79f5611b76b11fb)

IanMayo:
I fixed it? Well, I wasn't expecting that :-) . Let's wait for @ayeshare to verify I haven't broken her code.

helenayele:
sure @IanMayo

ayeshare:
@IanMayo It works fine. No regressions

IanMayo:
@helenayele - I'm going to merge it, since I don't expect the sea-riders to produce narratives with 1000s of entries.

helenayele:
okay

ayeshare:
> Thanks Ayesha, it is now very fast.
> But I see unnecessary undo and redo actions, clicking on undo/redo doesn't have any action seen for the user. is there any background work.
> 
> See the video, even if I click undo/redo 2 -3 times, I see nothing changed.

Yes, I am seeing this problem when some shapes are already present.  The problem is that the entire tree is refreshed after deleting a shape.

helenayele:
> Yes, I am seeing this problem when some shapes are already present. The problem is that the entire tree > is refreshed after deleting a shape.
But I see this even I don't delete anything the above step can be taken as example, I just change the time, and apply.

IanMayo:
@ayeshare: we introduced an optimisation for when a narrative is entered - when a narrative entry is modified, only update the narratives layer.

But, @helenayele has highlighted the problem that we have a performance problem when a non-narrative entry is modified.  _Maybe_ we need to look at the profiler and see what is causing the performance problem.  _Maybe_ there's another modification we can make - such as only modifying a tree node if it has changed.

PR #4186, opened 2019-05-21
4153 Specify Location

### LINKED ISSUE #4153: Specifying location in Debrief-Lite
Debrief-Lite deliberately has fewer controls than full Debrief.

But, we're missing a piece of functionality - changing the location for an item, such as the corner of a rectangle, or the location of a label:
image

In full Debrief we support this through several mechanisms:
1. drag feature mouse mode:  when in this mode, the cursor goes green when over something draggable, and it moves when dragged
2. copy location: the user can right-click on the plot, select "Copy cursor location", then "Paste location" over a location attribute in the property window
3. property editor: the user can expand the location in the property editor to reveal the individuals fields for the location (lat degs, mins, secs, etc).  This is the way exact locations get entered - for example if they've come from an email

In D-Lite, we don't have the drag mouse-mode (1).   Since our property editor is modal, the user can't then right-click on the map to copy a location (2).  Debrief-Lite does have a popup property editor:
image

(it's old & clunky, but it works)

So, we're still missing the ability to set a location to be some point on the map (as with 1. and 2.).

saulhidalgoaular:
Fix #4153


IanMayo:
thanks for making a start @saulhidalgoaular 

But, the right-click to copy location isn't the preferred solution, since the Properties Window opens as a modal window - so the user can't right-click a point on the map.

So, the better solution (as recorded in the later comment on the issue) is to support drag/drop mouse modes (Drag Point and Drag Shape).

saulhidalgoaular:
Good morning @helenayele . I think it is ready for testing.

IanMayo:
@saulhidalgoaular  - we have another outstanding item on this.  The cursor should change as it goes over a selectable item.

I've added logic to `DragWholeFeatureElementTool` to decide when to change the cursor.  But, I haven't implemented changing the cursor.

_Maybe_ it's worth using the red/green hand icons from full-Debrief.  I think the hand reflects `Drag` more than the current popup properties window.

The logic isn't straightforward regarding how to change the cursor, or I would have done it already ;-)

But, it certainly looks achievable.

helenayele:
I am able to use drag element for Circle but can't use it for ellipse shape at all.
[Image from Gyazo](https://gyazo.com/69b0cf233e0941c0b1bcff2453d0c66e)
But even for circle it is not that much smooth. i.e I am able to drag element after so many tries, see the below video, even if I try my best I couldn't expand the circle within 7 seconds.
[Image from Gyazo](https://gyazo.com/0c295cba12cd52723c98deb59569b519)

helenayele:
On this fix, drag feature works perfect on all shapes including circle with exception of ellipse.
[Image from Gyazo](https://gyazo.com/733e4e25dd46efc996dd8d3da6db2766)
### In general I have only 2 comments:
1. Drag element is not working for Ellipse shape
[Image from Gyazo](https://gyazo.com/580e5f0cb444594f71aac50ca5fb2707)
2. Drag Whole feature works even if the cursor color is not changed to Green 
[Image from Gyazo](https://gyazo.com/8aaaec5b16629234c124343c4b2ae9d3)

helenayele:
See the behavior of Drag whole component on rectangle in Full debrief
[Image from Gyazo](https://gyazo.com/ae8eb9339eae56869a8f4b7924751ee1)

The shape will not move unless cursor hand color turns to green.

helenayele:
Sometimes when I try to resize polygon from second node(2) it will not resize it .
[Image from Gyazo](https://gyazo.com/beef6061b14d7a706172872151625863)

helenayele:
The last issue can be handled in separate bug #4226 . So approving..

PR #4192, opened 2019-05-22
fixed regression

### LINKED ISSUE #4191: New shapes not shown when inserted
If we insert a new shape (such as a rectangle) it only gets shown on the plot if we have pre-selected the target layer.

Steps to reproduce:
* open boat1.rep
* switch to _insert_ tab
* click on rectangle
* select target layer (`Chart Features`)
* observe that rectangle not shown

But, if we select `Chart Feature` from the target layer drop-down it does appear.

ayeshare:
Fixes #4191

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
@@ -204,7 +204,6 @@ import MWC.GUI.Properties.PlainPropertyEditor;
 import MWC.GUI.Properties.Swing.SwingCustomEditor;
 import MWC.GUI.Tools.Chart.RightClickEdit;
 import MWC.GUI.Tools.Chart.RightClickEdit.PlottableMenuCreator;
-import MWC.TacticalData.NarrativeWrapper;
 
 public class SwingLayerManager extends SwingCustomEditor implements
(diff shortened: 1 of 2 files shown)


IanMayo:
Strange @ayeshare - we're removed an update call, which forces an update?

ayeshare:
> Strange @ayeshare - we're removed an update call, which forces an update?

Please note I removed the interface implementation datalistener2

the datalistener2 had a method with newitem and the layer.  The bug was in this code.

The correct fix was to add the new node (for the item added), if it was not found, instead of recreating the entire tree.

IanMayo:
@ayeshare - I believe you're going to be making some modifications to this PR.

helenayele:
Now able to open sample.dpf file without any issues, but there is performance issue while inserting shapes to a layer- it takes long time (5-6 seconds) to insert a shape , see the below video:
[slow_while_inserting_shapes.zip](https://github.com/debrief/debrief/files/3216289/slow_while_inserting_shapes.zip)

ayeshare:
> -6 seconds) to insert a shape , see the below video:


I tried with sample.dpf, the shapes are inserted instantly.  I tried adding, boat1.rep,boat2.rep,sample.dpf,shapes.rep and then inserted shapes, it still added instantly.

IanMayo:
I'm going to merge this to reduce the amount of un-merged code.

We may have a performance issue on Helen's machine. We'll have to defer the performance issue until we have a user with a PC on which the problem occurs.

PR #4211, opened 2019-05-28
avoid print outs

theanuradha:


IanMayo:
thanks for this @theanuradha  .  But, if unit tests fail, do we still see the results in Travis?

theanuradha:
on failed cases result will be shown with what test failed. if you need the log with details maybe we can use same technique we used in rcptt log submit

PR #4220, opened 2019-05-30
4209 Track Not Found message.

### LINKED ISSUE #4209: D-Lite fails silently to import sensor data on wrong track
When user tries to import sensor data on wrong track , D-Lite reports that the import was successful by showing this dialog.
[Image from Gyazo](https://gyazo.com/8d76315875d00be8deb2269203edafbd)

 But if user tries to load wrong sensor data for the track, D-Lite doesn't import the sensor data, instead it logs this on the console:

"Error:Not storing sensor, track not found" 
But Full debrief does report that the sensor data doesn't get the correct track by showing "Sensor Track not found dialog" :
[Image from Gyazo](https://gyazo.com/9c0282ee0b9b7656ce45a42704a320b6).
Even if Full debrief reports this, if I click okay button of this dialog by selecting one of the track, the sensor data is imported like this 
[Image from Gyazo](https://gyazo.com/7e4a4813144d6091151f00e5bf2e03b9)
I don't think Full debrief should load this sensor data in invalid track name, and I think this should be improved in full debrief also.
### Steps to reproduce:
1. Launch D-Lite
2. Open samplescale.dpf
3. Drag sen_frig_sensor dsf

saulhidalgoaular:
Fix #4209

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
@@ -121,7 +121,7 @@ import MWC.Utilities.ReaderWriter.PlainImporter;
 
 public class DebriefLiteApp implements FileDropListener
 {
-  
+
   /**
    * introduce a preferences helper, particularly to give default font sizes
    *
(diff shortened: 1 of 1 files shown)


saulhidalgoaular:
Good morning @helenayele .
I think it is ready for testing.

PR #4232, opened 2019-05-31
4229 added show Error log view support

### LINKED ISSUE #4229: Include link to Error Log in dialog
For one data import, we won't fail on the first error, we'll complete the import, then direct the analyst to view the import failures in the Error Log.

I could put "Please view the errors in Window > Show View > Error Log".  But, it would be great if the Error Log could be clickable.

A fallback would be to include a "Close and view Error Log" button on the dialog popup.

If this is possible, it should be added to the SWTEclipseHelper class.

theanuradha:
Fixes #4229


IanMayo:
Closing. Solution applied manually in target branch.

PR #4236, opened 2019-05-31
4018 Correct color

### LINKED ISSUE #4018: Set the background default color of highlighter to white
Currently when I select green color for highlighter I get magenta applied on the hughlighter, and when I select  magenta color, I see green color is applied on on the highlighter.
[Image from Gyazo](https://gyazo.com/e6a9e300eadb249d72284abc8b268f7c)

[Image from Gyazo](https://gyazo.com/b25b7aca55a6c4b852ea00f694d44ef2)

saulhidalgoaular:
Fix #4018


saulhidalgoaular:
Good afternoon @IanMayo .

I am passing now the rectangle default color as parameter, because It cannot be white in Debrief Lite. It is getting defined in the StepControl for now, but if you prefer to define it at any level above, please tell me.

Greetings.
Saul.

PR #4254, opened 2019-06-06
handled alphanumeric label in read polygon

### LINKED ISSUE #3872: Trouble with variable length lists (e.g. ;POLYGON) when label starts with digit
The Polygon and polyline formats allow for a variable length.  Debrief determines that there are no more coordinate pairs when it recognises a non-numeric label (see below).

````
;POLY: @@ YYMMDD HHMMSS YYMMDD HHMMSS DD MM SS.S H DDD MM SS.S H DD MM SS.S H DDD MM SS.S H DD MM SS.S H DDD MM SS.S H  XX...XX
;POLY: @@ DD MM SS.S H DDD MM SS.S H DD MM SS.S H DDD MM SS.S H DD MM SS.S H DDD MM SS.S H  XX...XX
;; symb, start date (optional), end date (optional), lat & long 1, lat & long 2, lat & long xxx (until end of line, or non-numeric label encountered)
````

We should make it more capable - so it can allow a label that starts with a digit.

Maybe fancy regexp is needed.

### PR DESCRIPTION
**Pull request by ayeshare**
Fixes #3872

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportPolygon.java
+++ b/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportPolygon.java
@@ -165,70 +165,67 @@ public class ImportPolygon extends AbstractPlainLineImporter
 		Integer counter = new Integer(1);
 		// create the Polygon object
 		final PolygonShape sp = createShape(nodes);
-
 		while (st.hasMoreTokens())
 		{
-			// meet the label
-			final String sts = st.nextToken();
+		  // meet the label
+		  final String sts = st.nextToken();
+		  if (!isNameLabel(sts) && Character.isDigit(sts.charAt(0)))
+		  {
+		    try
+		    {
+		      // now the location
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by IanMayo** [comment]
Hey, @ayeshare - the tests look good :-)

The Travis build is breaking because of tests in another class. Sadly they appear to be timezone-sensitive.  That's my problem.

I'll be looking at this, later today.

File: org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportPolygon.java [hunk 1]
```diff
@@ -169,7 +169,7 @@ else if (hasStartEndDates(line))
 		{
 		  // meet the label
 		  final String sts = st.nextToken();
-		  if (!isNameLabel(sts) && Character.isDigit(sts.charAt(0)))
```
IanMayo: This may not have an immediate benefit to REP file import, but I really like the fact that you've identified and implemented the optimisation.

PR #4263, opened 2019-06-12
Custom implementation of zoom tools

### LINKED ISSUE #3948: Reverse drag to zoom out
When in Zoom mode in full Debrief, the user can drag from bottom-right to top-left to perform a zoom out operation.

This reduced the amount of switching between mouse modes.

See if the GeoTools zoom tool could be extended to optionally include this mode.

The zoom out operation is interpreted as:
> the user is selecting the region he wants the current perspective plotted into

So, if the user draws br-tl rectangle at the top-right of the current plot, the current viewport will shrink to occupy that rectangle, and a large new area of coverage to the South-West is now displayed.

** Note **

In #4263 Ian has provided a partial implementation of this feature.  It zooms out proportionally to the selected area, but it has one flaw.  In full Debrief the old viewport centre is moved to the centre of the selected area.  In this Debrief-Lite implementation the new zoomed out area is centred on the centre of the selected area.

ayeshare:
Fixes #3948

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/MyZoominTool.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/MyZoominTool.java
@@ -1,0 +1,79 @@
+package org.mwc.debrief.lite.gui;
+
+
+import java.awt.Point;
+import java.awt.geom.Point2D;
+
+import org.geotools.geometry.DirectPosition2D;
+import org.geotools.swing.MapPane;
+import org.geotools.swing.event.MapMouseEvent;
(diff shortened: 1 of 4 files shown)


IanMayo:
The full Debrief implementation has this behaviour:  when the mouse is dragged bottom right to top-left, the current plot is zoomed out so that the currently visible area is compressed down into the area selected with the mouse.

The implementation in this PR just zooms out by 50%.

I've just had a couple of hours [no, correction - **it's now 4 hours**] trying to do the mapping for the full Debrief implementation, and I've got myself completely stuck.  I'm going to defer the issue until we have a team member with deep GeoTools skills.

PR #4264, opened 2019-06-12
fixed build script to remove references to hardcoded build date

### LINKED ISSUE #4258: D-Lite build should use valid version number
The D-Lite build should take its version number from the `MANIFEST.MF`.  But, currently it's using a hard-coded value.

The build users a hard-coded filename when it deletes existing versions. I guess we'll have to use a wildcard instead.

ayeshare:
Fixes #4258

### DIFF
--- a/org.mwc.debrief.lite/build.xml
+++ b/org.mwc.debrief.lite/build.xml
@@ -120,8 +120,8 @@
 	<target name="clean" depends="quiet,init" description="Clean the plug-in: org.mwc.debrief.lite of all the zips, jars and logs created.">
 		<delete file="${build.result.folder}/debrief_lite.jar"/>
 		<delete file="debrief_lite.jar"/>
-		<delete file="${plugin.destination}/org.mwc.debrief.lite_${product.version}.jar"/>
-		<delete file="${plugin.destination}/org.mwc.debrief.lite_${product.version}.zip"/>
+		<delete file="${plugin.destination}/org.mwc.debrief.lite_*.jar"/>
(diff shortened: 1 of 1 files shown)


org.mwc.debrief.lite/build.xml
```diff
@@ -119,8 +120,8 @@
 	<target name="clean" depends="quiet,init" description="Clean the plug-in: org.mwc.debrief.lite of all the zips, jars and logs created.">
 		<delete file="${build.result.folder}/debrief_lite.jar"/>
 		<delete file="debrief_lite.jar"/>
-		<delete file="${plugin.destination}/org.mwc.debrief.lite_1.0.377.jar"/>
-		<delete file="${plugin.destination}/org.mwc.debrief.lite_1.0.377.zip"/>
+		<delete file="${plugin.destination}/org.mwc.debrief.lite_${product.version}.jar"/>
```
IanMayo: @ayeshare - won't this try to delete any files with the current version?  It's more likely that we will have files that are for the previous product version.  I think we need a wildcard in this delete statement, to allow for all product versions,

IanMayo:
@ayeshare - I still think we have more work to do on the Ant build.   We seem to have quite an inefficient build sequence (attached).
[out.txt](https://github.com/debrief/debrief/files/3296270/out.txt)

I'm pretty sure we can organise the build so fewer steps get repeatedly called.

Note: I've modified the build script in two ways:
* change the default target to the one we use to build D-Lite
* comment out the line that hides the logging, so we can understand whats' happening.

IanMayo:
Close. Fixed in other branch

PR #4265, opened 2019-06-12
introduce new capability - proportional zoom out

IanMayo:
Supports #3948 

We proportionally drag out ok.  

But, to match Full Debiref, the old centre should be placed in the centre of the dragged area.

This implementation puts the centre of the dragged area as the new centre of the viewport.


helenayele:
Sure, I try to put the results of the same zooming out action as follows. 
[Image from Gyazo](https://gyazo.com/acf4ef2a05311007cd0fed62e988d194)

I tried to zoom-in like this
[Image from Gyazo](https://gyazo.com/b65e1eeb8f3acaa1db55fb0ddf92e3df)

IanMayo:
Ok, @helenayele - now I understand the difference.  We will occasionally get these small differences between the versions.  The slightly different "safety border" around the selected area isn't a problem.

PR #4266, opened 2019-06-12
Added builddate.java and replaced build date

### LINKED ISSUE #4257: Build step to update Title-bar date
D-Lite includes a build date in the title bar:
image

I'm pretty sure this is hard-coded somewhere.  But, I'd like it to be actual date the build is performed.

If I remember correctly, we're doing the D-Lite build in Ant.  So, I think we need something like this:
* create a BuildDate.java class, with $DATE_HERE$ string const
* generate string from current date (in Ant)
* replace some $DATE_HERE$ with the date string (in Ant)
* at run-time, set the title text as `BuildDate.toString()` (or something like that).

So, during development we'll see $DATE_HERE^ in the title-bar, but that's ok.

ayeshare:
Fixes #4257

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/BuildDate.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/BuildDate.java
@@ -1,0 +1,24 @@
+/*
+ *    Debrief - the Open Source Maritime Analysis Application
+ *    http://debrief.info
+ *
+ *    (C) 2000-2016, Deep Blue C Technology Ltd
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the Eclipse Public License v1.0
+ *    (http://www.eclipse.org/legal/epl-v10.html)
+ *
+ *    This library is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/BuildDate.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package org.mwc.debrief.lite;
+
+/**
+ * @author Ayesha
+ *
+ */
+public class BuildDate
+{
+  public static final String BUILD_DATE="12 Jun 2019 17:30";
```
IanMayo: @ayeshare - I think you're committing the modified file. This one hasn't got the `$DATE_HERE$` marker.

Hmm, I think we're going to keep on getting this issue.  I've got an alternate suggestion:
* we introduce a file titled "BuildDateMaster.tzt" - it contains the current contents of BuildDate.java
* in the build, we make a copy of the file, and rename it to BuildData.java
* we make the change to this file
* we can then mark BuildDate.java as `ignore` in Git - so it won't get flagged up as a changed document after each release.

IanMayo:
Thanks @ayeshare - we've got some twisted build dependencies that mean we replace the build date lots of times.  Hopefully this PR and the other `build.xml` PR can be merged tidily, and a refined build process created.

I'd also like you to switch away from the deprecated build step, please:
image

IanMayo:
Close. Fixed in other branch

PR #4277, opened 2019-06-14
added support for nmea files

### LINKED ISSUE #4271: Need to allow NMEA import in D-Lite
While discussing the July release with the scientists, it became clear that it would be valuable if sea-riders could drag/drop NMEA data files into Debrief-Lite.

There's a sample file in `sample_data \ other_formats \ NMEA_trial.log`

In addition to drag/drop onto an existing plot, we should add it to the `Import` types.

The legacy class to perform this import is in `Debrief.ReaderWriter.NMEA.ImportNMEA`

ayeshare:
Fixes #4271

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/NMEA/ImportNMEA.java
+++ b/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/NMEA/ImportNMEA.java
@@ -1015,7 +1015,6 @@ public class ImportNMEA
       {
         contactHolder = new BaseLayer();
         contactHolder.setName("WECDIS Contacts");
-        _layers.addThisLayer(contactHolder);
       }
 
       // ok, build the track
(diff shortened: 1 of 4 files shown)


IanMayo:
Yes, @helenayele - there is an issue with the time slider coverage.

@ayeshare - you'll have to look at how we handle importing `boat1.rep` - because this does set the time controller bounds properly.

IanMayo:
I'm actually going to merge this, to avoid conflicts.

@helenayele - could you create an issue to cover the `Time Controller` problem?

helenayele:
1. File name: we don’t need to set it for WECDIS data
2. Console debug: I’ve removed those lines
3. Save as .dpf: yes, that’s intended behaviour
4. The time controls should be enabled - so the sea-rider can replay the tracks.. But, they’re currently not going to the right time period is filed at  #4282

PR #4309, opened 2019-06-28
make map and track use mercator projection

ianturton:


IanMayo:
Merging into parent branch, in order for further minor mods to be conducted.

PR #4320, opened 2019-07-04
changed field smallstep to long

### LINKED ISSUE #4314: Small step with 1 hour value is not retained
If a user tries to configure small step  = 1 hour and apply, the value is not retained.

#### Steps

1. Launch D-Lite and open boat1.rep file
2. Open time menu
3.Open  Edit time properties and set Small step = 1 hour
4. Apply and close the property window
5. Open this edit time properties window and check the value of small step
See that it is 1/10 sec instead of 1 hour.
[Image from Gyazo](https://gyazo.com/e27af95f7ec7cd2f3461c41df319f1b6)

ayeshare:
Fixes #4314


IanMayo:
@ayeshare - we've got compiler problems for this branch.  I'm having a look.

Issue #4099, opened 2019-04-30
M1.6.2.5 View Diary (Narrative)


IanMayo:
As a ***sea-rider*** I wish to ***review recorded text comments*** so that ***I can understand more about the plot I'm looking at***.

## Background
In addition to GPS tracks, analysts and sea-riders can receive a range of other timestamped textual observations. These may be captured by a sea-rider during an exercise, by a member of staff on board the ship (like a Ship's Log), or from some automated system on board.

These observations can indicate the decision-making process of one of the participants, or some unusual event on board (such as a system crash), or an observation that isn't otherwise available to the analyst (such as weather observation).

The name for a single item in the narrative is a **narrative entry**.

## Implementation
Debrief has a mature narrative format for `.REP` files, which allows an optional category indicator that further helps the sea-rider to find data of interest.

In agreement with the customer, we're dedicating a right-hand panel of Debrief-Lite to the display of Narrative data. 

Hopefully we'll be able to use some modern UI layout that makes the information easier to read, maybe like a series of tweets.

## Capabilities
The narrative view should have these capabilities:
* highlight the entry nearest to the current Debrief time
* use formatting to indicate entries that come from different sources
* ability to make plot jump to the time of an entry
* ability for sea-rider to add new individual narrative entries (probably taking time from current Debrief time)
* ability for sea-rider to transcript bulk narrative entries (maybe being taken from a handwritten notebook)
* ability to show an icon/formatting for particular types of narrative entry (maybe)
* ability for narrative entries to be deleted
* ability to mark some entries as "Pause Points".  When Debrief is playing, it will pause when it hits a pause point (more here: #4036 )
* export one or more narrative entries to the clipboard

clenoble:
Link to current wireframes (slides 19 to 22): [debrief-lite-narrative-panel](https://xd.adobe.com/view/b84612dd-e964-4322-699d-90574b10b452-ce77/screen/85c60d0d-54dd-4987-93b8-3f6bd26a1199/Debrief-lite-Narrative-panel-Filters-Pause-Format-)

PR #4323, opened 2019-07-04
4099 narrative jtable

saulhidalgoaular:
Supports #4099


IanMayo:
@saulhidalgoaular - you need to disable double-click to edit on the table cells:
image

In Phase 2, double-click to set time may fix the above.  But, let's get this issue fixed in Phase 1, just in case we don't get to Phase 2.

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/narratives/NarrativeEntryItemRenderer.java
```diff
@@ -0,0 +1,123 @@
...
+      final JPanel innerPanel = new JPanel();
+      innerPanel.setLayout(new BorderLayout());
+      innerPanel.add(header, BorderLayout.NORTH);
+      innerPanel.add(name, BorderLayout.CENTER);
+      mainPanel.add(innerPanel, BorderLayout.CENTER);
+
+      JLabel highlightIcon;
+      if (isSelected)
+      {
+        highlightIcon = new JLabel(Utils.getIcon("icons/16/highlight.png"));
```
IanMayo: @saulhidalgoaular - I think it would be quicker to create instances of these 4 icons, as final fields, retrieved, in a no-arg constructor.

What do you think?

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/narratives/NarrativePanelToolbar.java
```diff
@@ -237,22 +233,20 @@ public NarrativePanelToolbar(final LiteStepControl stepControl,
...
     {
 
       @Override
       public void componentResized(final ComponentEvent e)
       {
-        narrativeEntryItemRenderer.setPanelWidth(_narrativeList.getWidth());
-        narrativeEntryItemRenderer.setWrapping(model.isWrapping());
-        _toolbarInstance.updateRowHeights();
+        model.setPanelWidth(_narrativeList.getWidth());
+        model.setWrapping(model.isWrapping());
```
IanMayo: @saulhidalgoaular  - do we need to `setWrapping` when the component has resized?

Issue #4341, opened 2019-07-09
Button collapse all panels.


IanMayo:
In the title bar of the app, we have undo and redo buttons.

It would be great to have a 'collapse all' button in there, too.

When clicked the Outline, Graphs and Narratives panels will all collapse.  When clicked again, they will expand.

The button doesn't need to toggle, or remember it's state.  If any panel is open, they're all collapsed.  If no panel is open, they're all expanded to their previous state (similar to the double-click action).

I don't think we have an icon for this right now.  So, temporarily we'll just use the fit-to-window button.

## Update

Aah, there's an improvement to the above processing. If, each time we perform "collapse", we remember which panels actually got collapsed, then when we perform "expand", we can simply expand the items in that list.  

This will act as "memory" - restoring the state of the panels before they were collapsed.

But, if, while collapsed, a single panel is opened, then the next time the button is clicked, that single panel will be collapsed, and that single panel will be opened on the next click.

PR #4348, opened 2019-07-09
4341 collapse panels

IanMayo:
Supports #4341


IanMayo:
@helenayele - so, it has a different icon depending upon if it's going to expand or collapse the panels.

Sure, that's possible.  But, the state of the panels can be seen.

Aah, there's another issue.  If the user presses "Collapse", then the icon will switch to 'expand'.  Imagine the user then drags the Outline open. The required behaviour for this is that if any panel is open when the button is clicked, they will all collapse.

So, right now, the icon is "expand" - but clicking on it will perform "collapse".  So, I think we need an icon that doesn't indicate state.

helenayele:
Okay understood, can we have tool-tip please?

helenayele:
While collapsing/expanding, if the narrative panel was collapsed before, it will be somehow hidden when we select collapse button.
[Image from Gyazo](https://gyazo.com/5767c20981f51ee50a05c55944353ed1)
### Steps:
1. Launch D-Lite 
2. Collapse Narrative panel 
3. Click Collapse/expand button multiple times -

IanMayo:
@ayeshare - yes, this is incomplete.  But, I'm going to merge this PR as it is, to get customer feedback, then carry on in another PR.

helenayele:
@IanMayo I think you should mean Me(@helenayele )?

PR #4354, opened 2019-07-10
4340 create shape projection

### LINKED ISSUE #4340: Failure in projection transform error while inserting shapes
When I insert shapes like label,etc, I get the following exception being thrown  and the shape is also not shown on the map but it is listed in outline view. 
[Image from Gyazo](https://gyazo.com/062130c064ac8b93f406eadb5a1fffc0)

[Failure in projection transform.txt](https://github.com/debrief/debrief/files/3373526/Failure.in.projection.transform.txt)

IanMayo:
Fixes #4340


IanMayo:
@helenayele  - this was a scary one, but it looks like I'm on top of it.

Issue #4345, opened 2019-07-09
Snail mode fade colours


IanMayo:
When we have the `sample.dpf` tracks in snail mode, the shade doesn't blend to the water color, it seems to shade to white.  We also struggle with the XOR rendering overwriting (and thus deleting) itself.

Plus, the color change is in quite large steps, not in small continuous ones.

It would be good if we could improve this.

Here's an example:
![image](https://user-images.githubusercontent.com/1108513/60996636-e9fc9f80-a34c-11e9-86c6-738a6cee2a82.png)

IanMayo:
The fade colours in snail still don't work. Can we turn off fade?

IanMayo:
Ok, have switched to non-XOR rendering.

I was worried that GeoTools would incur a large cost in redrawing Natural Earth, or background TIFs, but it caches them - only redrawing them on viewport change.

Here's a demo:
![image](https://user-images.githubusercontent.com/1108513/60996764-37790c80-a34d-11e9-8eac-71ee723263e6.png)

PR #4362, opened 2019-07-10
4345 snail colors A

IanMayo:
Supports #4345 

Improve effectiveness of snail trail plotting, by optimising XOR rendering


IanMayo:
Drop this, the other method proved better

PR #4389, opened 2019-07-11
4337 - Adding narratives to a Base Layer.

### LINKED ISSUE #4337: Narratives are not loaded if we have narratives layer in outline panel

saulhidalgoaular:
Fix #4337

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/BaseLayer.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/BaseLayer.java
@@ -202,6 +202,11 @@ public class BaseLayer extends Plottables implements Layer, SupportsPropertyList
 	 */
 	private final boolean _orderedChildren;
 
+  /**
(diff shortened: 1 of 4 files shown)


IanMayo:
Thanks for the PR, @saulhidalgoaular. When you get chance, could you give me an overview of the approach you've taken?

There are lots of changes in the code.

org.mwc.cmap.legacy/src/MWC/GUI/BaseLayer.java
```diff
@@ -202,6 +202,11 @@
 	 */
 	private final boolean _orderedChildren;
 
+  /**
+   * property type to signify data being added or removed
+   */
```
IanMayo: Hmm, @saulhidalgoaular we don't refer to this string - doesn't that suggest that no-one is listening to it?

IanMayo:
Closing. Have produced simpler solution in #4406

PR #4391, opened 2019-07-12
remember collapsed state of panels

### LINKED ISSUE #4341: Button collapse all panels.

ayeshare:
Fixes #4341

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
@@ -972,7 +972,7 @@ public class DebriefLiteApp implements FileDropListener
         alphaListener, alpha, path);
   }
 
-  public void savePanelState()
+  private void savePanelState()
   {
     openPanelNames.clear();
     if(!outlinePanel.isCollapsed()) {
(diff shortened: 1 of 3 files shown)


helenayele:
But if graph panel was opened while in collapsed state, when I click on this collapse/expand button, the graph panel remains there - will not be closed. and it will be closed when I click it again. 
Before this fix, it was closed on my first click.

[Image from Gyazo](https://gyazo.com/b444fea3f14d100abda90ab4c9b901f1)
see in this video, first it was in collapsed state , and I opened the graph panel, then When I click on collapse/expand button , the graph panel is not closed. When I click it again, it gets closed.

IanMayo:
@helenayele - would you mind retesting?  I changed the logic a bit

Issue #4350, opened 2019-07-10
Plot doesn't refresh immediately after zoom-in/zoom-out


helenayele:
After opening a plot , I try to zoom-in/out but the plot takes more than a minute  to apply the operation.

### Steps to reproduce:
1. Launch D-Lite
2. Open sample_lots_of_sensors.dpf file
3. Select zoom-in button on the view menu
4. See that it takes more than one minute to zoom-in

IanMayo:
With the sensor data visible, I would expect a delay. 

But, without sensor data visible, zoom/pan are instantaneous for me.

Could you re-test?

helenayele:
It takes around 43 seconds to do zoom-in operation on sample.dpf file here.
[zoom-in_on_sample_file.zip](https://github.com/debrief/debrief/files/3377463/zoom-in_on_sample_file.zip)

IanMayo:
Helen - is that with your _new_ laptop?

Are any errors being logged?

helenayele:
Sure..yes it is with the fastest and latest laptop
[sample.zip](https://github.com/debrief/debrief/files/3377495/sample.zip)
. 
No errors are logged on the console.

IanMayo:
Aaah, it's that Chart Lib layer.  If you switch that off, it's instantaneous.
![image](https://user-images.githubusercontent.com/1108513/60971113-5b235f00-a31b-11e9-9114-7f7beb096d0e.png)

I can only imagine each feature in it is triggering a redraw.

helenayele:
By the way the issue is on/off . First I tried with boat1.rep file both zooming-out and fit-to-window works perfect, then I opened sample.dpf file and when I try to zooming-out, it takes too much time to effect the zooming. Then When I try with boat1.rep again, it becomes again slow.

IanMayo:
Ok. I think I've got it.  

After doing `Close` on a plot, then opening a new file, the plot no longer updates properly.  You have to trigger an update by resizing the panel, or something like that.

helenayele:
The issue is still reproducible
Steps:
1. Open boat1.rep
2. Close using close button
3. Open boat2.rep file
4. Select Zoom-out
See that it doesn't respond quickly - it takes some time

helenayele:
Sometimes, zooming-out doesn't work for this file 
[boat1.zip](https://github.com/debrief/debrief/files/3386111/boat1.zip)
and logs this
2019-07-12T13:33:43.639-0700  SEVERE  null
[...]

PR #4402, opened 2019-07-13
4350 zoom fail on DPF file

IanMayo:
Clear mapcontent before clearing layers - so FeatureLayer objects still have their feature source.

Don't reload background coastline


IanMayo:
Fixes #4350

PR #4413, opened 2019-07-14
4337 Special Case for Full-Debrief

### LINKED ISSUE #4337: Narratives are not loaded if we have narratives layer in outline panel

saulhidalgoaular:
Fix #4337
Full-Debrief Case

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/creators/shapes/CoreInsertShape.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/creators/shapes/CoreInsertShape.java
@@ -17,9 +17,7 @@ package org.mwc.debrief.core.creators.shapes;
 import java.util.Date;
 import java.util.Enumeration;
 
-import org.eclipse.core.internal.runtime.Activator;
 import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.MultiStatus;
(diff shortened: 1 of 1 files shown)


org.mwc.debrief.core/src/org/mwc/debrief/core/creators/shapes/CoreInsertShape.java
```diff
@@ -118,12 +123,35 @@ protected String getLayerName()
...
+              final String messageCannotBeEmpty = "Message cannot be empty";
+              final String messageCannotBeEmpty2 = "You need to have at least one character in your layer name";
+              if ( res == null || res.isEmpty() )
+              {
+                Status status = new Status(IStatus.ERROR, Activator.PLUGIN_ID,  messageCannotBeEmpty2, new Exception(""));
+
+                ErrorDialog.openError(Display.getCurrent().getActiveShell(), title, messageCannotBeEmpty, status);
+                
+                res = null;
+              }else if ( "narratives".equals(res.toLowerCase()) )
```
IanMayo: let's not use a magic string, let's use the constant that's defined in `NarrativeEntry`

saulhidalgoaular: oh, ok ok

PR #4421, opened 2019-07-15
4337 Do case insensitive string comparison

### LINKED ISSUE #4337: Narratives are not loaded if we have narratives layer in outline panel

IanMayo:
Fixes #4337


IanMayo:
Ok, we can extend the Base Layer behaviour to avoid this. But, we're getting into a really unlikely scenario.

helenayele:
In this fix, I am not allowed also to rename existing layer with "Narratives" name.

PR #4422, opened 2019-07-19
fixed ant build

### LINKED ISSUE #4419: D-Lite Ant build
We can't run the D-Lite build from the command line.

I _think_ it's because it's using some old .jar files in `org.mwc.cmap.geotools`.

We need to fix this, so that new D-Lite releases can be build locally.

ayeshare:
Fixes #4419


IanMayo:
Hey, @ayeshare - the Travis build is breaking. But, it looks more like a Travis config issue than our code.  Let's leave it a few hours, and see if Travis makes it go away.

```
$ ~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"
Ignoring license option: BCL -- using GPLv2+CE by default
install-jdk.sh 2019-07-17
Expected feature release number in range of 9 to 14, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .

```

IanMayo:
@ayeshare - separate to the Travis issue, the `ant` build is still failing for me (see attached).

I'm on my new laptop, which hasn't had a maven build performed yet - so the compiled files aren't available.

But, for a _real_ build, that we have confidence in, we need to do a fresh build from `ant` - to be sure it's using the newest versions. 

I _guess_ you could trigger the compilation of the _upstream_ plugins. Or, I guess you could just add the upstream source folders - so they're available to the Ant compilation.

Aah, but, hold on.  I think we distribute `org.mwc..xxx.legacy` files with D-Lite.  So, we can't just take the _current_ copy of the .jar and use it in the dist.  We do need to generate fresh ones.

Aah, hold on (again).  _Maybe_ we just perform a top-level `mvn clean verify` - which does a `clean` - and then we can be sure that the d-lite `ant` build has used the newest versions.  Is that the best strategy?

ayeshare:
> @ayeshare - separate to the Travis issue, the `ant` build is still failing for me (see attached).
> 
> I'm on my new laptop, which hasn't had a maven build performed yet - so the compiled files aren't available. Ant error listing attached.
> [list.txt](https://github.com/debrief/debrief/files/3411324/list.txt)
> 
> But, for a _real_ build, that we have confidence in, we need to do a fresh build from `ant` - to be sure it's using the newest versions.
> 
> I _guess_ you could trigger the compilation of the _upstream_ plugins. Or, I guess you could just add the upstream source folders - so they're available to the Ant compilation.
> 
> Aah, but, hold on. I think we distribute `org.mwc..xxx.legacy` files with D-Lite. So, we can't just take the _current_ copy of the .jar and use it in the dist. We do need to generate fresh ones.
> 
> Aah, hold on (again). _Maybe_ we just perform a top-level `mvn clean verify` - which does a `clean` - and then we can be sure that the d-lite `ant` build has used the newest versions. Is that the best strategy?

@IanMayo Tested this, it seems to be working fine. I did a full clean removing all traces of all 3 .jar files and building again and it works fine.  Did you work on this branch?

IanMayo:
@ayeshare - I've just done `mvn clean` followed by `mvn clean verify` - and it's still failing:
image

Can you tell me more about how you did `full clean`?  I guess we need to include that in the Ant build.

Issue #4418, opened 2019-07-14
D-Lite test implementation


IanMayo:
Once Helen has produced a shopping list of tests, this task is to start developing them, including adding them to our automated build process.

ayeshare:
the test are run as part of all target in the build.xml under legacy project. I tried to add lite tests under debief.test2 project in the AllTests.java, but they never launched. There was a problem importing zoomintool.java so I had to remove that(temporarily) to test launching the tests from maven.  Even the setting classpath was flaky.

ayeshare:
@IanMayo After upgrade to flamingo 2.5.1, the tests are not working.

PR #4433, opened 2019-08-06
Added a test for launching application

ayeshare:
Supports #4418


IanMayo:
A nice start @ayeshare.  

As the codacy review indicates, your test should include at least one `assert` or `fail` call - which is quite valid.

I guess you need to build on the test to verify that data is present.  One way is to gain access to the `Layers` object directly.  Another way is to access the tree in the Outline view.  I guess both are valid.

org.mwc.debrief.lite/src/test/java/org/mwc/debrief/lite/tests/TestFileImport.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ */
+public class TestFileImport extends BaseTestCase
+{
+  
+  public void testImportRepFile()
+  {
+    System.out.println("started test");
+    File[] f = new File[1];
+    f[0]=new File("c:/Users/ayesha/git/debrief/org.mwc.cmap.combined.feature/root_installs/sample_data/boat1.rep");
```
IanMayo: @ayeshare - clearly, this is just a path on your PC.  You need to trim it to be a relative path.

org.mwc.debrief.lite/src/test/java/org/mwc/debrief/lite/tests/TestFileImport.java
```diff
@@ -0,0 +1,55 @@
...
+    });
+    try
+    {
+      Thread.sleep(200);
+    }
+    catch (InterruptedException e)
+    {
+      e.printStackTrace();
+    }
+    //do assertions here
```
IanMayo: Yes, we need to test that the data got loaded.

The use of `invokeLater()` may introduce a threading issue - since we can't run the `assert` test until it has completed.

So, maybe there's an `invokeAndWait()` that we can use instead.

org.mwc.debrief.lite/src/test/java/org/mwc/debrief/lite/tests/TestLiteLaunch.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class TestLiteLaunch extends BaseTestCase
+{
+
+  public void testTitleAndIcon()
+  {
+    JFrame ribbonFrame = DebriefLiteApp.getInstance().getApplicationFrame();
+    assertNotNull(ribbonFrame.getIconImage());
+    assertNotNull(ribbonFrame.getTitle());
+    assertTrue(ribbonFrame.getTitle().contains("Debrief Lite"));
```
IanMayo: I _think_ it would be better to refer to the original field (if present) that contains "Debrief Lite", rather than to use this _magic string_.

org.mwc.debrief.lite/src/test/java/org/mwc/debrief/lite/utils/TestUtils.java
```diff
@@ -15,10 +15,85 @@
...
+      Component parent, String klass, int index) {
+
+    // Debug line
+    //System.out.println("Class: " + parent.getClass() +
+    //    " Name: " + parent.getName());
+
+    if (parent.getClass().toString().endsWith(klass)) {
+      if (counter == index) { return parent; }
+      ++counter;
+    }
```
IanMayo: @ayeshare - I'm nervous of the use of this static field. It doesn't look thread safe.  Could you introduce some comments to explain what's happening in this method?

Hmm, a more robust strategy could be to pass `AtomicInteger` down through the call chain.  The value of the element can be updated.

IanMayo:
Thanks for the further progress in this @ayeshare . It's doing just the kind of testing I'd hoped it would.  Once we've finished this package of testing, I can see we'll have a good foundation for writing a wider set of system tests.

IanMayo:
@ayeshare - I'd like to work towards getting the PR into a tidy state so that we can merge it.

Is it integrated with the Maven build?

Issue #4436, opened 2019-08-06
Light weight tracks are not loaded in snail mode


helenayele:
When I try to play lightweight tracks in snail mode, the track's positions are not seen in the plot, but the time player controls are enabled and seems working. But if I try to edit the time properties like small step value and apply, every control in time menu will be disabled like this
[![Image from Gyazo](https://i.gyazo.com/8eae54766ee44016c56864e3e68ba4b9.png)](https://gyazo.com/8eae54766ee44016c56864e3e68ba4b9)

### Steps to reproduce:

1. Load BULK_RED_TRACKS.rep
2. Change to Snail Mode
3. Select Play 
4. See that it shows nothing in the map panel and looks like this

[![Image from Gyazo](https://i.gyazo.com/15918f4eaf00f880262b1b5c5c77f3e3.gif)](https://gyazo.com/15918f4eaf00f880262b1b5c5c77f3e3) 
5. Try to edit small steps value and apply
See that the time controls all become disabled like the first image.

saulhidalgoaular:
Fixed in https://github.com/debrief/debrief/pull/4437

helenayele:
Verified - able to see the tracks moving in snail mode as well :
[![Image from Gyazo](https://i.gyazo.com/adb06428d5b25f5abbd13dc4e8a6b29a.gif)](https://gyazo.com/adb06428d5b25f5abbd13dc4e8a6b29a)
But in Normal mode, when I turned off the visibility of the light weight track, the markers are not hidden, filed it as separate issue on #4459

PR #4437, opened 2019-08-06
4436 highlighters for lightweight tracks

IanMayo:
Supports #4436 

Not complete - just a start.


IanMayo:
This PR isn't complete.  

One reason why the highlighters were failing was that there wasn't a primary track. (see below)
image


That's fixed in my above commit.

But, when we use the normal highlighter, we only get a single rectangular highlight plotted. This still needs to be resolved.

saulhidalgoaular:
Good morning @IanMayo 

I think It is fixed.

PR #4454, opened 2019-08-10
4167 Angle Correction.

### LINKED ISSUE #4167: Some difference in XY Graphs between full debrief and D-Lite
Using the same files, when I try to generate xy graph in full debrief and D-Lite, I see some differences. For example the below graph is Bearing graph on sen_tracks.rep file.
[Image from Gyazo](https://gyazo.com/8525b16f02d66fc34805580c285b17e6)

On the left is on full debrief and on the right is on D-Lite.

saulhidalgoaular:
Fix #4167


IanMayo:
@saulhidalgoaular - is this related to how the XY Plots look different between Lite and Full?

Could you show before & after screenshots?

saulhidalgoaular:
> @saulhidalgoaular - is this related to how the XY Plots look different between Lite and Full?
> 
> Could you show before & after screenshots?

I think yes. This is how Debrief Lite looks now.
image

IanMayo:
Hmm, `ShowTimeVariablePlot3` is used by Full Debrief.  If that's broken, it would be broken in Full Debrief, wouldn't it?

saulhidalgoaular:
Full Debrief is correct because it is using the angle at line.
`XYPlotGeneratorButtons.java@118`

Issue #4353, opened 2019-07-10
New import message format


IanMayo:
An analyst (SE) wishes to load track data from another source.  

The source can output multiple tracks of data at a time in the OTH-Gold format.

View the data, produce a demo version of the data, to develop against, then trial it.

IanMayo:
Sample, of multiple track contact report (from U/C spec dated 2007):
````
MSGID/CTG81.0/GOLD/1215/APR/RIMPAC 00 CTC/T7123/UNEQUATED-UNKNOWN//SSGN/SUB//XX///HIT/14//3/10/10 /3442/F2K000162412/CAOB/03 POS/121410Z9/APR/2540N1/16204W3/VISUAL//10NM//340T/12K RADB/121410Z9/APR/0923B/FRONT PACE/12.34GHZ/000123.45/8100.446 /123.123/5.0SPC/CIRC
EOB//D18445994 SIGNA/XXX/123.45HZ/6,10,4/88/10.5/121410Z9/APR/VDSACT/STEADY SIGAM/UNIDENTIFIED SIGNAL IN MAIN MACHINERY SPACE MODEG/SUBMERGED/OTR/850FT/E/IN-LAYER DECIBELS CTC/T7164/UNEQUATED-UNKNOWN/TRDMARK/SSN/SUB//////00/3447/3 LOB/121420Z0/APR/2510N0/16210W0/120T/TACTAS/1.2DEG RADB/121420Z0/APR/0924Z/TRAP DEER/13.20GHZ/000200.55/4987.531/222.001 RMKS/EVALUATED AS TYPE II NUC
ENDAT
````

The 3rd party supplier clarified:
```
Our system only produces the sets listed below.
MSGID
CTC
POS
RMKS
ENDAT
```

So a streamlined version of the original example would be:
````
MSGID/CTG85.0/GOLD/1215/APR/DANPAC 00 
CTC/T7123/UNEQUATED-UNKNOWN//SSGN/SUB//XX///HIT/14//3/10/10/3442/F2K000162412/CAOB/03 
POS/121410Z9/APR/2540N1/16204W3/VISUAL//10NM//340T/12K 
POS/121415Z4/APR/2541N2/16206W5/VISUAL//11NM//20T/14K 
RMKS/EVALUATED AS TYPE WHISKY-NINE
CTC/T7164/UNEQUATED-UNKNOWN/TRDMARK/SSN/SUB//////00/3447/3 
[...]

PR #4466, opened 2019-08-15
4353 Import OTH data

IanMayo:
First solution for #4353


IanMayo:
@helenayele - there's not all that much to test here, just have a play with it really. The issue contains the path to a sample data-file.

PR #4475, opened 2019-08-23
4228 remove cursor in middle for drag element

### LINKED ISSUE #4228: Avoid showing green cursor at the center of shapes while using Drag Element
We are showing green hand cursor at the center of Rectangle while using drag element, and even if the cursor changes to green ,dragging it has no effect. See the below video:
[Image from Gyazo](https://gyazo.com/0d7012beafa673c70bdbf57282697e2f)

For circle shape it behaves differently i.e if I drag from center , the circle will move
[Image from Gyazo](https://gyazo.com/a42f12c603964b807af512047f27bc33)

saulhidalgoaular:
Fix #4228


IanMayo:
Hmm, @saul. An alternate strategy would be to implement drag on centre for rectangle shapes. I didn't realise it wasnt working. Hopefully the logic can be copied from the circle shape.

saulhidalgoaular:
> Hmm, @saul. An alternate strategy would be to implement drag on centre for rectangle shapes. I didn't realise it wasnt working. Hopefully the logic can be copied from the circle shape.

Good morning. Ok, I am implementing it.

PR #4478, opened 2019-08-27
Merge recent dev changes into Master

IanMayo:


IanMayo:
@helenayele - could you do some general testing of both full-Debrief and debrief-lite from this PR?

It sweeps up quite a lot of changes.

IanMayo:
@helenayele - which data-file was that using?  

I _Think_ it's associated with all those missing chart areas.

PR #4486, opened 2019-08-30
4474 NPE Fix for track loading

### LINKED ISSUE #4474: NPE while playing lightweight track after adding normal tracks to the map panel
While playing RED_BULK_TRACKS.rep file if I add normal tracks like boat2.rep file and start playing, D-Lite will be flooded with NPE. And the time slider will not be moving but it increments the time by the configured small step value.

### Steps to reproduce:

1. Launch D-Lite
2. Load  RED_BULK_TRACKS.rep file
3. Start playing, make the small step value to 60 min
4. While playing it, drag boat2.rep 
5. Start playing
6. Even if it seems playing, the time slider doesn't move, but the date seems incrementing as shown on the Date label.
https://screencast-o-matic.com/watch/cqj0jXtIO3
[NPE_while_playing_lightweightTrack_boat2.txt](https://github.com/debrief/debrief/files/3533544/NPE_while_playing_lightweightTrack_boat2.txt)

saulhidalgoaular:
Fix #4474

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/TotePainter.java
+++ b/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/TotePainter.java
@@ -512,7 +512,7 @@ public class TotePainter implements StepperListener, CanvasType.PaintListener,
 
 			final MWC.GenericData.Watchable[] list = primaryTrack
 					.getNearestTo(oldDTG);
-			if (list.length > 0)
+			if (list != null && list.length > 0)
 				oldPrimary = list[0];
 		}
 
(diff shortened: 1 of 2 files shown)


IanMayo:
@saulhidalgoaular - I'm nervous that there are multiple places that assume a zero-length array is returned.   Eclipse is showing 77 calls to `getNearestTo()`.  I think a more robust solution would be to find the scenario that is returning null, and switching it to returning an empty array.

_Hopefully_ in some future Java upgrade we'll be able to either use an annotation to say the return value must be non null, or use `optional` return values, or some tidy solution :-)

saulhidalgoaular:
> @saulhidalgoaular - I'm nervous that there are multiple places that assume a zero-length array is returned. Eclipse is showing 77 calls to `getNearestTo()`. I think a more robust solution would be to find the scenario that is returning null, and switching it to returning an empty array.
> 
> _Hopefully_ in some future Java upgrade we'll be able to either use an annotation to say the return value must be non null, or use `optional` return values, or some tidy solution :-)

Yes, you are correct. I have added 2 better solutions. I would suggest #4490 .

But I consider merging #4491 would be good to the project in general too. It would prevent any future NPE.

PR #4494, opened 2019-09-05
Default Date for Non time Related Items

### LINKED ISSUE #4439: Change default Date for newly created labels
Currently the default value for start and end Date of newly created label is set as 31/12/69 and time is set as 23:59:59
[Image from Gyazo](https://gyazo.com/18e077b9bef9f645969b257b766468e2)

And user complains about this and Ian suggests like this
 "it would be more logical to use today’s date (or the mid-point of the currently loaded data)"

### Steps to reproduce:

1. Launch D-Lite
2. Insert Label
3. Open label property and see the Start and End date

saulhidalgoaular:
Fix #4439


saulhidalgoaular:
Replaced by https://github.com/debrief/debrief/pull/4498 with a different approach.

PR #4498, opened 2019-09-08
4439 non set time debrief lite

### LINKED ISSUE #4439: Change default Date for newly created labels

saulhidalgoaular:
Fix #4439

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Properties/DatePropertyEditor.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Properties/DatePropertyEditor.java
@@ -141,6 +141,7 @@ abstract public class DatePropertyEditor extends
 
   static protected final String NULL_DATE = "dd/MM/yy";
   static protected final String NULL_TIME = "HH:mm:ss";
+  static protected final String NOT_SET_TEXT = "Not Set";
 
   /**
    * date formats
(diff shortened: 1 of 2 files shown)


saulhidalgoaular:
Good evening @helenayele .
I have added a minimum size for the Text Fields.
Regarding the number 2, I think it was done like that by design. I can change it if needed.

IanMayo:
> Regarding the number 2, I think it was done like that by design. I can change it if needed.

Yes, that seems like valid behaviour.

IanMayo:
Hello @saulhidalgoaular  & @helenayele - having "Unset" as a value does seem a tidier solution - it correctly indicates that there isn't a time assigned.

Hmm, I guess we could alternatively use `dd:mm:yy` and  `hh:mm:ss` as the "unset" markers, since these will guide/prompt the user in how to format the text.

How does that sound?

BTW - I don't see the "not set" markers in the current version of this branch, I just see the current date/time.

helenayele:
I can see the not set for null  date time values : [Image from Gyazo](https://gyazo.com/6c1db6cee1636d31399b2f5ab3f4850d)

I would also suggest to put the date format value instead of 'Not Set' for null values.

IanMayo:
Minor change to make box wider , so it shows full content of date/time.

PR #4500, opened 2019-09-08
4497 Invalid value handling

### LINKED ISSUE #4497: Show an error message to the user when an invalid time is inserted.
In Debrief Lite, when you edit a time property and you add an invalid value, It doesn't show any error message, looking like it was saved after pressing apply.
Steps to reproduce:
1. Open in Debrief Lite the properties editor of a time-related object.
2. Add an invalid value and press "Apply". (No error shown)
3. Open the properties editor again. (Same value it had before)

image

saulhidalgoaular:
Fix #4497

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
@@ -249,6 +249,7 @@ import MWC.GUI.Layer;
 import MWC.GUI.Layers;
 import MWC.GUI.ToolParent;
 import MWC.GUI.hasPropertyListeners;
+import MWC.GUI.Properties.Swing.SwingDatePropertyEditor;
 import MWC.GUI.Properties.Swing.SwingPropertiesPanel;
 import MWC.GUI.Tools.Action;
 import MWC.GUI.Undo.UndoBuffer;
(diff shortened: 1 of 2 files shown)


saulhidalgoaular:
Hi @helenayele . Let me know what you think about this message box. If you would like to give a different approach, just let me know.

We have reached 4500 tickets+pulls!!!!!!

helenayele:
1. Currently we show the same message for all error types :
For example when user provides valid date but in other day format, we tell the user that he/she has input invalid date 
[Image from Gyazo](https://gyazo.com/76657e0017ec5a6fe51221e558527c56)
So it would be very helpful if we tell the client that "User should enter the date in the correct format i.e dd/mm/yy 
2. When user enters invalid values , from the message better to remove "Property" like "End Time has invalid value.....,"
[Image from Gyazo](https://gyazo.com/37766d8983a18512a9da929e7fac01eb)

saulhidalgoaular:
Hi.
There are many kind of date formats. I have changed the message to the following, is that Ok?

image

Regarding the name of the property, I am using the original name in the error message ("Time end" in your case). It means it will also work for "Time start" or any Date Property. Changing it will imply changing the Property Names also. I can do it if needed.

org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
```diff
@@ -1156,7 +1157,7 @@ private void update(final PropertyDescriptor pd, final Object data)
     // current one for this object - or if there isn't a current value
     if ((newVal == null) || (!newVal.equals(oldVal)))
     {
-      if (newVal == null && pe instanceof SwingDatePropertyEditor)
+      if (newVal == null && newVal instanceof HiResDate)
```
saulhidalgoaular: @IanMayo . I _think_ you meant here `DatePropertyEditor` . I have pushed it in the next commit.

saulhidalgoaular:
It still needs changes. I will retake it after I check the Sensor Arc Issue. Please, don't review it yet.

org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
IanMayo: No, IMHO we can test the type of the object being edited, which is more direct than checkin the type of editor in place.

But, I may be wrong...

IanMayo: @saulhidalgoaular  was right all along :-)  we've already established that newVal is null, so we know we can't do type checking on it.

PR #4508, opened 2019-09-14
1410 sensorarc fill

### LINKED ISSUE #1410: SensorArc shading
FD said:
Also I have been playing with SensorArcs and note that I cannot turn the shading off, i.e. just have the line of the arc. The help suggests I can have the values of 0 (non-filled), 1 (transparent), 2 (semi- transparent) in the rep file but 0 doesn’t seem to do anything and from within the GUI there is no non-filled option.

saulhidalgoaular:
Fix #1410

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
@@ -1568,6 +1568,16 @@ public class SWTCanvasAdapter implements CanvasType, Serializable, Editable,
       _sg2d.draw(shape);
     }
   }
+  
+  @Override
+  public void nofillShape(Shape shape)
+  {
+    if (_sg2d != null && shape != null)
+    {
+      _theDest.setAlpha(255);
+      _sg2d.draw(shape);
+    }
+  }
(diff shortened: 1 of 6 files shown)


IanMayo:
Nice work @saulhidalgoaular 

You've missed one part. The sensor-arc `filled` status needs to be persisted to XML, and read in form XML, too.

I was about to do it, but we'll both benefit from you knowing how to do it.  If you look for instances of `getSemiTransparent` you'll find the class that handles XML marshalling of this object.

Note: a new XML attribute will have to be declared in that class. I suggest it's called `filled`.

Oh, and even more complex, the schema definition for `DynamicCoverage` will have to be modified.  The Debrief XML format is defined in a file called `debrief_plot.xsd`.

I've just had a look, and the definition is currently missing the attribute definition for `SemiTransparent` - so that will have to be added, too.

No problem if you get stuck with the above, I'm quite happy to help out :-)

IanMayo:
I've tested the storage, I did this:
* load boat1.rep, boat2.rep, boat_1_2_arcs.rep

I then opened `Nelson \ Dynamic Shapes \ fwd dynamic`, and changed the `Filled` and `Semi transparent` attributes for the first few shapes.  I saved the plot as `.dpf`, then re-opened.  The elements had the new (correct) values.

PR #4511, opened 2019-09-17
4450 undo bug track remove

### LINKED ISSUE #4450: Undo and Redo issues while removing track
After I delete tracks from the outline panel, I am not able to retrieve the deleted tracks using undo/redo functionality.

### Steps :
1. Launch D-Lite
2. Open BULK_BLUE_TRACKS.rep
3. Delete the first T23..tracks - see that undo button is enabled - this is fine
4. Delete HVU track ---see that the redo button now gets enabled - this button should not be enabled since we didn't select undo button
5. Delete the remaining T23A track
6. Now select Undo button , see that only the last removed  T23A track is retrieved by this action
https://screencast-o-matic.com/watch/cqj6eNOPlA

Note :
This issue is not reproducible with shapes though.

ayeshare:
Fixes #4450


IanMayo:
Tested, works fine.

PR #4514, opened 2019-09-19
reset undo buffer and clipboard on new or close plot

### LINKED ISSUE #4512: While resetting, D-Lite should forget all undo/redo action
When user selects New action or resetting, all undo and redo actions should be removed.

But currently when user resets, undo/redo actions are still enabled,(remembers the previous actions before resetting).

### Steps to reproduce :
1. Load BULK_BLUE_TRACKS.rep
2. Remove all tracks in this file 
4. Select New to reset D-Lite
4. See that undo action is still enabled. Here it should forget every action after resetting.
5. Select undo action, it retrieves the deleted tracks.

ayeshare:
Fixes #4512

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
@@ -86,6 +86,7 @@ public final class UndoBuffer extends Observable
 {
   static protected final int undo = 1;
   static protected final int redo = 2;
+  private static final int UNSET = -1;
   private final Vector<Action> theActions;
   private int presentAction;
 
@@ -250,7 +251,7 @@ public final class UndoBuffer extends Observable
   }
   public void resetBuffer() {
     theActions.clear();
-    presentAction=-1;
+    presentAction=UNSET;
     bufferChanged();
   }
   
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
@@ -248,4 +248,10 @@ public final class UndoBuffer extends Observable
 
     return res;
   }
+  public void resetBuffer() {
+    theActions.clear();
+    presentAction=-1;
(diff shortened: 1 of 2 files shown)


org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
```diff
@@ -248,4 +248,10 @@ public final String undoLabel()
 
     return res;
   }
+  public void resetBuffer() {
+    theActions.clear();
+    presentAction=-1;
```
IanMayo: @ayeshare - I'm nervous that "-1" is a magic number [1]

I've just looked at the code.  `presentAction` doesn't get initialised. I _think_ our code logic means that would not be an issue, but I'd be happier if it did get assigned something.

So, to solve both of the above I suspect we should introduce a private static member field title `UNSET`, which takes a value of -1.

We can then initialise `presentAction` using it, and we can also set the value using it `resetBuffer`



[1]. https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

ayeshare: this is done now.

org.mwc.cmap.legacy/src/MWC/GUI/Undo/UndoBuffer.java
```diff
@@ -86,6 +86,7 @@
 {
   static protected final int undo = 1;
   static protected final int redo = 2;
+  private static final int UNSET = -1;
   private final Vector<Action> theActions;
   private int presentAction;
```
IanMayo: @ayeshare - you should also initialise `presentAction` to `UNSET`

ayeshare: done

IanMayo:
@helenayele - could you re-test this, please?

helenayele:
Sure Ian ,Testing it..

IanMayo:
@helenayele - I'm happy to merge this, once you've tested it.

PR #4518, opened 2019-09-24
4517 class exception draganddrop

### LINKED ISSUE #4517: ClassCastException is thrown while changing the filled and semi-transparent properties of SensorArc
While changing the properties of SensorArc like filled and Semi-transparent properties, ClassCastException is thrown.
[ClassCastException_cannot cast to baseLayer.txt](https://github.com/debrief/debrief/files/3646613/ClassCastException_cannot.cast.to.baseLayer.txt)

### Steps to reproduce

1. Open boat1.dpf
[boat1.zip](https://github.com/debrief/debrief/files/3646618/boat1.zip)
2. From fwd dynamic(40 items) , Change the semi-transparent or filled properties to YES, if it was No, or change to No if it was Yes
3. Do the same for all items.
4. See that the above ClassCastException is thrown.

saulhidalgoaular:
Fix #4517

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/DragDropSupport.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/DragDropSupport.java
@@ -503,7 +503,7 @@ public class DragDropSupport implements DragSourceListener, DropTargetListener
             // note: we don't allow drag/drop reorganisation of
             // top level layers
           }
-          else
+          else if ( parent.getEditable() instanceof BaseLayer && ((EditableWrapper) event.item.getData()).getEditable() instanceof BaseLayer )
           {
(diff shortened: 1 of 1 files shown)


saulhidalgoaular:
Please, don't merge until I go a bit deeper in this ticket.

saulhidalgoaular:
Ok. I have found the reason of this. What was happening is that a sensor was getting dragged into another sensor.
This patch is valid for this case (and other more), because it checks if the destination of the drag is a BaseLayer.

I _think_ it is ready for testing/merging.

IanMayo:
Thanks for that @saulhidalgoaular .  Do you mean a `SensorArc` was being dragged onto another one?

saulhidalgoaular:
@IanMayo Yes. Exactly.

PR #4529, opened 2019-09-25
4526 collapseall button dlite

### LINKED ISSUE #4526: [3] No way to collapse all in outline view
I think here user wants to collapse all items in outline view. Currently we can't collapse /expand all the elements in one click. 
But here user can hide the whole outline panel by selecting collapse/expand button, better to inform user if possible.

saulhidalgoaular:
Fix #4526


helenayele:
1- Collapse all button doesn't collapse all expanded items ,rather it collapses the selected layer.
2- Collapse all button gets disabled if no layer is selected even if it is expanded.
[Image from Gyazo](https://gyazo.com/8ce582a9f7a9bd008ad8e88fb4cb5327)

org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/outline/OutlinePanelView.java
```diff
@@ -1149,7 +1149,7 @@ protected void initForm()
     final EnabledTest notNarrative = getNotNarrativeTest();
     final EnabledTest notIsLayer = getNotLayerTest();
```
IanMayo: @saulhidalgoaular  & @helenayele  - I don't know how this typo fell through the net 🤣

saulhidalgoaular: oohhh...

Issue #4532, opened 2019-09-25
Objects not rendered during Drag operation for Full Debrief


IanMayo:
When we moved to Eclipse 2019_06, shapes stopped being rendered as they were dragged.

Two instances of this problem were fixed last month (`RangeBearing.java` and `ZoomIn.java`) 

But, I've now identified 3 more occasions where it must be fixed:
![image](https://user-images.githubusercontent.com/1108513/65608201-c1785f00-dfa5-11e9-852f-e471da443b95.png)

You can see the intended behaviour in the `master` branch. Hold on, no you can't.  You need to switch to an older version of Eclipse (~2018?) to see the lines get rendered properly during drag.

These 3 buttons control mouse dragging.  Here's how to use each mode:
1. **Drag track segment** Open `midflow.dpf`.  Switch this toggle mode on (a new view may open, titled `Bearing residuals`).  Now use the mouse to drag the end one of the short red track segments.  While the segment is being dragged, the new position should be displayed.
2. **Drag component** . Open `sample.dpf`. Switch to this toggle mode, and try to drag the corner of a rectangle. As the rectangle is dragged, the new size should be displayed
3. **Drag whole feature**.  Still with `sample.dpf` open, switch to this toggle mode. Try to drag a rectangle - the whole shape should be dragged.

Note: the changes made earlier in the month actually spanned several commits:
https://github.com/debrief/debrief/commits/develop/org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/actions/ZoomIn.java

helenayele:
I can't reproduce the issue even, so @IanMayo  if you did get this issue before the fix, please verify it.

IanMayo:
Yes, I did get the issue, video here: https://youtu.be/p3jLgR0PiRQ

The issue is now fixed.

PR #4537, opened 2019-09-25
4532 switch new GC api to avoid repaint issues on drag events

theanuradha:
#4532

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/actions/DragFeature.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/actions/DragFeature.java
@@ -156,9 +156,6 @@ public class DragFeature extends CoreDragAction
     }
   }
 
-  final static org.eclipse.swt.graphics.Color fc =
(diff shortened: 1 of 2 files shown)


org.mwc.debrief.core/src/org/mwc/debrief/core/actions/DragFeature.java
```diff
@@ -263,21 +299,13 @@ public void doMouseDrag(final org.eclipse.swt.graphics.Point pt,
...
+         _myCanvas.getCanvas().redraw();
+         _myCanvas.getCanvas().update();
+       
 
-        // draw new track
-        drawHere(gc, offset);
-
-        // remember the last location
-        _lastLocation = newLocation;
-        // ok, let's ditch the GC
```
IanMayo: Hey, @G - I know we no longer call this code, but `fc.dispose()` looks like a mistake.  We only create it once, so we _should_ get `is disposed()` calls if it's called in the future.

Never mind, we don't dispose it at all now.

PR #4541, opened 2019-09-26
4540 Removing cached time String.

### LINKED ISSUE #4540: Time in Narratives (Inside Narrative Panel) are not getting updated after edition.
Time showed in the Narrative Panel is not getting updated after editing narratives.

image

Steps to reproduce:
1. Load narratives.
2. Edit one narrative, changing the time.
3. Apply and Save.

saulhidalgoaular:
Fix #4540

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/TacticalData/NarrativeEntry.java
+++ b/org.mwc.cmap.legacy/src/MWC/TacticalData/NarrativeEntry.java
@@ -287,7 +287,10 @@ public final class NarrativeEntry implements MWC.GUI.Plottable, Serializable,
 
   public final String getDTGString()
   {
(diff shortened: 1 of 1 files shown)


IanMayo:
@saulhidalgoaular  - I've changed the strategy for this fix.  IIRC, we introduced the cached value of the Date-String, to handle very, very long narratives.

But, I see that our implementation has two hash-codes, we were clearing the one value on a DTG change, but we were reading from the _other_ value.

So, I've removed the unnecessary hashed String value, and corrected the logic.

Does this solution seem ok?

@helenayele  - this will need re-testing, please.

saulhidalgoaular:
> So, I've removed the unnecessary hashed String value, and corrected the logic.
> 
> Does this solution seem ok?

Yes, It is perfect.

PR #4551, opened 2019-09-28
4533 Different Drag and Drop approach.

### LINKED ISSUE #4533: [6] Difficulty in opening Outline, Graph and Narratives panels by dragging
Client comment:
`the outline / graphs / narrative tabs sometimes don’t respond to attempts to drag them open`

Currently it is hard to open Narrative panel by dragging at all, you need to double-click.
[Image from Gyazo](https://gyazo.com/a08e6777cbb32fb6cf01acf9d2404900)

There is also difficultly in opening graph panel. 
But Outline panel is  expanded by default. And I don't see any difficulty in expanding it either to right or left.

saulhidalgoaular:
Fix #4533

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/DebriefLiteApp.java
@@ -998,9 +998,12 @@ public class DebriefLiteApp implements FileDropListener
       public void componentResized(final ComponentEvent e)
       {
         // TODO . This must be change once we update geotools.
-        // mapPane.setVisible(false);
-        // mapPane.setVisible(true);
-        mapPane.repaint();
+        // Reverted ec7262026be4cbe07c7c521687703ddfa1acfb97
+        // I _think_ it is causing the behavior described here.
+        // here https://github.com/debrief/debrief/issues/4051#issuecomment-511193288
+        mapPane.setVisible(false);
+        mapPane.setVisible(true);
+        //mapPane.repaint();
       }
     });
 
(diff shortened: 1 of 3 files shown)


IanMayo:
This pull request **introduces 2 alerts** and **fixes 2** when merging a79566f34ae5351cfc498ca259bda598b49c74ca into 4dfa26411615fc3dc533754559e2b464568246a3 - [view on LGTM.com](https://lgtm.com/projects/g/debrief/debrief/rev/pr-ac4c253d75751d44bcfff6de160ad8b67351ab34)

**new alerts:**

* 2 for Subtle call to inherited method

**fixed alerts:**

* 2 for Subtle call to inherited method

---

**Warning** - Automated code review for debrief/debrief will be disabled on October 1, 2019. You can avoid this by <a href="https://github.com/apps/lgtm-com/installations/new">installing the LGTM.com GitHub App</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

*Comment posted by [LGTM.com](https://lgtm.com)*

IanMayo:
@saulhidalgoaular - the two LGTM warnings seem valid.  I'm sure our code works, but it would be good to clarify which `setOpaque` we're referring to.     You _may_ have to use trial & error to determine which one we're actually referring to.

IanMayo:
@saulhidalgoaular  - there looks like one remaining issue:

[Click does not reveal](https://youtu.be/gbI5qq3iH1k "Click does not reveal")

When the app starts, we can double-click to reveal a panel.  It has some `calculated` width to use (presumably respective to the app size).  If we then drag a panel to resize it, we can double-click to go to/from that (remembered) size.    But, if we drag a panel to close it, we no longer have double-click to open, presumably because the `remembered` size is zero.   So, if the `remembered` size was zero, when we double-click, we should return to the `calculated` size.

Well, I think that's the solution :-)

helenayele:
After I drag the panel and close it, when I double click on the panel, it doesn't open it on my first double-click, but it opens on my second double-click.
https://screencast-o-matic.com/watch/cq6niwuloY

PR #4568, opened 2019-10-04
4566 merge sensors

### LINKED ISSUE #4566: Keep sensor color on Merge Track Segments
An analyst has 4 blocks of sensor data.

Each has a different sensor color.

When the 4 blocks are merged, the sensor cuts all take the color of the first one.

So, in the merge operation, we should check if any successive sensors have a different default color to the first.  If this happens, set the sensor color for cuts for successive sensors.

We should include unit tests that prove the above works for blocks of data with different colors, and with equal colors.

To test:
* load `sample_lots_of_sensors.dpf`
* switch on a couple of (nice & small) sensors in `OSPREY`
* change the ensure the sensors are of different colors
* observe that you can see the blocks of sensor lines in different colors on the plot
* right-click on one of the sensors and select `Merge into [this sensor]`
* ensure the sensor bearing lines don't change color

You could also try changing the color of a few sensor cuts within a sensor, then merging that sensor with another. The color of those cuts should still be their changed color.

IanMayo:
Fixes #4566

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GenerateTrackFromActiveCuts.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/GenerateTrackFromActiveCuts.java
@@ -493,7 +493,7 @@ public class GenerateTrackFromActiveCuts implements
           else
           {
             _logger.logError(IStatus.INFO,
-                "ot creating 'Generate track from active cuts': Missing range or bearing data (or is ambiguous) for sensor:"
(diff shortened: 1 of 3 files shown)


IanMayo:
@helenayele - could you give me a series of steps to reproduce the above error please?

helenayele:
The error "null org.mwc.cmap.core
Missing range data from sensor:sensor 3:98
" shown when I right-click on the sensors after selecting the two sensors :
[Image from Gyazo](https://gyazo.com/dfbb266cfd2d3cd7f987128ec2405261)

IanMayo:
@helenayele - they are just warnings.   

I've pushed a change that gives a longer message, and relaxes it from `Warning` to `Info`

PR #4569, opened 2019-10-04
4565 undo drag track segment

### LINKED ISSUE #4565: Provide Undo for drag track segment
While using Debrief, I thought I was in zoom mode, not drag-track-segment.

When I tried to zoom in on an area, it actually dragged a section of track.

Steps to reproduce:
* Open `midflow2.dpf`
* Switch to drag track segment line
* Drag part of a track segment
* Switch to zoom in mode
* Observer that it's not possible to undo drag-track.

So, it's switching mouse mode that's disabling the Undo.

* [ ] Check we don't start drag operation when clicking no-where near a track
* [ ] Ensure Undo is provided for drag-track operations

IanMayo:
Fixes #4565 

Note: now, we only put actions on the Undo buffer if they're undo-able.  If we put non undo-able operations on the buffer, the buffer gets blocked.


IanMayo:
Thanks for that @helenayele - could you create that as a separate issue, please?  I don't think it's related to this fix.

We'll need an accurate set of steps to reproduce, too.

helenayele:
Okay I have filed the above issue at #4572

PR #4579, opened 2019-10-16
Save As failing

### LINKED ISSUE #4578: Fails to re-open on `Save-as`
An analyst is performing a save-as operation.

Within this operation the file does get successfully saved. But the "new version" doesn't re-open, reporting Null Pointer Errors at `PlotEditor.java:2424` and `CorePlotEditor.java:661`

I'm unable to reproduce on a dev build on MACOS.  

Download 3_0_490 on a MS Windows device and investigate.   This _may_ be due to the Eclipse upgrade to E4.x

Reproducible on windows.

IanMayo:
Fixes #4578 

Store the page before we let the async close process delete the site() that it's hanging off.  Issue prob due to different E4 processing.


IanMayo:
Thanks @helenayele - could you see if the delay happens in versions before September 2019 (before we jumped to E4.x), please?

IanMayo:
If the old versions of Debrief have the same delay, then it's not because of the move to E4.x.

In that case, then I presume the delay is just the time taken to re-load the file. That's a processor-peformance dependent issue, that I don't think we can overcome. 

Does this sound right to you, @ayeshare ?

IanMayo:
Closing this, I want to include it in a new release.

PR #4581, opened 2019-10-18
4535 unable to change unit

### LINKED ISSUE #4535: [10] Unable to change the Rng/Brg measurement unit in D-Lite
> rng /bearing off the tool is too small

> i can’t figure out how to change it to miles`

Client said that he/she can't change the Rng Measurement units (yards in the screenshot below). The measured range/bearing is shown at the bottom-left in the image below

Client also said the font size was too small to be legible on small devices.

Screen Shot 2019-09-30 at 11 26 09

### Expected Result:

We need to implement option page that enable the user to configure all measurements units of his interest.

The displayed range and bearing should be in a larger font, to be visible on small screens.

saulhidalgoaular:
Fix #4535


IanMayo:
@Saul - the implementation works, but I think we need some _visual clue_ that a drop-down will appear.

Oh, and the user wouldn't need to select the units each time the mode is switched on.  They may choose "Yards" and want to keep those units all day.

IanMayo:
I'd like @saulhidalgoaular  to provide an implementation that doesn't require the units to be selected each time the button is selected. 

If the Range/Bearing button is on the far-right of the group, a pop-down menu button could be placed to the right of . The icon for this button would be some combination of the Range/Bearing button and a config/properties icon.

I'm going to request a new review from @helenayele once this is done, please.

IanMayo:
Aah, @saulhidalgoaular - here's an example of a drop-down menu that is separate to the button:
image

So, the user can either click on "Save" or the drop-down list.

Ooh dear, _maybe_ this isn't available for radio buttons.

saulhidalgoaular:
Good morning. I will restart this branch fresh..

saulhidalgoaular:
We can close this PR. It has been replaced by a custom JRibbonButton Implementation.

PR #4594, opened 2019-10-23
4531 start end date consistency

### LINKED ISSUE #4531: [13] start / end time for visibility of annotations is inconsistent
User comment

> Shape's start/end date is not applied correctly for the visibility of the shape.

### Steps to reproduce:
1. Load shapes.rep file
2. From Annotations layer , edit the Time Start to have a time value of `05:10` of Ellipse:test ellipse shape like this
[Image from Gyazo](https://gyazo.com/c0754a21f91a93175380995c990bff3e)
3. Start playing the track from time controller
4. See that when it reaches at 05:10, a marker on the ellipse shape is started to be shown and lasts for 3 seconds. See this video:
https://screencast-o-matic.com/watch/cqQTIJuVUI

saulhidalgoaular:
Fix #4531

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
@@ -1190,17 +1190,16 @@ public class Layers implements Serializable, Plottable, PlottablesType
             if (dtg != null)
             {
               res = extend(res, dtg);
-
-              // also see if it this data type an end time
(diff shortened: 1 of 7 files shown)


saulhidalgoaular:
> if start time = end time, again the shape will not be shown
It is like Ian mentioned, the bar has to step exactly over the time specified. But, yes, it has been implemented to show it.

PR #4604, opened 2019-10-30
4527 Add the interpolated point.

### LINKED ISSUE #4527: [5] Snail trail does not work well when interpolating over a sparse track
### Steps to reproduce:
* load `sparse_track.rep` (new file added [here](https://github.com/debrief/debrief/blob/develop/org.mwc.cmap.combined.feature/root_installs/sample_data/sparse_track.rep) )
* switch to snail mode
* notice how sometimes there's a gap between the marker and the trail

[Sparse interpolated track in snail mode](https://youtu.be/3yC2h9p171M "Sparse interpolated track in snail mode")

saulhidalgoaular:
Fix #4527 

Good morning @IanMayo 

This patch solves gap mentioned in the issue, but I am not sure if the rest of the Snail behavior is exactly what the application must do.

What I mean is that maybe the Snail must have a fixed size (all the time), which means drawing it from the interpolated points created, instead of the points from the original track.

I can change the application to that behavior if needed.

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/SnailDrawTrack2.java
+++ b/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/SnailDrawTrack2.java
@@ -140,6 +140,10 @@ final class SnailDrawTrack2
         final LightweightTrackWrapper track = (LightweightTrackWrapper) trk;
         dotPoints = track.getUnfilteredItems(new HiResDate(0, dtg.getMicros()
             - _trailLength), new HiResDate(0, dtg.getMicros() + 2000));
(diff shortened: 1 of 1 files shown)


IanMayo:
> What I mean is that maybe the Snail must have a fixed size (all the time), which means drawing it from the interpolated points created, instead of the points from the original track.

Yes, @saulhidalgoaular - I see what you mean.  If the analyst requests a 30-minute snail trail, and the first point is on an interpolated point, the line should go back 30 minutes, even if it ends on another interpolated point.

Yes please. @saulhidalgoaular - I'd like you to implement this, please.

helenayele:
Works fine as well in my side

PR #4613, opened 2019-11-01
4534 add label here

### LINKED ISSUE #4534: [7] Impossible to mark a position when a ship calls out a contact at range / bearing
Client said that he/she can't copy position by right click, (and there aren’t vectors). this makes it hard to, for example, mark the position when a ship calls out a contact at range / bearing.
mark a position say that when a ship calls out a contact at range / bearing

So we need to find a way to mark a position.

saulhidalgoaular:
Fix #4534

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GenericData/WorldLocation.java
+++ b/org.mwc.cmap.legacy/src/MWC/GenericData/WorldLocation.java
@@ -885,4 +885,11 @@ public class WorldLocation implements Serializable, Cloneable
 		lt.testPerpDistanceFrom();
 	}
 
+  public void setValues(WorldLocation worldLocation)
+  {
+    setLat(worldLocation.getLat());
+    setDepth(worldLocation.getDepth());
+    setLong(worldLocation.getLong());
+  }
+
 }
--- a/org.mwc.cmap.legacy/src/MWC/GenericData/WorldLocation.java
+++ b/org.mwc.cmap.legacy/src/MWC/GenericData/WorldLocation.java
@@ -10,7 +10,7 @@
  *
  *    This library is distributed in the hope that it will be useful,
  *    but WITHOUT ANY WARRANTY; without even the implied warranty of
- *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. 
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  */
 package MWC.GenericData;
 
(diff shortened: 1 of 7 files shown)


helenayele:
Okay now it looks like this :
When user tries to measure rng\brg values, rectangle is drawn in addition to the line being measure - this is okay
But if I right-click on specific place I get  `add label under cursor` ctx menu and label is added in Markers layer.
But it would be very helpful if the measured value is copied in the clipboard, and after user adds the label, he can paste these values on the label. Because currently if the user can't remember the measured values even if s\he puts label on that position.

https://screencast-o-matic.com/watch/cqXfoVUVmR

IanMayo:
@saulhidalgoaular - I also get the red rectangle that Helen gets on Measure Rng/Brg:
Screen Shot 2019-11-04 at 18 08 48

IanMayo:
@helenayele - I don't think we should do the `Copy` step that you suggested. This is because in D-Lite we don't have `Paste` in the Location Property editor.  This `Add label ...` is a kind of workaround to avoid users having to do that step (particularly when one of the main reasons users to `Paste Location` is to add a new marker)

saulhidalgoaular:
oh, well spotted. It has been fixed.

helenayele:
Okay now the rectangle is removed while measuring and it looks like:
[Image from Gyazo](https://gyazo.com/cf865c8a09e088f2fee3e59fd5546cfd)

Hopefully we will implement this copy-paste feature in the future, otherwise it will not fully support our users need.

IanMayo:
> Hopefully we will implement this copy-paste feature in the future, otherwise it will not fully support our users need.

Morning @helenayele :-) 

I'm keen that we both understand how the tool is to be used.

One reason (the main reason) for users copying a position and pasting it into a Location was to put a label marker on that exact spot.  But, now they can right-click and drop a label on that location.
The other reason for copying a position is to set the corner of a rectangle to a particular location.   That use case is met by dropping a marker on that position, switching to `Drag Feature` mode, and dragging the corner to that marker.

Is there another use case that still requires "Copy cursor location"?

helenayele:
Okay Ian thanks for the explanation. May be I miss understood the requirement. But my fear is that user puts the marker (label) on his/her point of interest, and went to other location and put other marker again, and finally what if he forgets his marked(labels) , how can he remember the measured values?

IanMayo:
Sure. The way to handle this situation is for the analyst to use the `Text Label` on the markers to record what that marker represents.  That's already something that analysts do.

PR #4618, opened 2019-11-03
4576 Errors after resizing

### LINKED ISSUE #4576: Bunch of errors while resizing D-Lite
While resizing D-Lite, after some time , a bunch of errors are logged continuously.
Errors:
[Image from Gyazo](https://gyazo.com/c046da3c3775bac178d30ce7f19dddb9)
Please see this video:
https://screencast-o-matic.com/watch/cq6j26urmm

saulhidalgoaular:
Fix #4576 
It fixes the cases explained.
In the case in the video, it is trying to repaint the map when it has an invalid size. I am checking it now.
Regarding the zooming in, I am preventing to a maximum zoom - in.

Aside of that, I see that when resizing a lot the application, it starts to zoom out a lot. However, I would suggest to apply this to prevent it.

https://github.com/geotools/geotools/pull/2324
Mentioned here
https://gis.stackexchange.com/questions/315396/zoom-geotools-jmappane-to-retain-data-area-on-resize

(just enabling the `fixedBoundsOnResize` after updating GeoTools)

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/Highlighters/PlotHighlighter.java
+++ b/org.mwc.debrief.legacy/src/Debrief/GUI/Tote/Painters/Highlighters/PlotHighlighter.java
@@ -191,42 +191,45 @@ public interface PlotHighlighter extends Editable
         // convert to screen coordinates
         final Point tl = proj.toScreen(wa.getTopLeft());
 
-        final int tlx = tl.x;
-        final int tly = tl.y;
-
-        final Point br = proj.toScreen(wa.getBottomRight());
-        // get the width
-        final int x = tlx - _mySize;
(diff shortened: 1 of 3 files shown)


IanMayo:
I _guess_ we have to intercept the resize operation, and ensure the bounds are always _just a little bit less than +/- 90 degs.

IanMayo:
@saulhidalgoaular - I think the issue is related to us adding a "buffer" around the viewport when we resize the plot.

If we repeatedly click on zoom-out, D-Lite won't let us zoom out past a particular point.  

But, when we resize D-Lite (or the collapsible panels) we keep zooming out further and further.

Could you see if we can intercept this resizing?   _Maybe_ we can restrict how far out the map can zoom, similar to how we constrain the "Zoom Out" button.

PR #4635, opened 2019-11-08
fix project build path in demonstrator

theanuradha:


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/menu/DebriefRibbonTimeController.java
```diff
@@ -1114,28 +1118,35 @@ public static void setButtonsEnabled(List<Command> commands,
...
+        // switch the icon
+        final Image playStopinImage = MenuUtils.createImage(image);
+        final ImageWrapperResizableIcon imageIcon = ImageWrapperResizableIcon
+            .getIcon(playStopinImage, MenuUtils.ICON_SIZE_16);
+  
+        playCommandButton.setExtraText(tooltip);
+       // playCommand.setIconFactory(ResizableIconFactory.factory(imageIcon));
+        
+        playCommandButton.setIcon(imageIcon);
+        if(playButton!=playCommandButton)
```
IanMayo: I don't quite understand why we're comparing buttons here, but it does look like a symptom of all the static public objects - where we can't keep track of what we're looking at.  IMHO...

theanuradha: `playButton` I keep ref soon as user hit play, so i found some case button u have in variable and button passed to method not same so in case it is not same I set icon as one I collect at event is the one in UI.

PR #4638, opened 2019-11-08
4546 proper color for graph

### LINKED ISSUE #4546: [4] Graph label markers not always take the right track color
Problem:  while the sparse track drawn at the top has a range of colors, the graph is all in the default color for the track.  The line should change color to match the position it's plotting.

image

To reproduce:
* load `sparse_track.rep`
* view graph of course

saulhidalgoaular:
Fix #4546

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/PlainWrapper.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/PlainWrapper.java
@@ -176,25 +176,25 @@ abstract public class PlainWrapper implements Plottable, Serializable,
 
   /**
    * property name to indicate that the text in the label has changed
-   * 
+   *
    */
   public static final String TEXT_CHANGED = "TEXT_CHANGE";
 
   /**
    * the name of the property change event to fire should this object get moved
-   * 
+   *
    */
   public static final String LOCATION_CHANGED = "LOCATION_CHANGED";
 
   /**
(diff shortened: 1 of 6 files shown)


IanMayo:
@saulhidalgoaular - I'm looking at this fix.  I've got a breakpoint in the `is this interpolated` logic, but it isn't getting called.  Are there steps to make the code use that logic?

Hmm, _I think I'm going mad_.  On the `develop` branch I no longer have this issue.

@helenayele - on Monday could you check if you can reproduce this in `develop`.

saulhidalgoaular:
Good evening @IanMayo 
It is still reproducible in `develop`. I am attaching here the steps.

bug

helenayele:
@IanMayo  yes, it is reproducible in develop like this :
[Image from Gyazo](https://gyazo.com/036b378280263e7a606a9dae177f31cf)

IanMayo:
@helenayele - I've switched the test mechanism to use different logic. I'd appreciate you re-testing it, please.

helenayele:
Sure , now it keeps its color in the color even if interpolate points is set to true.
[Image from Gyazo](https://gyazo.com/eb2d47428360c807fadc9e4ac41d7d49)

IanMayo:
@helenayele - I'm not sure how you do it, but could you mark the PR as "approved" please?

PR #4643, opened 2019-11-11
Reduce Math Transformation instance

### LINKED ISSUE #4485: Reduce instances of Geo Transform in D-Lite
We declare the projection transform in both LiteMapPane and in GeoToolMapProjection

We should just use the one in GeoToolMapProjection - and pass it into LiteMapPane.

But, it _looks_ like the reverse transform is declared in LiteMapPane:

LiteMapPane:
````
      theTransform = CRS.findMathTransform(worldCoords, worldDegs);
````

GeoToolMapProjection:
````
      data_transform_val = CRS.findMathTransform(worldDegs, worldCoords);
````

We should certainly switch to just using the latter one.  But, along the way we should verify if/where we're reversing the transform used in the first - to check we're not doing the wrong transform.

saulhidalgoaular:
Fix #4485


IanMayo:
I'm pleased you'd taken this on @saulhidalgoaular  - I'll try to work through it tomorrow.

PR #4646, opened 2019-11-12
4645 Time controller ribbon enabling/disabling.

### LINKED ISSUE #4644: Display mode and highlighter buttons are not disabled on empty plot
All the new Display mode buttons and highlighter buttons are kept enabled even if there is no any plot opened in the map panel.
[Image from Gyazo](https://gyazo.com/6245cb31adae6bb48aae146456933bd5)
 
But when I try to change some values in the property window doesn't throw Null pointer exception.

### LINKED ISSUE #4645: Time Controller is not enabled when a narrative replay is added
After adding `narrative.rep`,  Time Controller slider should be enabled to see how it progressed during time.

saulhidalgoaular:
Fix #4645 
Fix #4644


saulhidalgoaular:
Good morning.

I am adding in this patch another `state`, which I called `TIME_ENABLED_STATE`. It is used for cases like, having only narratives or labels with start+end time (without tracks). It will enable the time bar, but not the `display mode` and `highlighter mode`, because there is no track.

I am also trying to approach it with a dependency injection vision, leaving the enable/disable to listeners triggered by the state change.

PR #4647, opened 2019-11-12
4545 control+del is also added as required

### LINKED ISSUE #4545: Ctrl + V  and Ctrl + X shortcuts are not working
I have been using ctrl+v to paste the copied item from clipboard. But now it doesn't work. The same is true for ctrl +x to cut an item.

But  ctrl +C  is working for  copy purposes.

Did we have shortcut for delete action ? ctrl +d ?

saulhidalgoaular:
Fix #4545 
Control + DELETE


IanMayo:
@helenayele - could you check this? 

It may also fix ctrl-delete for positions, as in #4651

IanMayo:
Yes, we can delete track points in this branch.

PR #4653, opened 2019-11-13
4651 delete track points

### LINKED ISSUE #4651: Unable to remove a position using short-cut keys or delete button
When I try to remove a position from outline view using Delete button or shortcut keys like ctrl + delete key or delete key only, the position will not get removed from outline view. Instead it throws this 
"INFO:  Exception caught:Failed trying to remove 0501 from 120505.00 (4 positions)
INFO:  Exception caught:Failed trying to remove 0501 from 120505.00 (4 positions)" 

### Steps :

1. Open shapes.rep
2. From Track:Nel_Style 2 layer ,remove 0501 either using delete +ctrl key or Delete button from the outline toolbar.
3. See that it is will not be removed.

IanMayo:
Fixes #4651


IanMayo:
Ok, merging.  We'll leave the ctrl-del feature to #4545

PR #4659, opened 2019-11-15
Include downloaded JRE in windows distribution

IanMayo:
I've come up with a strategy where for Travis builds we include a JRE with the MS-Windows dist.

It _seems_ to only take 10 seconds to download the build, and it saves having to include the JRE in our GitHub repo.


theanuradha:
@IanMayo I was working a solution where you can update Jre as well But this seems good and not sure you want/need to update JRE via Eclipse update process.

Update process seems help when we move along but I let you decide

PR #4670, opened 2019-11-19
Don't call the fireModified, because it is all checked at the end

### LINKED ISSUE #4134: Refresh issues with large Narrative files

saulhidalgoaular:
Fix #4134

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/Utilities/ReaderWriter/PlainImporterBase.java
+++ b/org.mwc.cmap.legacy/src/MWC/Utilities/ReaderWriter/PlainImporterBase.java
@@ -208,7 +208,7 @@ public abstract class PlainImporterBase implements PlainImporter
     {
       _newLayers.add(theLayer);
     }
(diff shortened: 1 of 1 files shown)


helenayele:
To load large narrative file like narrative_bulk.rep it takes  2 minutes and 20 seconds . And also we need to show some progress bar/or dialog to the user while it loads the file. 

I suggest to improve the speed again and also to show progress message to the user.

See the below recording while loading bulk narrative file.
https://screencast-o-matic.com/watch/cqXYjBUtMy

helenayele:
But the refresh speed is now greatly improved, I just load the Narrative_bulk.rep file and try to edit some narratives , and it takes 2-3 seconds to update /refresh.

IanMayo:
@helenayele - let's not worry about the slow import speed for now. That really is an extreme file, that we don't expect to encounter in D-Lite.

PR #4681, opened 2019-11-20
4661 split legs

### LINKED ISSUE #4661: Slice track into segments on temporal gaps
Sometimes tracks have long breaks.  By default Debrief joins the breaks with a straight line.

But, that can suggest the platform travelled somewhere that it didn't.

The solution is to break the track into segments, which is done by right-clicking on the first/last point in the leg and selecting "Split".

It would be better if Debrief could find the gap, and slice the track accordingly.  There are two ways in which this could be supported:
* A new track-level operation called "Slice into legs".   This would trigger a drop-down menu that provides a series of time-interval thresholds:
    * Gaps over 1 minute
    * Gaps over 1 hour
    * Gaps over 1 day
    * Gaps over 1 week
* A formatting helper that can go into a REP file, to trigger the above.

### Split tracks command

You should be able to load the `sparse_tracks.rep` data-file. It contains several periods of where data is missing for over an hour. So, once loaded, you should be able to right-click on it and select `Split track into segments on gaps over...` > `1 Hour`.  It should split into 3 legs.

Alternatively, open the file and try to split over 1 day, and nothing will happen.

Or split it at 1-minute, and it will slice into lots of legs.

### Split tracks formatting helper

Analysts are able to embed some formatting commands into `.rep` files. As Debrief processes these, it creates a "formatter", which is then applied to subsequent loaded data.  Read more here:
https://debrief.github.io/tutorial/reference.html#replay_format_annotations

An attached zip-file contains some sample data that includes formatting helpers (view the files in text mode to see).  The files are based upon the `sparse_tracks.rep` data-file.  They include different versions of the Split Track format helper.
[SparseTracks.zip](https://github.com/debrief/debrief/files/3865218/SparseTracks.zip)

IanMayo:
Fixes #4661 

Allow undo on split into segments

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/SplitTracksIntoLegs.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/ContextOperations/SplitTracksIntoLegs.java
@@ -15,6 +15,7 @@
 package org.mwc.debrief.core.ContextOperations;
 
 import java.util.ArrayList;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


IanMayo:
Great, thanks Helen.

IanMayo:
Ok @helenayele - ready for re-test.  The warnings shouldn't appear now.

IanMayo:
Yes, the implementation doesn't support that. It would be complex processing to store the state of those segments sufficient to allow multiple levels of undo.   I'm happy with the implementation so far.

PR #4702, opened 2019-12-09
4701 Geotiff Rendering

### LINKED ISSUE #4701: GeoTiffs loaded but not rendered
It looks like the move to J11 has broken the GeoTiff rendering.

When we drag a GeoTiff onto the plot, the image should be shown.

Aah, or maybe it's the GeoTools upgrade, since that's what is rendering them.

If you don't have a GeoTiff - I can provide one.

saulhidalgoaular:
Fix #4701

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/loaders/TifLoader.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/loaders/TifLoader.java
@@ -16,6 +16,10 @@ package org.mwc.debrief.core.loaders;
 
 import java.io.File;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


IanMayo:
Hello @saulhidalgoaular - I'm interested in the background to this fix. Can you explain what you discovered, and how this fixes it please?

saulhidalgoaular:
Good afternoon @IanMayo.

Sure. What I saw is that the correct service provider was not getting loaded, despite it is in the META-INF file of the library as expected. Missing service was `URLImageInputStreamSpi()`, which is able to load an image from an URL.

My proposal is to check if it is loaded, and if not then I load it manually.

PR #4709, opened 2019-12-17
4648 Update treenode outline

### LINKED ISSUE #4648: When I use ctrl + x to cut an item, as can be seen in the below video, items below the selected layer will become invisible.
And also when I use ctrl + x to cut an item, as can be seen in the below video, items below the selected layer will become invisible.
https://screencast-o-matic.com/watch/cqXXiZUlCR

### Steps:
1. Launch D-Lite
2. Load shape.rep file
3. From annotation layer, select the first label and cut it using ctrl +x shortcut keys
4. See that all elements below it will be invisible

_Originally posted by @helenayele in https://github.com/debrief/debrief/issues/4545#issuecomment-552325946_

saulhidalgoaular:
Fix #4648


saulhidalgoaular:
This fix is very related with this: https://github.com/debrief/debrief/pull/4502
I have tested it and both issues look fixed. Please, check the mentioned ticket when testing.

PR #4711, opened 2019-12-19
4397 switch to correct mode dpf

### LINKED ISSUE #4397: Switch D-Lite tools to correct mode on load DPF
DPF files store the snail display mode.

But, when we load them into D-Lite the toolbar doesn't update, so it may mistakenly show the wrong mode.

Update the ribbon toolbar to match the loaded data.

saulhidalgoaular:
Fix #4397


saulhidalgoaular:
Good evening @IanMayo 
I am merging request to `4592` because the flamingo update is almost ready.
Please, test it before merging @helenayele

saulhidalgoaular:
Hi @helenayele 
Well spotted. I am including it too. It is fixed now.

PR #4732, opened 2020-01-19
4730 Layers performance Fix.

### LINKED ISSUE #4730: Handling large track files
I'm working with some large track files.

Two large tracks are in the attached zip file.  The first one takes a few seconds to load, but the second one takes ages.  I _Think_ it's because it's doing a screen redraw after loading each new point.

I can think of two way around this:
* cache up all the positions for a track in the ImportReplay class, and add them at once at the end.
* load the points into a new track, but only add the track to the `Layers` entity once we've finished loading the file
* it _may_ be an option to introduce the ability to pause the `Layers` object from sending out modified/extended/reformatted events.  Then, un-pause it, and send out an extended event manually.

[Archive.zip](https://github.com/debrief/debrief/files/4078077/Archive.zip)

saulhidalgoaular:
Fix #4730

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
@@ -597,6 +597,9 @@ public class Layers implements Serializable, Plottable, PlottablesType
    */
   private final PropertyChangeListener _extendedListener;
 
+  private final HashMap<String, Layer> _name2layer =
+      new HashMap<String, Layer>();
+
   /**
    * a list of layers
    */
@@ -700,7 +703,7 @@ public class Layers implements Serializable, Plottable, PlottablesType
       {
         current.append(thisL);
       }
-      
+
       _name2layer.put(thisL.getName(), thisL);
     }
 
@@ -740,7 +743,7 @@ public class Layers implements Serializable, Plottable, PlottablesType
     }
 
     addThisLayerDoNotResize(layer);
-    
+
     // and fire the extended event
     fireExtended(null, layer);
   }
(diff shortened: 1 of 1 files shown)


IanMayo (commented):
@saulhidalgoaular  - could you give me an overview of how the fix works please?

IanMayo:
@saulhidalgoaular  - I _think_ the breaking tests _may_ because of the change in how we're managing layers.  *But* I have been working on the ASW document importer recently.  Aah, let me run the tests locally, to see if its a timezone issue.

saulhidalgoaular:
> @saulhidalgoaular - could you give me an overview of how the fix works please?

Good afternoon. Yes. Slow performance happened when you add both .rep in the same plot.
 What was happening is that to add each element, you were checking all the previous element. It is O(N^2), which is a lot for a lot for 50.000 nodes.
I simply cached the elements in a Hash, reducing the complexity to approximately O(N). So, next time an element is added, we do a pseudo-constant check for the previous element.

IanMayo:
Hmm, but there are only a couple of layers loaded.  

Aah, or do we try to descend down through track looking for more layers?  Which then means looping through all the fixes?

saulhidalgoaular:
> Hmm, but there are only a couple of layers loaded.
> 
> Aah, or do we try to descend down through track looking for more layers? Which then means looping through all the fixes?

Correct.

IanMayo:
> Aah, or do we try to descend down through track looking for more layers? Which then means 
> looping through all the fixes?

> Correct

Aah, in that case I would rather add a 'marker' interface to TrackWrapper that will prevent Layers from descending down through its children.  Maybe 'ICantContainLayers'.

That will avoid the hashing requirement.  Right?

saulhidalgoaular:
> > Aah, or do we try to descend down through track looking for more layers? Which then means
> > looping through all the fixes?
> 
> > Correct
> 
> Aah, in that case I would rather add a 'marker' interface to TrackWrapper that will prevent Layers from descending down through its children. Maybe 'ICantContainLayers'.
> 
> That will avoid the hashing requirement. Right?

Correct. If you prefer that alternative. I can implement it.

IanMayo:
Dropping this in favour of alternate solution in #4733

PR #4733, opened 2020-01-19
4730 Interface Tagging . Don't check the Elements that cannot contain layers inside

### LINKED ISSUE #4730: Handling large track files

saulhidalgoaular:
Fix #4730

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/ICantContainLayers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/ICantContainLayers.java
@@ -1,0 +1,6 @@
+package MWC.GUI;
+
+public interface ICantContainLayers
+{
+
+}
--- a/org.mwc.cmap.legacy/src/MWC/GUI/ICantContainLayers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/ICantContainLayers.java
@@ -1,5 +1,24 @@
+/*
+ *    Debrief - the Open Source Maritime Analysis Application
+ *    http://debrief.info
+ *
+ *    (C) 2000-2020, Deep Blue C Technology Ltd
+ *
+ *    This library is free software; you can redistribute it and/or
+ *    modify it under the terms of the Eclipse Public License v1.0
+ *    (http://www.eclipse.org/legal/epl-v10.html)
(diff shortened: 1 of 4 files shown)


org.mwc.cmap.legacy/src/MWC/GUI/ICantContainLayers.java
```diff
@@ -0,0 +1,6 @@
+package MWC.GUI;
+
```
IanMayo: @saulhidalgoaular - could you add a comment explaining the role of this interface?

saulhidalgoaular: Sure. 1 Minute

org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
```diff
@@ -488,7 +488,12 @@ private static Layer checkLayer(final Layer layer, final String name)
         {
           return l;
         }
-        else if ( !(l instanceof ICantContainLayers) )
```
IanMayo: @saulhidalgoaular  - I've slightly restructured, so I can document why we have that path through the logic

It won't be a problem, will it?

IanMayo:
Nice work Saul, the tracks now load _in a flash_.

org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
saulhidalgoaular: Your change is perfect!!!!

PR #4734, opened 2020-01-20
limits the height and width to not expand beyond the frame.

### LINKED ISSUE #4726: Don't allow graph panels to grow out of the screen size
It shouldn't be possible to make the graph panel larger than the visible window. if we allow this D-Lite will goes to out of screen area and not possible to get its upper toolbar menus.
[Image from Gyazo](https://gyazo.com/8c0ae86bf5868d8486d5899525056003)

Video:
 https://screencast-o-matic.com/watch/cYVjIhvwW6

ayeshare:
Fixes #4726

### DIFF
--- a/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JXCollapsiblePaneWithTitle.java
+++ b/org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JXCollapsiblePaneWithTitle.java
@@ -189,12 +189,14 @@ public class JXCollapsiblePaneWithTitle extends JXCollapsiblePane
 
             bounds = getBounds();
             final int maxheight = DebriefLiteApp.getInstance().getApplicationFrame().getHeight();
-            //165 is the height of ribbon
-            if(bounds.height<maxheight-165) {
+            //minor adjustments +5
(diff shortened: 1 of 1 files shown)


org.mwc.debrief.lite/src/main/java/org/mwc/debrief/lite/gui/custom/JXCollapsiblePaneWithTitle.java
```diff
@@ -184,7 +188,15 @@ public void mouseDragged(final MouseEvent event)
             }
 
             bounds = getBounds();
-            bounds.height = (bounds.height - oldHeight) + newDimension;
+            final int maxheight = DebriefLiteApp.getInstance().getApplicationFrame().getHeight();
+            //165 is the height of ribbon
```
IanMayo: @ayeshare - can't we determine the actual current height of the ribbon?

That would be a longer term fix

ayeshare: Replaced the fixed height with height of ribbon.

ayeshare:
> @ayeshare - just the change to avoid those _magic numbers_ for the component sizes, please.

done.

PR #4736, opened 2020-01-20
4729 tma speed performance

### LINKED ISSUE #4729: Optimise UI performance when dragging TMA tracks
An analyst is working on a really large dataset:
* 8 day ownship track at 15 minute positions (~66k items)
* 3 hours of sensor data, with 300 sensor cuts
* TMA solution for whole period of sensor data, with 18 solutions, plus infills

When he drags the TMA solution, each redraw of the Stacked Dots takes around 0.1 seconds (very roughly).  This isn't interactive enough to keep up with his mouse movements.

We _suspect_ the time lag is because of the volume of processing being conducted, maybe in these areas:
* time taken to get ownship position at time of sensor cut
* number of values recalculated
* number of items re-plotted in JFreeChart on the stacked dots

Once Ian has produced a sample dataset for this issue, we can start considering/identifying bottlenecks.

One suspicion is that we're re-calculating/re-rendering more data than is necessary.  When a single leg of data is being dragged, we only have to re-do the points for that leg - the values for the other legs are unchanged.  

In the scenario in question we have 18 legs (call it 20).  If we only re-calculated the data for the current leg, that would only take 5% (1/20) of the time that it's currently taking.

saulhidalgoaular:
Fix #4729

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/Wrappers/Track/RelativeTMASegment.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Wrappers/Track/RelativeTMASegment.java
@@ -990,11 +990,6 @@ public class RelativeTMASegment extends CoreTMASegment implements
 
     }
 
-    final Double courseVal = new Double(MWC.Algorithms.Conversions.Degs2Rads(
-        this._courseDegs));
-    final Double speedVal = null;
-    updateCourseSpeed(courseVal, speedVal);
-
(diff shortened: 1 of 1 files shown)


IanMayo:
Ian thinks we should drop this, since it looks like the course and speed values for the points in the leg don't get updated.  An alternative (`setSpeedQuiet()`) strategy has been devised.

PR #4741, opened 2020-01-20
4735 log reducing

### LINKED ISSUE #4735: Reduce excessive error logging in JUnit Tests
The Travis build log is _polluted_ with debug logging lines. They _almost certainly_ related to some maths work Ian was working on related to detecting when a ship turns.  

They were inserted so that Ian could look at the performance on 'real' user data, using a deployed build of Debrief.  But, that venture was suspended.

So, they really should be commented out.  It's a case of doing `search-files` for some identifier in the error listing.

saulhidalgoaular:
Fix #4735

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Editable.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Editable.java
@@ -397,8 +397,8 @@ public interface Editable
           final int len = pd.length;
           if (len == 0)
           {
-            System.out.println("zero property editors found for " + toBeTested
-                + ", " + toBeTested.getClass());
+            //System.out.println("zero property editors found for " + toBeTested
+            //    + ", " + toBeTested.getClass());
             return;
           }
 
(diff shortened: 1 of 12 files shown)


org.mwc.cmap.legacy/src/MWC/GUI/Editable.java
```diff
@@ -397,8 +397,8 @@ public static void testTheseParameters(final Editable toBeTested)
           final int len = pd.length;
           if (len == 0)
           {
-            System.out.println("zero property editors found for " + toBeTested
```
IanMayo: @saulhidalgoaular  the error logging in this file is needed - it helps us track badly configured editors (such as when we don't provide the getter/setter for a named property)

So, we should keep the `system.out` entries in this file.

org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/ambiguity/AmbiguityResolver.java
```diff
@@ -1462,7 +1462,7 @@ else if (valueDelta > 180)
                 + " lastBrg:" + (int) lastCut.getBearing() + " brg:"
                 + ((int) cut.getBearing()) + " brg delta:" + brgRate
                 + " gap rate:" + gapRate + " OS zig trip:" + TRIP_ZIG;
-            doLog(logger, stats, time);
```
IanMayo: @saulhidalgoaular - in this file, a _slightly_ less intrusive change would be to just comment out the contents of `doLog()`.

IanMayo:
Thanks @saulhidalgoaular  - all good, just two minor changes, please.

saulhidalgoaular:
> Thanks @saulhidalgoaular - all good, just two minor changes, please.

Changes Applied :+1:

PR #4742, opened 2020-01-21
Implemented dynamic labels for undo/redo and collapse

### LINKED ISSUE #4640: D-Lite Ribbon Tooltips
Once we've upgrade the new Radiance Flamingo ribbon, we should provide tooltips for the Ribbon Bar buttons.

Display Mode & Highlighter properties buttons in particular could do with tooltips.

But, let's defer this until we've integrated @ayeshare 's code.

ayeshare:
Fixes #4640


helenayele:
Can we do the same for highlighter as well. We are showing the same tool-tip for both square and symbol properties :[Image from Gyazo](https://gyazo.com/1a158e4192dadad148e834461cf3b26d)

PR #4751, opened 2020-01-23
4749 rendering performance large files

### LINKED ISSUE #4749: Rendering issue while opening large files
When I try to open files from https://github.com/debrief/debrief/files/4082161/Archive.zip , it takes around 4-5 minutes to open the second file. But if the tracks in the files were saved as invisible, the loading does't take more than a second or a couple of seconds.
So we need to improve the performance while rendering those large files.

### LINKED ISSUE #4752: Outdated Bounds for LightweightTracker
Steps to reproduce:
1. Open `lightweight_test_tracks.rep`
2. Edit the properties of the first element of the Track "FisherOne_002" (like in the picture)
image
3. Change the Degrees of the location from 0 to 10
4. Apply all the changes.
5. Fit to Windows (It should show all the elements)

saulhidalgoaular:
Fix #4749 
Fix #4752

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
@@ -1288,7 +1288,7 @@ public class TrackWrapper extends LightweightTrackWrapper implements
     {
       setRelativePending();
     }
-    
+
     flushBounds();
   }
 
@@ -2005,7 +2005,6 @@ public class TrackWrapper extends LightweightTrackWrapper implements
   @Override
   public final WorldArea getBounds()
   {
-    _bounds = null;
     if (_bounds == null)
     {
       // we no longer just return the bounds of the track, because a portion
--- a/org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
+++ b/org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
@@ -2008,7 +2008,7 @@ public class TrackWrapper extends LightweightTrackWrapper implements
     if (_bounds == null)
     {
       WorldArea res = null;
-      
+
       // we no longer just return the bounds of the track, because a portion
       // of the track may have been made invisible.
 
@@ -2127,9 +2127,10 @@ public class TrackWrapper extends LightweightTrackWrapper implements
           res = this.getSegments().first().getBounds();
         }
       }
-      
+
       return res;
-    }else
+    }
+    else
     {
       return _bounds;
     }
@@ -2867,7 +2868,7 @@ public class TrackWrapper extends LightweightTrackWrapper implements
     try
     {
       calculateBoundCache();
-      
+
       // check we are visible and have some track data, else we won't work
       if (!getVisible() || this.getStartDTG() == null)
       {
@@ -2962,7 +2963,8 @@ public class TrackWrapper extends LightweightTrackWrapper implements
 
       // lastly - paint any TMA or planning segment labels
       paintVectorLabels(dest);
-    }finally
+    }
+    finally
     {
       setCachedBounds(null);
     }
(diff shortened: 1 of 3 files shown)


IanMayo:
@saulhidalgoaular -  I'm interested. Could you give some details about what bottleneck(s) you found, and how you fixed them?

IanMayo:
@saulhidalgoaular  - it's an interesting test failure here: https://travis-ci.org/debrief/debrief/builds/641002881#L5102

I wouldn't want to fix the test by changing the expected result - maybe it's highlighting a problem in the new strategy.

saulhidalgoaular:
> @saulhidalgoaular - I'm interested. Could you give some details about what bottleneck(s) you found, and how you fixed them?

Good afternoon. Idea is similar to the other bottlenecks we are facing. In this case, the algorithm to draw the map was O(N^3). What I did was removing the most inner loop to reducing it to O(N^2). The idea was that the bounds was pre-calculated, without needing to iterate all the elements (fixes) again every time. I am invalidating the cache every time an update is done in the track.

IanMayo:
> I am invalidating the cache each time ...

I _think_ we need to be careful about catching all changes to the track.  In the previous optimisation, we generated the value(s) fresh each time - so changes to the underlying data didn't matter.  

We _may_ be able to take our cache invalidation problem away by just clearing it at the start of TrackWrapper.paint().  We will have to calculate it once in the paint, but we won't have to worry about invalidation.

What do you think?

IanMayo:
Oh, and @saulhidalgoaular  - do you have any timings for the performance improvement?

org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
```diff
@@ -1385,6 +1385,7 @@ public void clearPositions()
   @Override
   public final void closeMe()
   {
+    flushBounds();
```
IanMayo: I _think_ we should always flush the cache _after_ making data changes - so this should appear at the end of this message.

org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
```diff
@@ -2015,15 +2005,15 @@ public FixWrapper getBacktraceTo(final HiResDate searchTime,
   @Override
   public final WorldArea getBounds()
   {
-    if (cachedBound == null)
+    _bounds = null;
```
IanMayo: I'm sure we shouldn't be clearing the cache for each `getBounds()` 😆

IanMayo:
@helenayele - this change aims to being a speed increase when plotting very large tracks.

We've got the speed increase by caching the track size (bounds) - reducing the number of times we have to recalculate it.

But, we have a challenge in `cache invalidation`.  We have to be sure that each time something changes, we empty the remembered bounds, so it gets recalculated.

So, when testing this PR, it's worth trying to consider lots of ways of changing tracks: 
* turn track segments on and off (for TMA tracks)
* dragging TMA tarcks
* resampling tracks
* turning sensor data on & off

You should then 'fit to window' and ensure Debrief is using the correct size for the data.

Note: this part of the testing isn't really about large track performance - so you can use any data-files you like :-)

org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
saulhidalgoaular: Oh, ok

org.mwc.debrief.legacy/src/Debrief/Wrappers/TrackWrapper.java
saulhidalgoaular: oh yes. I did the commit while I was doing the performance test. :(

IanMayo:
@saulhidalgoaular - the logic for this seems more complex than necessary.  If we're only interested in the rendering performance - we only have to make changes to support the `paint()` method.  We can leave the other `getBounds()` calls unchanged, and only address them if/when they cause performance problems.   By leaving the other `getBounds()` calls unchanged, we reduce the chance of knock-on effects.

IanMayo:
Dropping - simpler fixes in #4754

PR #4754, opened 2020-01-27
Reduce calls to getBounds() during D-Lite render process

IanMayo:
Don't override parent getDataArea call

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GenericData/WorldArea.java
+++ b/org.mwc.cmap.legacy/src/MWC/GenericData/WorldArea.java
@@ -788,6 +788,17 @@ public final class WorldArea implements Serializable
 			ww5.extend(wa3);
 			assertTrue("Extending using area", ww5.equals(wa3));
 		}
+		
+		public final void testExtend2()
+		{
+      final WorldArea ww4 = new WorldArea(wa1);
+      final WorldArea ww5 = new WorldArea(wa1);
+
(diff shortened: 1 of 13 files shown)


org.mwc.cmap.legacy/src/MWC/GenericData/WorldArea.java
```diff
@@ -788,6 +788,17 @@ public final void testExtend()
...
+		
+		public final void testExtend2()
+		{
+      final WorldArea ww4 = new WorldArea(wa1);
+      final WorldArea ww5 = new WorldArea(wa1);
+
+      assertNull("Null test OK for static extend", extend(null, null));
+      assertTrue("First Null for static extend.", ww4.equals(extend(null, ww4)));
+      assertTrue("Second Null for static extend.", ww4.equals(extend(ww4, null)));
+      assertTrue("Extending using area static extend", wa3.equals(extend(ww5, wa3)));
```
IanMayo: I'm nervous about the test on line 800.  If we've extended ww5 to include wa3, then it should not be the same as wa3.  At least, I don't think so.

Hmm, it's too late to adopt immutable objects for this, but it would be nice to slowly move towards them :-)

IanMayo: Aah, @saulhidalgoaular  - don't worry, I've got it.  wa3 includes ww5 - so it's right to return wa3.  Sorry to doubt you :-)

PR #4758, opened 2020-01-28
4752 Deleting cache

### LINKED ISSUE #4752: Outdated Bounds for LightweightTracker

saulhidalgoaular:
Fix #4752

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
@@ -734,7 +734,6 @@ public class Plottables implements Plottable, Serializable, PlottablesType,
   @Override
   public void propertyChange(PropertyChangeEvent arg0)
   {
-    // TODO Auto-generated method stub
     
   }
 
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
@@ -197,11 +197,6 @@ public class Plottables implements Plottable, Serializable, PlottablesType,
 	 */
 	private boolean _visible;
 
-	/**
-	 * remember the area covered by the plottables
-	 */
-	private WorldArea _myArea;
-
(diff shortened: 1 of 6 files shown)


IanMayo:
@saul - do you know there isn't an impact from removing the cached area in Plottables?

saulhidalgoaular:
> @saul - do you know there isn't an impact from removing the cached area in Plottables?

Good afternoon. I have confirmed that it is needed to fix the bug. It is a cache that is not being validated after modifying the elements.

org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
```diff
@@ -347,15 +347,8 @@ public synchronized void paint(final CanvasType dest)
...
-		{
-			// yup, get on with it...
-			_myArea = recalculateAreaCovered();
-		}
-
-		return _myArea;
-
+		// yup, get on with it...
+		return recalculateAreaCovered();
 	}
```
IanMayo: so, @saulhidalgoaular  - shouldn't we remove the `_myArea` field?

saulhidalgoaular: oh yes. 1 min

IanMayo:
Hey @saulhidalgoaular  - the build isn't happy :-)

IanMayo:
@helenayele  - we've merged a whole bunch of PRs today.  Some of them are quite old, and there's a chance something has changed in Debrief since they were originally tested.

So, while testing this PR (which relates to storing the bounds of items), it would be good to do some double-checking more widely across D-Lite and Full Debrief.  Just so we know we're on a good foundation before we do an automated code cleanup.

PR #4772, opened 2020-02-04
4743 Correct Imea import dialog position

### LINKED ISSUE #4743: Display `Import NMEA` dialog on correct monitor
An analyst is dragging some NMEA into Full Debrief.  A sample of this file format is in `other_formats\NMEA_TRIAL.log`.

When these files are dragged in, the `Import NMEA Data` dialog opens.

But, on this analyst's multi-monitor setup, the dialog is show in the centre of the first monitor, but Debrief is on monitor 2.    That meant that repeatedly the analyst was unaware of the dialog waiting.

The `Import NMEA Data` dialog should open either at the centre of the monitor currently displaying Debrief, or in the centre of the Debrief application windows.

On my MACOS, the dialog is shown on the Primary monitor.  Sometimes that's right for Debrief, on other occasions it isn't.

saulhidalgoaular:
Fix #4743


IanMayo:
We can only really test this on a multi-monitor. I've just tested it, and it works fine, whether Debrief is on the "Primary" or "Secondary" monitor.

PR #4792, opened 2020-02-17
4782 Remove validation for Re-sample data at property

### LINKED ISSUE #4782: Remove validation for Re-sample data at property
After opening positions property window, and saving without modifying anything, it popups "Invalid property value" dialog box that state that Resample data at property has invalid value :
[Image from Gyazo](https://gyazo.com/0c8d2dca56776252cd7e6e68c6e8e580)

It shouldn't show this error dialog as user doesn't modify anything.

### Steps to reproduce:
1. Load boat1.rep
2. Double click '120500.00(402 positions)' layer inside Track:Nelson layer to open its property
3. After the property window is opened select Apply without modifying anything
See that the above error dialog box is shown for user.

saulhidalgoaular:
Fix #4782

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
@@ -1091,7 +1091,7 @@ abstract public class PlainPropertyEditor implements PropertyChangeListener {
 
 		// find out if the value for this parameter is different to the
 		// current one for this object - or if there isn't a current value
-		if (newVal != oldVal && (newVal == null) || ( newVal != oldVal && !newVal.equals(oldVal))) {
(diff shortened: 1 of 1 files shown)


org.mwc.cmap.legacy/src/MWC/GUI/Properties/PlainPropertyEditor.java
```diff
@@ -1091,7 +1091,7 @@ private void update(final PropertyDescriptor pd, final Object data) {
 
 		// find out if the value for this parameter is different to the
 		// current one for this object - or if there isn't a current value
-		if ((newVal == null) || (!newVal.equals(oldVal))) {
+		if (newVal != oldVal && ((newVal == null) || ( !newVal.equals(oldVal)))) {
```
IanMayo: @saulhidalgoaular  - is there a chance of a null-pointer exception here?

It _looks like_ a null newVal could get passed into the right-hand clause.

PR #4793, opened 2020-02-18
Implemented dialog for selecting narrative types

### LINKED ISSUE #4764: Let analyst select what types of narrative to import
Analysts import diaries of timed comments from MS Word documents in a specific format. These entries are called "narratives".

The current logic works through the MS Word document (or pdf, actually), and loads each line directly into the plot.

But, sometimes the analyst is only interested in a subset of the entry types.

So, we should load the entries into memory, and let the analyst choose which types to load.

We should _probably_ change the current logic from loading the data immediately, to storing the entries into a hashmap, indexed by what type of entry it is (`narrativeEntry.type`).   Once they're all loaded, popup a custom dialog that shows a list of the types, with checkboxes against each.  There should also be `Select all` and `Select None` buttons.

I guess the above logic could be injected at line 2465 of ImportNarrativeDocument.  Once that import loop is complete we would call a new selectTypesToImport helper (to be created).  This will be initialised similar to the `setNarrativeHelper` processing. We'll have a _dumb_ one for our unit tests, and a GUI version for actual users.

If the user presses `Cancel` the import is cancelled. But, if the user presses `Ok` just those selected types of entries are loaded.

Here's an example of raw narrative content, taken from the `FCS_Narrative.doc` sample data-file:
image

Most of the early entries are of type `OOW COMMENT`, but the later ones are `FCS`.

I _expect_ there to be from 5-15 types of entries in a real import document.

ayeshare:
Fixes #4764 

Added a dialog to select narrative types


IanMayo:
Nice work @ayeshare - though I think the `code cleanup` should make some more of the fields final, particularly in the custom dialog.

helenayele:
When I load FCS_narratives.pdf file on blank plot, I get it loaded with weird characters. Can we avoid loading on empty plot.
[Image from Gyazo](https://gyazo.com/48a87e3ec0e6cb226b6c24d48fba5c7a)

ayeshare:
> When I load FCS_narratives.pdf file on blank plot, I get it loaded with weird characters. Can we avoid loading on empty plot.
> [Image from Gyazo](https://gyazo.com/48a87e3ec0e6cb226b6c24d48fba5c7a)

This is a different issue altogether. This is how eclipse workbench handles any file it doesnt know how to open.

helenayele:
When selecting Cancel button of this custom import  dialog, it imports all narrative entries' type
[Image from Gyazo](https://gyazo.com/da9b7162c330f10d65a7986be4770f1a)

helenayele:
Still narrative entries are imported after selecting cancel button of custom narrative entries import wizard 

[Image from Gyazo](https://gyazo.com/be95f2df33b87084426aa40868631166)

IanMayo:
Hello @ayeshare - couple of to-dos:
* [ ] The list of entry types is in a scrolled area with a size that's taller than the list of items. This means there is an unnecessary scroll bar, and it's possible to scroll to where there aren't any check-boxes
* [ ] For occasions when there are lots of entry types, a user may wish to expand the dialog to view all of them. It's a simple flag to make it resizable [1.]
* [ ] You should apply the code cleanup to your modified files

[1.] https://stackoverflow.com/questions/26673121/resizeble-dialog-in-java-swt

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Word/ImportNarrativeDocument.java
```diff
@@ -2184,55 +2376,72 @@ public void processThese(final ArrayList<String> strings) {
...
-						}
+						type = thisN.type;
+					} else if (thisN.platform != null) {
+						type = "None";
+					} else {
+						type = null;
+					}
+					if (type != null) {
+						if (typeVsCount.get(type) == null) {
+							typeVsCount.put(type, Integer.valueOf(1));
```
IanMayo: @ayeshare - I'm pretty sure you can just used `1`, rather than `Integer.valueOf(1)`

IanMayo:
Ok, @helenayele - narrative visibility is a separate issue.  If it isn't recorded already, could you create an issue for it?

IanMayo:
@helenayele - I've made some refactoring changes, I could do with you re-testing it, please.

helenayele:
Okay I will re-test it

helenayele:
> Ok, @helenayele - narrative visibility is a separate issue. If it isn't recorded already, could you create an issue for it?
Oh , this is not reproducible in develop branch , I think Narratives visibility is fixed by Saul last time and merged already.

PR #4806, opened 2020-03-04
Fixed an issue with filter

### LINKED ISSUE #4788: Adjust time-controller coverage after "Remove not visible:
We have a Full Debrief issue where the Outline View rendering slows down the deletion of large numbers of positions (1000s).  So, there is now a method available when right-clicking on the `Positions` child of a track.  If the analyst uses `Filter to time period` to focus on a limited window of data, he/she can use `Delete positions not currently visible` to delete the other data.  This happens effectively instantly.

But, there's an issue:
> In the Time controller I have 6 days of track, I filter the time period to one day,
> then “delete positions not currently visible”, the Time Controller still shows all 
> 6 days. “I click on Filter Plot Data to Selected Time Period” and I still have all 
> six days on the Time Controller.
> 
> All the data before and after has been deleted.

So, after we've run this operation, we have to trigger the operation that updates the bounds of the Time Controller.  The data currently loaded _may_ still cover the original period if several tracks are loaded. But, if there is only one large track, then the Time Controller period should shrink to match the newly trimmed track.

ayeshare:
Fixes #4788 
the filter was not working if the slider was changed when the toggle was not on

### DIFF
--- a/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/TimeController.java
+++ b/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/TimeController.java
@@ -132,6 +132,10 @@ public class TimeController extends ViewPart implements ISelectionProvider, Time
 				if (tNow != null) {
 					final TimePeriod period = _controllablePeriod.getPeriod();
 					if (period != null) {
+						_wholePanel.setEnabled(true);
+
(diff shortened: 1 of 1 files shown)


IanMayo:
@helenayele - what did you do to get the new behaviour?

helenayele:
We need to turn on the "Filter Plot data to selected time period' make it work.

PR #4807, opened 2020-03-05
enable close button after loading nmea file

### LINKED ISSUE #4770: Close button is disabled for NMEA file types
After opening NMEA_TRIAL.log file, Close button remains disabled. And there is no way to use this close button to close such file types.
But we can use File > New button to close the file and reset D-Lite.

### Expected Result :
We should be able to use Close button to close such files types like the other files types.

ayeshare:
Fixes #4770


IanMayo:
@helenayele - I _expect_ we get the same behaviour across all importers:  close is enabled if the file import completes, regardless of the volume of data loaded.  

Let's not worry about it ;-)

Issue #4800, opened 2020-02-26
Early steps towards Database import


IanMayo:
Tomorrow we're holding a workshop with users to discuss their requirements for pulling database data into Debrief.

The backend database ("Pepys" - pronounced _Peeps_) and it's import library is still under construction right now. But, we can start to take some exploratory steps towards importing the data.

There may be subtle changes to the schema, but it's now fairly stable:
https://docs.google.com/drawings/d/1CXE6pW6lvAKcCP8O1spRNSYBwl6ajJ3HX9dCSRm7nK0/edit?usp=sharing

Eventually the users will be connecting to a PostGres networked database. but for our own testing we're using standalone SQLite database files.  I've attached one that contains 543 positions (states) for 29 platforms (ship), taken from 25 data-files.

Old one (no locations): [test.zip](https://github.com/debrief/debrief/files/4256426/test.zip)
Updated file (with location): 
[test2.db.zip](https://github.com/debrief/debrief/files/4299173/test2.db.zip)

We won't be ready to start implementing the UI until Celine has designed it.  But, we can _hit the ground running_ if we perform some non-UI setup steps:
* create new Debrief plugin:  org.mwc.debref.pepys
* incorporate SQLite JDBC driver jar(s)
* create new Pepys class in the plugin, that can:
    * connect to a SQLite database file
    * list data-files in the database
    * return all of the *State* measurements for a database id, as a series of TrackWrapper objects, containing one FixWrapper for each State record.

IanMayo:
Hello @saulhidalgoaular  - we now have a IU design:
[Debrief-Pepys-importer-UI-20200302.pdf](https://github.com/debrief/debrief/files/4274602/Debrief-Pepys-importer-UI-20200302.pdf)

The database meets this schema:
https://docs.google.com/drawings/d/1CXE6pW6lvAKcCP8O1spRNSYBwl6ajJ3HX9dCSRm7nK0/edit

We may need to make some database changes to allow the necessary querying/filtering we require.  

Aah, we may also need some more testing data in the database - particularly for Tasks.

IanMayo:
@saulhidalgoaular  - for the Time Period filter, we already have some usage of the Nebula Data and Time controls in the `Maintain Contributions` View, screenshot here:
![image](https://user-images.githubusercontent.com/1108513/75871105-f239ab80-5e03-11ea-88cd-026217bd7fc3.png)

IanMayo:
[...]

PR #4809, opened 2020-03-06
[WIP] - 4800 early steps db

saulhidalgoaular:
Support #4800


org.mwc.debrief.pepys/build.xml
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
IanMayo: @saulhidalgoaular - we don't use `build.xml` for other RCP projects.

IanMayo:
@saulhidalgoaular - I added `debrief.pepys` to the `org.mwc.debrief.combined.feature` - so it gets included in the product deployment.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Replay/ImportTrackSplit.java
```diff
@@ -0,0 +1,52 @@
+package Debrief.ReaderWriter.Replay;
+
```
IanMayo: @saulhidalgoaular  - I wasn't expecting this file. I presume it's unrelated to the `Pepys` work, right?

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Comments.java
```diff
@@ -0,0 +1,134 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *******************************************************************************/
+
+package org.mwc.debrief.pepys.model.bean;
+
+import java.util.Date;
+
+public class Comments implements AbstractBean, FilterableBean, TreeStructurable{
```
IanMayo: @saulhidalgoaular - since this represents a single row, it should be called `Comment` in my opinion.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/FilterableBean.java
```diff
@@ -0,0 +1,24 @@
...
+ * modify it under the terms of the Eclipse Public License v1.0
+ * (http://www.eclipse.org/legal/epl-v10.html)
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *******************************************************************************/
+
+package org.mwc.debrief.pepys.model.bean;
+
```
IanMayo: @saulhidalgoaular - for us to do this filter in Java, we have to load all the rows, and then apply the filter - on the client.

I _hope_ we can apply the filter at the server, and only get the datafiles that contain data matching the field.  Then, when the user selects "Import data" we get all the rows, but SQL filtered (_where_) according to time period or area of coverage.  Then, we only pull data onto the client (deserialise) that is actually going to be loaded into Debrief.

Happy to talk this through further, if necessary.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Contacts.java
```diff
@@ -0,0 +1,336 @@
...
+
+package org.mwc.debrief.pepys.model.bean;
+
+import java.util.Date;
+
+import MWC.GenericData.WorldLocation;
+
+public class Contacts implements AbstractBean, FilterableBean, TreeStructurable{
+
+	private int contact_id;
```
IanMayo: @saulhidalgoaular - there are lots of fields here, and I fear we may have to change the file to match schema changes that don't relate to Debrief. So, I'd suggest only storing those that can be exploited by Debrief.  Aah, maybe you were waiting for me to tell you which ones :-D

* time
* bearing
* rel_bearing
* freq
* major
* minor
* orientation
* classification
* subject_id
* source_id

Aah, actually, subject and source would be better than subject_id and source_id.  I'll pass that task to the database chap.

* location
*

IanMayo:
Hello @saulhidalgoaular - all good, though just the one issue about us trying to do filtering on the server, rather than the client.  Let me grab a diagram:
image


So, we use the filters on the left-hand side to give the database server our bounds.  It will return a tree-style set of query results that lists a set of datafiles.   The analyst will select which ones he/she wants to load. Then, when the `Import` happens, we run a series of database queries, getting the contents of each datafile, but we filter the query according to time/location bounds.  So, we only receive rows (measurements) at the client that meet the filters.  All other filtering is done at the server.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Comments.java
saulhidalgoaular: Good morning. Yes. I totally agree that. I would also suggest the change of the names in the database tables to singular because they represent entities.

The thing is that I am using reflection to get the name of the table, from the class name. Changing it would mean the change of how I would retrieve the table now.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/FilterableBean.java
saulhidalgoaular: Correct. My idea was to filter on the server. I created this interface to have access to the field names to prepare the query for the server.

IanMayo: Aah, that's clever :-D

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Comments.java
IanMayo: > using reflection to get the name of the table

Aah yes, there are two conventions for table names. We took the decision a year ago to use plural names.

So, yes - we should keep `Comments` if we are going to use reflection.  But alternatively, is it an option to use a new annotation to store the @TableName?

Then we could have `Comment` in Java for a single comment instance, with the knowledge that the data is coming from the `Comments` table?

(this isn't very important to me, I'm just offering a suggestion).

IanMayo:
@saulhidalgoaular - can we have a busy-cursor while the `Apply` is running?

We get a new dialog each time we click on the button.  It would be better if the existing dialog just got pulled to the front, I think.

PR #4819, opened 2020-03-13
reset toggle state of zoomin button

### LINKED ISSUE #4424: When resetting D-Lite all the view menu items should hold the original state
Currently If I select 'Drag Whole Element' or 'Drag Element' from view menu and select New , It still keeps the selected state of 'Drag whole Element' or 'Drag Element' buttons.

It should select Zoom-in button by default - it should hold the original state when D-Lite launches.

ayeshare:
Fixes #4424


helenayele:
After reset the following items are still holding the previous set value :
1. Transparency
2. Time menu > Display mode, if I select snail display mode, it will be kept set after reset. 
3. Time Menu > Highlighter type , if I select Symbol highlighter type, it will be kept still selected after reset.
4. Filter to time control partially , say if I change the start and end time values of a track and reset it by Selecting file > New, and open the file again (rep file) , the start date will hold the correct default location but the end date will be located on the old location but will have the correct default end value .  Steps for this particular issue is 
1. Load boat1.rep
2. Change the start and end time to some values
3.  Reset
4. Load again boat1.rep file
5. See that End time hold the tracks' end time but the location of the end time is not reset to its original place.
[Image from Gyazo](https://gyazo.com/da68bfce4bdee8988dea52320b19007b)

ManigandanMS:
> As can be seen , in addition to the above issues , the following items are still not changed to default
> 
> 1. Auto step
> 2. Font size
> 3. Large step
> 4. Small step
> 
> And in Snail mode, All Edit display properties are not set to default after reset:

> these are :
> 
> 1. Fade points
> 2. Link positions
> 3. Plot names
> 4. Point size
> 5. Trail length
> 6. Vector stretch

@IanMayo  Resetting all these means we need to have a framework where there should be a way to reset in a single method call.
I looked up the code and I see that  Editable.java ( interface ) there should be a method added resetToDefaults() and reset implementation must be done here.

ayeshare:
> As can be seen , in addition to the above issues , the following items are still not changed to default
> 
> 1. Auto step
> 2. Font size
> 3. Large step
> 4. Small step
> 
> And in Snail mode, All Edit display properties are not set to default after reset:

> these are :
> 
> 1. Fade points
> 2. Link positions
> 3. Plot names
> 4. Point size
> 5. Trail length
> 6. Vector stretch

@IanMayo  Resetting all these means we need to have a framework where there should be a way to reset in a single method call.
I looked up the code and I see that  Editable.java ( interface ) there should be a method added resetToDefaults() and reset implementation must be done here.

IanMayo:
@helenayele  - I'm ok with the time step properties remaining the same

The analyst may have configure the step properties to perform a particular type of analysis (such as setting the step size to one day when looking at plots covering a long period).  If the analyst is doing this through a series of files, it could be frustrating to re-configure it each time.

IanMayo:
@helenayele - it does still seem worth ensuring the slider bars are updated correctly:
> See that End time hold the tracks' end time but the location of the end time is not reset to its original place.

PR #4821, opened 2020-03-17
Added hasnobounds

### LINKED ISSUE #4230: Expand shapes by default during insertion
When user adds shape using insert menu, the shapes are not fully shown unless user expands them using Fit to window action. It will be very handy for the user if the shapes gets expanded as soon as user inserts them.

Video:
[Image from Gyazo](https://gyazo.com/8d00ff4a9d1340f92f959d32b242afcb)

ayeshare:
Fixes #4230 

 to check whether shape is already added before rescaling

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
@@ -1410,6 +1410,9 @@ public class Layers implements Serializable, Plottable, PlottablesType {
 				Plottable p = (Plottable)edits.nextElement();
 				if(p.getBounds()!=null) {
 					noBoundsCount++;
+					//the very first shape is already added at this stage,
+					//so we check for future additions to prevent rescaling every time
+					// a shape is added
 					if(noBoundsCount>1) {
 						hasNoBounds=false;
 						break;
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
@@ -1398,5 +1398,26 @@ public class Layers implements Serializable, Plottable, PlottablesType {
 
 		}
 	}
+	
+	public boolean hasNoBounds(Plottable newItem) {
+		boolean hasNoBounds = true;
+		int noBoundsCount=0;
+		Enumeration<Editable> enumer = elements();
+		while(enumer.hasMoreElements()) {
+			Layer item = (Layer)enumer.nextElement();
+			Enumeration<Editable> edits = item.elements();
+			while(edits.hasMoreElements()) {
+				Plottable p = (Plottable)edits.nextElement();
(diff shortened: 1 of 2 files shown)


org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
```diff
@@ -1398,5 +1398,26 @@ public void walkVisibleItems(final Class<?> mustMatch, final OperateFunction fun
...
+		boolean hasNoBounds = true;
+		int noBoundsCount=0;
+		Enumeration<Editable> enumer = elements();
+		while(enumer.hasMoreElements()) {
+			Layer item = (Layer)enumer.nextElement();
+			Enumeration<Editable> edits = item.elements();
+			while(edits.hasMoreElements()) {
+				Plottable p = (Plottable)edits.nextElement();
+				if(p.getBounds()!=null) {
+					noBoundsCount++;
```
IanMayo: @ayeshare - why do we check that `noBoundsCount` is `> 1`?

Shouldn't we drop out after the first non-null from `p.getBounds()`?

ayeshare: If we drop out after the first non-null, then it will never rescale, because, the first time we are adding the very first shape.

IanMayo: ahh, I see. Thanks.

org.mwc.cmap.legacy/src/MWC/GUI/Layers.java
```diff
@@ -1410,6 +1410,9 @@ public boolean hasNoBounds(Plottable newItem) {
 				Plottable p = (Plottable)edits.nextElement();
 				if(p.getBounds()!=null) {
 					noBoundsCount++;
+					//the very first shape is already added at this stage,
+					//so we check for future additions to prevent rescaling every time
+					// a shape is added
 					if(noBoundsCount>1) {
 						hasNoBounds=false;
 						break;
```
IanMayo: @ayeshare - I'm pretty sure this will only break out of the inner loop.  I think it would be fixed if you switch to `return false`, and have `return true` at the end of the method.

IanMayo:
Helen - it's by intention that we don't zoom-to-fit every time we add a shape.

IanMayo:
@helenayele - it has been updated, could you re-test please?

IanMayo:
@helenayele - the main issue we're looking at here is when a new plot is being used.

If this is the first shape drawn onto a plot, the plot should resize to fit that shape, rather than just be a map of the world:
https://github.com/debrief/debrief/issues/4230

helenayele:
Okay it works perfect for the fist item.
[Image from Gyazo](https://gyazo.com/b043190e5a13704b10d24afca8195db0)

IanMayo:
(did close/reopen to try to wake up CI reporting)

PR #4822, opened 2020-03-18
tooltip for change visibility

### LINKED ISSUE #4093: Provide tooltips for undo/redo buttons
D-Lite now has undo/redo buttons.

It may help the analyst if they knew what operation they were about to undo/redo.

We can retrieve the name of the next operation on the undo/redo buffer.  We should be able to dynamically set the tooltip on the buttons.

Well, we could either manually change the tooltips each time the UndoBuffer is modified.  Or, we could _somehow_ provide an accessor that provides the relevant text each time the tooltip is about to appear.

Flamingo appears to have gone through a major change for it's 2.0 release.  We should probably defer this change until after then.

The radiance upgrade is covered in #4094

ayeshare:
Fixes #4093


IanMayo:
@ayeshare - I can make @helenayele 's changes.

PR #4828, opened 2020-03-23
4815 reading config file

### LINKED ISSUE #4815: Pepys Database Import - get config via Env Variable
The database connection details will be provided in an `INI` file. The path of the file will be provided in an environmental variable.

At startup, Debrief (or the Pepys viewer plugin, at least) will look for the environmental variable titled `PEPYS_CONFIG_FILE`.

The contents of this file will look like this: https://github.com/debrief/pepys-import/issues/195

Debrief will extract the connection parameters from the `database` section of the file.  The file will (initially) only require SQLite or PostGres connections.

saulhidalgoaular:
Fix #4815


org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/ImportDatabase.java
```diff
@@ -49,17 +56,38 @@ public Layers getLayers() {
 		};
 
 		final Shell shell = new Shell(PlatformUI.getWorkbench().getDisplay());
+		final DatabaseConfiguration databaseConfiguration = new DatabaseConfiguration();
 		try {
-			new SqliteDatabaseConnection().createInstance();
-			// new PostgresDatabaseConnection().createInstance();
-		} catch (final PropertyVetoException e) {
+			DatabaseConnection.loadDatabaseConfiguration(databaseConfiguration);
+
```
IanMayo: the details of the next block look a little untidy, with unused return parameters from the `createInstance()` calls. Let me see if I can improve it.

IanMayo: Nope, failed: https://github.com/debrief/debrief/pull/4830

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/db/config/ConfigurationReader.java
```diff
@@ -0,0 +1,120 @@
...
+			final DatabaseConfiguration databaseConfigurationTest = new DatabaseConfiguration();
+			parseConfigurationFile(databaseConfigurationTest, stream);
+
+			// Check Database Categories
+			final HashMap<String, String> databaseCategory = databaseConfigurationTest.getCategory("database");
+			final HashMap<String, String> archiveCategory = databaseConfigurationTest.getCategory("archive");
+
+			assertTrue("databaseCategory - size", databaseCategory.size() == 5);
+			assertTrue("archiveCategory - size", archiveCategory.size() == 2);
+
```
IanMayo: Could we have difference values for `db_username` and `db_password`?  

The code is _slightly_ weaker if it's testing for the same string twice.  We _may_ get username instead of password, but still get the same results string.

So, `abcdef` for username is fine, but password should be different, like `uvwxyz`.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/db/config/ConfigurationReader.java
```diff
@@ -0,0 +1,120 @@
...
+	static final String DEMILITER = "_";
+
+	static final String CATEGORY_START = "[";
+
+	static final String CATEGORY_END = "]";
+
+	private static final String CONFIG_SEPARATOR = "=";
+
+	private static final String CONFIG_COMMENT_CHAR = "#";
+
```
IanMayo: We have quite a lot of code to handle `ini` files.  

Would this method be an alternate solution? https://stackoverflow.com/a/41084504/92441

org.mwc.debrief.pepys/postgres.ini
```diff
@@ -0,0 +1,7 @@
+[connection]
```
IanMayo: This should be `database` - rather than `connection`.  

I guess the format changed at some point.

IanMayo:
@saulhidalgoaular - what's the status of this PR?

Does it include the code that was merged in 4815b?

saulhidalgoaular:
> @saulhidalgoaular - what's the status of this PR?
> 
> Does it include the code that was merged in 4815b?

Yes. Right now it has all my commits that were in `4815b`

PR #4835, opened 2020-04-01
4833 clear area bounds

### LINKED ISSUE #4833: Pepys - clear Area bounds
I filtered to an area (the current viewport).

Then I deleted the value in the area fields.  But, when I filtered, it was still restricted to those bounds.

So, we should not include an area filter if any of the text fields are empty.

saulhidalgoaular:
Fix #4833


IanMayo:
@saulhidalgoaular - could you resolve the conflict in `PepysImportController` please?

saulhidalgoaular:
> 
> 
> @saulhidalgoaular - could you resolve the conflict in `PepysImportController` please?

Good afternoon. Solved.

PR #4836, opened 2020-04-02
set max limit for fixing timecontroller hanging in large files

### LINKED ISSUE #4831: Pepys - Debrief Time Controller can't handle very long periods
I've loaded some Pepys data using a spatial query (no time filter).

It has 20 years of tracks loaded, but they're low volume. 4 tracks, all less than 500 points.

But, the time-controller is really, really slow.

We should run it in the profiler - and find/remove the bottleneck.

Note: this isn't Pepys specific. I'll try to produce a plot-file, to reproduce it.

[boat_long_period.rep.zip](https://github.com/debrief/debrief/files/4386281/boat_long_period.rep.zip)

ayeshare:
Fixes #4831

### DIFF
--- a/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/SliderRangeManagement.java
+++ b/org.mwc.cmap.TimeController/src/org/mwc/cmap/TimeController/views/SliderRangeManagement.java
@@ -117,16 +117,16 @@ abstract class SliderRangeManagement {
 
 		}
 	}
-	
-  private static final long DAY_STEP = 24 * 60 * 60000000L;
 
-  private static final long HOUR_STEP = 60 * 60000000L;
+	private static final long DAY_STEP = 24 * 60 * 60000000L;
 
-  private static final long MINUTE_STEP = 60000000;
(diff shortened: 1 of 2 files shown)


IanMayo:
Hmm, I'm worried about this fix.  

It avoids the performance problem, but I think it breaks the logic.  We rely on knowing the number of intervals on the slider to work out which date is represented.

When I play with `Filter to time period` The slider doesn't adjust to the dates I expect to see.

IanMayo:
@helenayele  - I'm pretty sure the Play on filter-to-period issue is an existing one. If that's true, I'd like it to be a separate issue, please.

helenayele:
Okay , tried to see how it behaves in develop branch, since full debrief doesn't respond well on this file, unable to test the slider on time controller.
[Image from Gyazo](https://gyazo.com/963723569731b1afc9b01c0ff22fee09)

But let's approve this PR and since the core issue is the responsiveness of full debrief.

helenayele:
> @helenayele - I'm pretty sure the Play on filter-to-period issue is an existing one. If that's true, I'd like it to be a separate issue, please.

Okay have filed it as separate issue at #4843 4843

PR #4846, opened 2020-04-04
4816 check all items in tree

### LINKED ISSUE #4816: Pepys Database Import - UI improvements
Here are some minor UI improvement requests. I've tried to sort them in priority order
* [x] Implement time filter support
* [x] Implement spatial filter support
* [ ] ~~"Mock" the `Use current viewport` by providing this bounding box:
(60N, 6W), (50N, 4W)~~ (no, we no longer need this)
* [x] Introduce column showing the time period covered by that data-file
* [x] Clicking a checkbox clicks all the checkboxes beneath it.  Similar for un-clicking.
* [x] Implement filtering by data-type
* [x] Add "connection test" button - this will use the config file (from env, if provided, to a test-connect, then confirm which database is connected to).  The button will have a label "Test".  A dialog box will open with a confirmation of success/fail.  Both will include connection details, but failure will include an error snippet.
* [ ] Introduce basic handling for typical errors: "database not found", "username not found", "wrong password", "not on this port", "sqlite file not found", ...
* [x] Implement free-text search over the right-hand listing. This will retain the whole _branch_ for matching items. So, it could be used to search for a platform (which will then show all it's children) or a data-file, (which will still display the data-file's parents)
* [x] Switch the "Apply" and "import" buttons to use the 48px icons for "Search" and "Import".  The 48px icons can be found in github.com/debrief/debrief-icons. If they're not already in `org.mwc.debrief.core/icons/48`, then they should probably be copied there, and then referenced from that plugin.
* [x] Introduce icons for `State`, `Contact` and `Comment` entries in the Data-Type search box. From the `debrief-icons` project, these icons can be used (repectively): `fix`, `bearing`, `narrative`.
* [x] Introduce icons in the collapsible-panel tabs:  `control_time`, `map` and `filter` icons, respectively
* [x] Adding busy cursor - https://github.com/debrief/debrief/pull/4809#issuecomment-601236035
* [x] Resizing to a bigger size
* [x] Bring the same Windows when the button `Import Database`  is pressed again.
* [x] Adding icons to the items of the tree.
* [x] Add 'Select All' checkbox to select all the available `State`, `Contact` and `Comment`
* [x] drag/drop .ini file onto database import dialog
* [x] drag/drop .sqlite database file onto database import dialog
* [x] Provide proper error message when user provides wrong db_type
* [x] Increase the size of the title

saulhidalgoaular:
Supports #4816

### DIFF
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/tree/TreeBuilder.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/tree/TreeBuilder.java
@@ -79,16 +79,18 @@ public class TreeBuilder {
 	 */
 	public static TreeNode buildStructure(final TreeStructurable[] items, final TreeNode root, final String filter) {
 		root.removeAllChildren();
+		final TreeNode subRoot = new TreeNode(TreeNode.NodeType.ROOT, "Database");
+		root.addChild(subRoot);
 
 		for (final TreeStructurable currentItem : items) {
 			final String platformName = currentItem.getPlatform().getName();
 			final String datafileName = currentItem.getDatafile().getReference();
 			if ((filter == null || filter.isBlank() || platformName.toLowerCase().contains(filter.toLowerCase())
 					|| datafileName.toLowerCase().contains(filter.toLowerCase()))) {
-				TreeNode datafileNode = root.getChild(platformName);
+				TreeNode datafileNode = subRoot.getChild(platformName);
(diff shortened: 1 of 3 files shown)


IanMayo:
Hello @helenayele  & @saulhidalgoaular  - I don't think it's better to expand the tree when "select-all" is clicked.  The expand button is available next to it - so it's only one extra click for the user. But, it may be a frustration if there's an unnecessary delay while a large tree is re-rendered.  (No, I don't know how large the tree could get :-) ).

PR #4858, opened 2020-05-02
4829 delta rate calculation

### LINKED ISSUE #4829: Extra Tote calculations
> I want rate of change of speed and rate of change of course as attibutes in the track tote

> that you can then plot in an xy or the stacked charts

We _could_ support these via toggle buttons in the Track Tote drop-down menu.  They would also appear as new options in the `Calculation` drop-down when opening an XY Graph.

saulhidalgoaular:
Fix #4829


helenayele:
@saulhidalgoaular  @IanMayo  , would you please provide me verification steps for this PR?

org.mwc.debrief.legacy/src/Debrief/Tools/Tote/Calculations/DeltaRateToteCalcImplementation.java
```diff
@@ -0,0 +1,102 @@
...
+
+			assertArrayEquals("Delta Rate Tote Calculation Rate", expectedAnswer, deltaRateRate, 1e-5);
+		};
+	}
+
+	public static double[] calculate(final double[] measure, final HiResDate[] time, final long windowSizeMillis) {
+		if (measure.length > 2 && time.length == measure.length) {
+
+			// Let's calculate the delta rate
+			final double[] deltaRate = new double[measure.length];
```
IanMayo, line 52: should this array be `length-1` - since we start at index 1. (since we're storing the intervals between the measurements)?

org.mwc.debrief.legacy/src/Debrief/Tools/Tote/Calculations/DeltaRateToteCalcImplementation.java
```diff
@@ -0,0 +1,102 @@
...
+			// Let's calculate the delta rate in a period.
+			final double[] deltaInPeriod = new double[measure.length];
+			final int[] countInPeriod = new int[measure.length];
+			double currentSum = 0;
+			int minSumIndex = 0;
+			int maxSumIndex = 0;
+			for (int i = 1; i < deltaInPeriod.length; i++) {
+				final long earlest = time[i].getMicros() - windowSizeMillis * 1000;
+				final long latest = time[i].getMicros() + windowSizeMillis * 1000;
+				while (minSumIndex < deltaRate.length && time[minSumIndex].getMicros() <= earlest) {
```
IanMayo, line 67: shouldn't it be `>=` earliest?

IanMayo:
good start :-)

IanMayo:
Saul, users will still need to be able to see graphs of normal measured course.

So, I think "course rate" and "speed rate" should be new instances of ToteCalculation.  Hmm, _or_, when we create the list of available calculations, we include them twice: once as "Course" and once as "Course Rate (abs)".

saulhidalgoaular:
Good morning, I also added the new calculation to Debrief Lite.

course

IanMayo:
Debrief-Lite - thanks @saulhidalgoaular , but that wasn't really necessary.  The new calculations are just for a few really, really nerdy scientists.  They're not the target audience for D-Lite.

Hmm, _maybe_ you should actually take it out from D-Lite, it may just scare the sea-riders.

PR #4863, opened 2020-05-06
4859 correct icons for tree

### LINKED ISSUE #4859: Make Icons uniform in the import Db dialog
We use different icons for the States, comments and Contacts in left and right columns of the db import dialog. See this screenshot:
[Image from Gyazo](https://gyazo.com/bb555bc71b46bf5e825599505b30590a)

See that we use different icons for states/comments/contacts in the left panel of this dialog. But we use the same icons for all these 3 objects in the right panel.
Can we use the same icons in the right panel as well?

saulhidalgoaular:
Fix #4859


IanMayo:
Can we have a screenshot please?

saulhidalgoaular:
Yes. 
Here is how it looks now.
image

IanMayo:
@saulhidalgoaular  - I don't think that footprint symbol is right.  Those are platform names, mostly ships, etc.

I can't see a "good" icon.  In my opinion, "multipath.png" is the _least worst_ option.

saulhidalgoaular:
Good afternoon. 

Here is an screenshot of the new look of the tree's platform node
image

IanMayo:
@helenayele  - I think the propeller icon is the "least worst" option, and we should probably use it.

PR #4864, opened 2020-05-07
4860 Disabling SRID for SQLite

### LINKED ISSUE #4860: Spatial filter support is not working
Currently filtering based on the current view port is failing to work. Whatever area we have in our plot, the states/contacts and comments are not filtered based on the current view port.

### Steps :
1. Open new plot area and zoom-in to Ethiopia or any other country
2. The try to select the Import db button and when the dialog opens go to the area filtering tab and select 'Use current viewport'  button
3. Select Apply
4. See that the whole db items are filtered. 
[Image from Gyazo](https://gyazo.com/534b74c4d415a1df5ba1598e2720e4b8)

Note 
This is regression - working before.

saulhidalgoaular:
Fix #4860


IanMayo:
@helenayele - had chance to check this?

PR #4877, opened 2020-05-23
4870 Centre Windows

### LINKED ISSUE #4870: Ability to centre XY-Plot on current time
It's possible to make an XY Plot "slide" with time, to only show a limited period of data.

1. In XY-Plot properties, set fixed duration toggle, plus duration
2. In XY-Plot toolbar (or drop-down), tick 'Grow Times'
3. Now shows set period running up to current time

But, it shows the window of time "before" the current time.  Analysts would like to the see window of time "around" the current time.

We would support this by adding a "Centre-Window" boolean option to the `Time-related` set of Properties in the XY-Plot properties panel. This will control how the time-axis is scaled/displayed.

Plot_Properties

saulhidalgoaular:
Fix #4870


org.mwc.cmap.legacy/src/MWC/GUI/JFreeChart/StepperXYPlot.java
```diff
@@ -219,6 +226,16 @@ public HiResDate getTimeZero() {
 	public boolean isGrowWithTime() {
 		return _growWithTime;
 	}
+	
+	private boolean _centreWindow;
```
IanMayo, line 230: @saulhidalgoaular - could member variables go near the top of the class, please - and could you also put in a quick comment to explain what `centreWindow` behaviour is?

PR #4883, opened 2020-05-29
4881 Import Button Improvement

### LINKED ISSUE #4881: Disable import button if nothing is selected and Add message for successful data files import
After user fetches data files from the server and selects "Import" button, it would be very nice if we provide a message like `{number of selected data files} data files successfully imported `  for successful import. And if the user doesn't select any data file , "import button" should be disabled . 

So the requirements here are 2 :

1- Disable import button if nothing is selected
2- Provide a message for successful import

saulhidalgoaular:
Fix #4881

### DIFF
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/controller/PepysImportController.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/controller/PepysImportController.java
@@ -353,7 +353,7 @@ public class PepysImportController {
 				if (event.type == SWT.Selection) {
 					final int importedItems = model.doImport();
 					final MessageBox messageBox = new MessageBox(_parent, SWT.OK);
-					messageBox.setMessage("Successful imported " + importedItems + " items");
+					messageBox.setMessage(importedItems + " data files is successfully imported");
 					messageBox.setText("Database Import");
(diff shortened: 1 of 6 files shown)


helenayele:
After fetching the latest code I get it now the import button is in disabled state while launching the import db dialog.
[Image from Gyazo](https://gyazo.com/99c94a44fd9a47486bdf01c4a055c2b2)

But we still need to restate the message of the dialog like 
[Image from Gyazo](https://gyazo.com/9237d81633437dede75a7e1f953ad625)
"1 data files is successfully imported"

helenayele:
Can we change also the title of this dialog like "Database Import" instead of  "Debrief NG" ?

helenayele:
The title of the dialog is changed but the message still not changed please see the above comment to get the correct message https://github.com/debrief/debrief/pull/4883#issuecomment-636137497
[Image from Gyazo](https://gyazo.com/213e53cbb94bf123909697fdcc061ef0)

helenayele:
I get the import button enabled after selecting apply button for the second time
follow this step : 
1. open pepys import button 
2.select Apply button 
3. Select one or 2 data files , this time import button becomes enabled 
4. Then select Apply button , this time nothing is selected but import button gets enabled.
Re-opening....

Issue #4723, opened 2020-01-10
Adopt Flamingo Split Button


IanMayo:
We've created our own Flamingo split-button that we use to optionally select the units for the Range/Brg tool: `SubstanceCommandToggleWithMenuButtonUI `

But, the Flamingo maintainer is developing a version within Flamingo: more [here](https://github.com/kirill-grouchnikov/radiance/blob/master/demos/flamingo-demo/src/main/java/org/pushingpixels/demo/flamingo/common/TestCommandToggleButtons.java#L116)

Once that code gets into a formal Flamingo release, we should switch from our own local copy to the formal Flamingo one.

helenayele:
Flamingo split-button is implemented for Rng\Brg button and it works as expected. 
See the below video:
[![Image from Gyazo](https://i.gyazo.com/c15498231e44b55061c77eb0433057dd.gif)](https://gyazo.com/c15498231e44b55061c77eb0433057dd)

When user selects the measurement units, this button will get selected by default.

Verified.

PR #4887, opened 2020-06-09
Adopt Flamingo Split Button (incl 3.0.0 upgrade) #4723

ayeshare:
Supports #4723 

Upgraded to flamingo 3.0.0


IanMayo:
I _hope_ @ayeshare  can include that capability in the drop-down menu handler:

So, if mouse is in zoom-in mode, and user selects units of `m`, then the mouse will be left in Rng-Brg mode.

PR #4889, opened 2020-06-10
4765 Verify overlapping legs

### LINKED ISSUE #4765: Avoid problem when Merge Track Segments contains overlapping legs
Analysts can use `Merge track segments` to merge two TMA tracks in Full Debrief

To reproduce:
* open `midflow2.dpf`
* open time sliders, to reveal all data
* select all of sensor data in `Sensor_A_2`, and generate a TMA solution from it.
* We should now have two TMA tracks
* Right-click on them, and select `Group tracks into 220415.22`
* Make this new combined track the `Secondary Track`
* The stacked dots rendering will fall over, since it's trying to generate multiple graph elements at the same time.  Errors will show on the console.

When the user clicks on `Group tracks into...`, we need to look at the data.  Each track has a set of legs, we need to find out if any legs overlap.  If they overlap with a `dynamic infill` we know we can just delete the infill. Indeed, one reason for combining the tracks could be that one fills in a large block of missing data in the other.

So, we need to loop through, say, the second track:
* for each leg, compare it to all the legs in the first track.  If it overlaps with a leg that's not a dynamic infill, remember the leg in the second track's name
* once we've looped through:
    * if there are overlapping legs, report the names of the overlapping legs to the analyst in a popup (message below)
    * if there weren't any overlapping legs, just combine the tracks:
        * loop through as before, but if the leg in the second track overlaps with a dynamic infill delete the dynamic infill
        * then insert the leg.

Title: Group tracks
Message: Cannot group tracks. The following legs from [second track name] overlap with legs in [first track name]:
[leg_1_name], [leg_2_name], etc, etc, etc

saulhidalgoaular:
Fix #4765


saulhidalgoaular:
Currently it shows a message error like this: 

image

helenayele:
@IanMayo  would you please merge this PR?

PR #4897, opened 2020-06-22
4891 latitude parse flexibility

### LINKED ISSUE #4891: Adding new position from Grid Editor
With the grid editor, it's possible to view the positions for a track:
image

It's also possible to add new positions.  But, it's very difficult to edit the location on the new positions.

I tried this:
* go to the bottom of the grid editor (that's actually the oldest points, it's in reverse order)
* click on the `row selector` for a row, in the left-hand column
* click on `insert row`.  It will create a duplicate of the current row
* select the new row. (it may have an empty `label` field
* change the time, to put it back 30 seconds.
* now try to edit the location, to move it two minutes up/down, and two minutes to the left/right
* but, as I leave the field the value returns to its previous position

Note: a couple of times this worked ok. But, when I went to record it, it failed :-)

saulhidalgoaular:
Fix #4891


helenayele:
@IanMayo  we need to merge this PR as well if nothing is left at your side.

IanMayo:
Demo of relaxed parser: 
grid_parse

PR #4900, opened 2020-06-26
Allowed multiple scales

### LINKED ISSUE #4129: Some console messages after undo/redo actions
The issue is mentioned at https://github.com/debrief/debrief/pull/4102#issuecomment-488693308
### Steps to reproduce

1. Launch D-Lite
2. Delete Chart Features layer
3. Add 2 shapes in different layers
4. Add 2 circles in one of the layer
5. Click undo so many times that no items are found in the map except chart features layer
6. Click again redo till all items are shown again
7. Insert 2 grid and 2 scale , but only 1 scale is created in Chart features layer
8. Click undo till all items are hidden
9. Click redo till all items are shown again - But see that grid is not shown
10 . Add Grid again
11. Click Undo
12. Click redo 
13. Repeat undo and redo , you will get atleast this message 
Failed trying to remove Grid from Chart Features (1 items)
or at most the explained console messages on the comment.

### Video that shows the issue:
[console_message_video.zip](https://github.com/debrief/debrief/files/3153100/console_message_video.zip)

ayeshare:
Fixes #4129. 

Implemented compareTo for ScalePainter to allow multiple scale objects with the same name by using hashcode comparison.


IanMayo:
@ayeshare  - CodeFactor is complaining that you've got an `equals` without a `hashCode()`.  If you moved the identity hash code into a `hashCode()` method, then `hashCode()` could be use for both `equals()` and `compareTo()`.

PR #4905, opened 2020-06-30
Implementation of Nisida Importer

### LINKED ISSUE #4896: Support new file format
Our users need to support a new file format.

The file format is detailed here:
https://docs.google.com/document/d/1rotbJ-yeSPKxcNND0v0JVF1LAh9VySVE9pNeC_ozz94/edit

We should create an importer in the `debrief.legacy` module, then support that importer in both Debrief and Debrief-Lite.

We can make the following assumptions about the format:
* there may be unexpected lines before, during, or after the lines of interest
* there may be multiple continuation lines for an entry. But, only the last "remarks" field may be extended.

saulhidalgoaular:
Fixes #4896


IanMayo:
Good start Saul :-). I was impressed by the comments, but then I realised they came from the Python version 🤣

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
```diff
@@ -0,0 +1,576 @@
...
+			if (track == null) {
+				track = new TrackWrapper();
+				track.setColor(DebriefColors.RED);
+				track.setName(platformName);
+				status.getLayers().addThisLayer(track);
+			}
+			status.setPlatform(track);
+
+			final String dateString = tokens[2];
+			final DateFormat dateFormatter = new SimpleDateFormat("MMMyy");
```
IanMayo, line 231: @saulhidalgoaular - this looks like unnecessary object creation `SimpleDateFormat` we should do it in the parent class, and just pass it in as a parameter

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
```diff
@@ -0,0 +1,576 @@
...
+	}
+
+	private static void processDetection(final String[] tokens, final NisidaLoadState status) {
+		// sample:
+		// [DayTime/DET/DetectingSensor/Bearing/Range in NM/TN/Own Lat/Own Lon/Position Source/Remarks/]
+		// 311200Z/DET/RDR/23/20/777/3602.02N/00412.12E/GPS/DETECTION RECORD
+		
+
+		/**
+		 *  Create a sensor using the expanded `Sensor Code` name plus the track number (TN).
```
IanMayo, line 468: @saulhidalgoaular  - if we can't find the sensor code, don't worry about throwing an error. Just use the short token.

These files are human generated, and can contain lots of typos.  So we can be relaxed about things.

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
saulhidalgoaular: https://github.com/debrief/debrief/pull/4905/commits/1fa769ea6d54c4c137a9dfe4761c7740a6e8aa3b

org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
saulhidalgoaular: Fixed at https://github.com/debrief/debrief/pull/4905/commits/b7787faa32aa4d068395ea853dae85884a2323e4

PR #4906, opened 2020-07-02
svg symbols copied to dlite dist

### LINKED ISSUE #4893: Can't use SVG symbol for tracks in Debrief-Lite
I was just recording a demo video for Debrief-Lite.

I tried to set the set the tracks symbol to be one of the SVG shapes. But, Debrief fell over:
image

There is a chance that we're not distributing the SVG shapes with Debrief-Lite.

ayeshare:
Fixes #4893 

copied svg_symbols folder to dist from cmap.legacy


IanMayo:
@helenayele - will you need a formal release to test this?

Aaah, @ayeshare - how about making a D-Lite release on your local PC, then zipping it up & copying into here, for Helen to test.

PR #4913, opened 2020-07-03
4908 unit test and more robust impl nisida

### LINKED ISSUE #4908: Adding more Unit Tests to Nisida Import and Creating it more robust to inconsistent input

saulhidalgoaular:
Fix #4908

### DIFF
--- a/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
+++ b/org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
@@ -176,9 +176,9 @@ public class ImportNisida {
 	}
 
 	public static boolean canLoadThisFile(final InputStream is) {
-		final BufferedReader br = new BufferedReader(new InputStreamReader(is));
-
 		try {
+			final BufferedReader br = new BufferedReader(new InputStreamReader(is));
+
 			String nisidaLine;
 			int lineNumber = 0;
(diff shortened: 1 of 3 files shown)


org.mwc.debrief.legacy/src/Debrief/ReaderWriter/Nisida/ImportNisida.java
```diff
@@ -351,14 +350,16 @@ protected static void processUnit(final String line, final NisidaLoadState statu
 		final String platformName = tokens[1];
 
 		// FIND THE PLATFORM.
-		TrackWrapper track = (TrackWrapper) status.getLayers().findLayer(platformName);
-		if (track == null) {
-			track = new TrackWrapper();
-			track.setColor(DebriefColors.RED);
-			track.setName(platformName);
-			status.getLayers().addThisLayer(track);
+		if (status.getLayers() != null) {
```
IanMayo: This class only makes sense if there is a layers object, so we shouldn't have to check for the presence of the layer object.

If this non-null check is because we pass in `null` for layers, then I think it would be better to fix the unit test, please.

saulhidalgoaular: Fixed at https://github.com/debrief/debrief/pull/4913/commits/027ed1e5f6c6d1bc73a19add6dd38bc89bf9d17d

PR #4916, opened 2020-07-10
set the preference store to trackshift plugin

### LINKED ISSUE #4904: Color change in Merge Tracks
"Do you know that when you Merge Track Segments – into new track (highlight infills) the track it produces is all the same colour."

I tried to reproduce this issue in `midflow2a.dpf` . When I merged the track legs into a new track, I got a plain black track.

screencast

Note - in the above screencast I had modified the leg colors before doing the merge.

So, it looks like the 'merge' tracks operation needs to be modified to re-use the colors in the original TMA legs.

ayeshare:
Fixes #4904 

The preference store used was CorePlugin's preference store and therefore this plugin was unable to start before Coreplugin.


org.mwc.debrief.track_shift/src/org/mwc/debrief/track_shift/TrackShiftActivator.java
```diff
@@ -61,6 +62,7 @@ public TrackShiftActivator() {
 	 */
 	@Override
 	public void start(final BundleContext context) throws Exception {
+		System.out.println("Started track shift");
```
IanMayo, line 65: @ayeshare - we don't need these debug lines, do we?

ayeshare: @IanMayo removed the debug lines

PR #4936, opened 2020-07-29
Fixed pasting track to a layer

### LINKED ISSUE #4929: Trouble saving data
* I have just opened Debrief and loaded the sample rep file BULK_BLUE_TRACKS
* In outline box I have selected the drop down menu and selected 'Create New Layer'.
* I named the layer 'Layer made in Outline'
* I selected the track 'T23_A' and cut it
* Then pasted into the layer
* I have saved the file as 'layer loss test.dpf'
* Following this I have the error: 'Exception Caught: Debrief Exporter not found for <SHIP NAME>
* The layer is now empty of any data.

ayeshare:
Fixes #4929 

Added Handler for TrackWrapper to be able to add trackwrapper to a layer

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/LayerManager/Swing/SwingLayerManager.java
@@ -1252,11 +1252,17 @@ public class SwingLayerManager extends SwingCustomEditor
 							otherLayer);
 					if (otherL != null) {
 						updateLayer(thisL, otherLayer, theTopLayer);
+						//it should not be a leaf
+						if(thisL.isLeaf()) {
+							thisL.add(otherL);
+						}
 						children.remove(otherL);
 					} else {
 						thisL.add(makeLayer(otherLayer, theTopLayer, treeCache));
 					}
-				} else {
+					
+				} 
+				else {
 					// hey, it's a leaf - just add it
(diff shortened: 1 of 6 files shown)


IanMayo:
@ayeshare - sorry, you've misinterpreted what is required: https://github.com/debrief/debrief/issues/4929#issuecomment-663082663

Your fix is quite valid, but for legacy reasons, Debrief _currently_ only works with Heavyweight tracks at the top level of the Outline View.  This is partly to speed up the task of providing the user with a list of currently loaded tracks. By only looking for them at the top level it avoids navigating down through the layer tree to find all tracks.

So, the fix we require at the moment is to prevent a HeavyWEight track from being pasted into a Layer (folder).

helenayele:
As per this fix, we have allowed to copy Heavyweight tracks inside a folder in full debrief.

IanMayo:
No, Helen, the fix is to block pasting a track into a layer.

helenayele:
> No, Helen, the fix is to block pasting a track into a layer.

But now we are able to paste Track:23_A in to our new layer like this 
[Image from Gyazo](https://gyazo.com/fbe8ded8794bf187a86156e27fc70e3e)
Is it okay?

ayeshare:
I will fix in a new PR

IanMayo:
I'm closing this PR, since we right now we don't want to allow pasting a heavyweight track into a layer.

PR #4939, opened 2020-07-30
block pasting a track to a layer

### LINKED ISSUE #4929: Trouble saving data

ayeshare:
Fixes #4929 

Fixed by preventing pasting track to layer


IanMayo:
@helenayele - while this is quite a simple fix, I've stepped in and refactored the code related to pasting items. So, it could do with some general testing around copying and pasting items and layers, including copy/paste of a layer into the top level (when there isn't a layer selected in the Outline)

So far these changes only relate to Full Debrief.

org.mwc.cmap.core/src/org/mwc/cmap/core/operations/RightClickPasteAdaptor.java
```diff
@@ -378,37 +378,40 @@ public static PasteItem createAction(final Editable destination, final Layers th
...
-					paster.setImageDescriptor(PlatformUI.getWorkbench().getSharedImages()
-							.getImageDescriptor(ISharedImages.IMG_TOOL_PASTE));
-					paster.setActionDefinitionId(ActionFactory.PASTE.getCommandId());
-				}
+			final boolean hasDest = destination != null;
+			final boolean contentsDynamic = clipboardContents[0] instanceof DynamicPlottable;
+			final boolean destinationDynamic = destination instanceof DynamicLayer;
+			// just check that the layers are compliant (that we're not trying to paste a
+			// dynamic plottable into a non-compliant layer)
+			if(hasDest && (!contentsDynamic || destinationDynamic)) {
```
IanMayo, line 402: @ayeshare - I swapped the logic around a bit here, related to dynamic plottables. I think you worked with dynamic plottables before.  Would you mind testing this still works properly please?

helenayele (commented):
1. In full debrief:
a. able to paste a layer on top label see this `copy of new layer` folder
b. able to paste a track inside a folder see `Track:T23_B` is copied inside `new layer' layer
[Image from Gyazo](https://gyazo.com/59265db27d01f2973b8e1bb73a0553c0)
c. Not possible to paste a layer inside another layer
2. In D-Lite
a. No option to copy  empty folder or a folder with shapes  at all
b. Able to copy a track folder but not possible to paste it anywhere
c. Not possible to copy a lightweight tracks at all
d. Not possible to copy dynamic shapes

Are all these expected ?

IanMayo:
Thanks for that, @helenayele 

## Full Debrief
We shouldn't be able to paste a Track into a Layer.  Maybe that's a flaw with my logic - I'm sure it _was_ working.

## D-Lite
Hmm, we should be able to copy lightweight tracks.  

I think it's a _technical debt_ that we have the two separate implementations.  I'll try to refactor the D-Lite version later today, so that it's used by both implementations.

helenayele (commented):
Now 
## Full Debrief
we are not able to copy tracks inside a folder
[Image from Gyazo](https://gyazo.com/27873732ec983ad7759cc2966746796e)

helenayele:
## Full Debrief
We can't copy a layer into a track - when we try to do this we get a message like this 
[Image from Gyazo](https://gyazo.com/dddf969c2f473fd00a5e595cc2b59eee)

or 
[Image from Gyazo](https://gyazo.com/0b07d2096968db020991c1d0c94214a2)
which is depend on the file being copied.
The first 2 dialogs are seen when we try to copy and paste `Special Layer' into Track:Style** layers
And the second dialog is seen when we try to copy and paste `TUA' folder into Tracks folder
As can be seen we get two dialogs at a time which is not user friendly. Instead we ca

helenayele (commented):
We still can't copy SensorArcs in D-Lite. But in full debrief we were able to copy and paste it before this fix.
[Image from Gyazo](https://gyazo.com/2b8ef164325c3543c634588f4f62c78e)

helenayele:
And
Not possible to copy and paste an empty layer into  a track - doing it will throw exception 
`Exception caught:failed to find track for sensor data: 0415` , but it should show valid message for a user

helenayele:
> And
> Not possible to copy and paste an empty layer into a track - doing it will throw exception
> `Exception caught:failed to find track for sensor data: 0415` , but it should show valid message for a user

While I try to copy and paste empty layer inside a track after launching full debrief nothing happens , meaning  don't get neither the error message(stack trace) nor get copied the empty folder inside a track

helenayele (changes requested):
@ayeshare  while I try  to copy a layer with lot of shapes from shape.rep file, it popups many dialogs:
[Image from Gyazo](https://gyazo.com/4cbee7ee220423f644a048a61f4aeb32)

Can we avoid this ? instead of displaying those many dialogs can we show a single dialog?
#### Step to reproduce:
1. Open shape.rep
2. Copy Buoys layer and paste it inside one of the Tracks e.x `Track: Nel Style
And You will get those bunch of dialogs

PR #4952, opened 2020-08-17
added buttons to import nisida and antares formats

### LINKED ISSUE #4950: Icons for Antares and Nisida files formats in D-Lite
In Debrief-Lite we have buttons to open the File Dialog for the assorted file types.

Helen has pointed out that we don't have icons for Antares or Nisida.  We should.

I've added icons to Debrief-Lite at 16, 24 and 48 pixels.

ayeshare:
Fixes #4950


helenayele (approved):
The actions for Nisidia and Antares file types looks like this
[Image from Gyazo](https://gyazo.com/07dd6606dc0cea73bbbdc91bce07debf)
And working as expected.

So approving...

PR #4956, opened 2020-08-27
bug fixed due using wrong variable name

IanMayo:


IanMayo:
@ayeshare  reported this:
image

helenayele (approved):
Now the last pasted item is not re-pasted again after redo action.
[Image from Gyazo](https://gyazo.com/b6d986c93afda020d7bf56f88d937925)

PR #4962, opened 2020-09-03
cleared reorder layer default configuration

### LINKED ISSUE #4957: Block being able to re-arrange Narrative columns
It's currently possible to re-order the columns in Full Debrief by dragging.

But, this is likely to cause confusion.  It may also break the hide/show column functionality.

We should block the ability to adjust Narrative Viewer columns.

ayeshare:
Fixes #4957


helenayele (approved):
No more possible to re-arrange the order of columns in narrative viewer.
[Image from Gyazo](https://gyazo.com/a62ef0f6f02dd80eb0a3791a0a13900d)
So approving...

PR #4972, opened 2020-09-18
Fixed path issue

### LINKED ISSUE #4964: Path of GeoTiff in DPF files not portable
We can include GeoTiff images in our plots - analysts use this a lot.

Debrief stores the absolute path to the image in the `layerPath` attribute.  This is ok for multiple users when the images comes from a shared network folder, since the path probably matches.  But, it fails if the dpf plus GeoTiff are zipped up and sent to someone else.  

Another example of this is for the `sample_data` that we distribute with Debrief.  A new user can load a `.dpf` file which includes a GeoTiff file in the current folder, but Debrief will fail to load it.

How do we handle this?  

1. When saving an `externalDataFile` object, see if it's in the same folder as the `.dpf` file.  If it is, just use a relative reference
2. When loading an `externalDataFile` object, if the file-load fails, have another go, using a relative path reference.
3. If the above fails, get the absolute reference to the `.dpf` file and see if the image file is present in the same directory.  If it is, load it.

Hmm, but the issue may be that we don't know the default folder location that Debrief is using as `home directory`.

We should investigate this further, to see what the options are.

**Note**: a recent PR introduces a GeoTiff image file, and a `.dpf` file that refers to it.  The `.dpf` needs to be manually _fixed_ to work on each new PC it's stored on, because of the above issue:
https://github.com/debrief/debrief/pull/4963

ayeshare:
Fixes #4964

### DIFF
--- a/org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/editors/CorePlotEditor.java
+++ b/org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/editors/CorePlotEditor.java
@@ -123,9 +123,6 @@ public abstract class CorePlotEditor extends EditorPart implements IResourceProv
 		ISelectionProvider, IPlotGUI, IChartBasedEditor, IUndoable, CanvasType.ScreenUpdateProvider, INameablePart {
 
 	private static final String CONTEXT_ID = "org.mwc.cmap.plotEditorContext";
-	// //////////////////////////////
-	// member data
-	// //////////////////////////////
 
 	/**
 	 * the chart we store/manager
@@ -156,26 +153,16 @@ public abstract class CorePlotEditor extends EditorPart implements IResourceProv
 	 */
 	protected PlainProjection _pendingProjection;
 
-	// drag-drop bits
-
 	protected DropTarget target;
 
 	private Vector<ISelectionChangedListener> _selectionListeners;
 
-	// private ISelectionChangedListener _selectionChangeListener;
-
 	private ISelection _currentSelection;
-
-	// private CanvasType.PaintListener _selectionPainter;
 
 	/**
 	 * keep track of whether the current plot is dirty...
 	 */
 	protected boolean _plotIsDirty = false;
-
-	// ///////////////////////////////////////////////
-	// dummy bits applicable for our dummy interface
-	// ///////////////////////////////////////////////
 
 	protected DataListener2 _listenForMods;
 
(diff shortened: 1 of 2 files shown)


helenayele (commented):
I just copied `plotWithGeoTiff.dpf` file in sample_data folder and also in `sample_data /newfolder` folder and found this error dialog while opening it.
Error_loading_file

IanMayo:
Ok, I think we need @ayeshare  to reproduce your example.  Hmm, _maybe_ the paths get mangled in a special way when Debrief is being run from Eclipse.

ayeshare:
> Ok, I think we need @ayeshare to reproduce your example. Hmm, _maybe_ the paths get mangled in a special way when Debrief is being run from Eclipse.

Yes @helenayele I need an example file to test.  The path of the tif file should be relative to the folder in which the dpf file is.
For example, if you have a dpf file in folder folderA and the tif file is in subFolderA under folderA, then the layerPath variable should be subFolderA/some.tif.  If the dpf file is in folder folderA and the tif file is in folderA, the layerPath variable should be some.tif

helenayele:
Okay here is the file named `PlotWithGeoTiff.dpf` and there is a tif file called `SP27GTIF.tif` in the same folder
[PlotWithGeoTiff.zip](https://github.com/debrief/debrief/files/5268098/PlotWithGeoTiff.zip)

org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/editors/CorePlotEditor.java
```diff
@@ -359,7 +363,8 @@ public void addThisLayer(final Layer theLayer) {
 				if (theLayer instanceof ExternallyManagedDataLayer) {
 					final ExternallyManagedDataLayer dl = (ExternallyManagedDataLayer) theLayer;
 					if (dl.getDataType().equals(MWC.GUI.Shapes.ChartBoundsWrapper.WORLDIMAGE_TYPE)) {
-						final GeoToolsLayer gt = new WorldImageLayer(dl.getName(), dl.getFilename());
+						
+						final GeoToolsLayer gt = new WorldImageLayer(dl.getName(), getFixedFilePath(dl.getFilename()));
```
IanMayo, line 367: @ayeshare  - your new method can return null (if the file can't be found).  I don't think we should create a `WorldImageLayer` if the file wasn't found.

org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/editors/CorePlotEditor.java
```diff
@@ -532,6 +537,40 @@ public void propertyChange(final PropertyChangeEvent arg0) {
 		// };
 	}
 
+	protected String getFixedFilePath(final String fileName) {
+		String tiffFilePath = null;
```
IanMayo, line 541: It would be better engineering to make this a `final String tiffFilePath` - then the compiler will help you in verifying that it gets assigned to a value (or null) in each logical branch.

helenayele:
Using relative path following this structure:
 sample_data->plotwithgeotiff.dpf
sample_data/sample_data/SP27GTIF.tif
The file now loads the tif file:
[Image from Gyazo](https://gyazo.com/351fffcd1d15fccd2e6addc627844aa0)
need to check  the absolute path though.

helenayele:
Setting the layer path with absolute path of the tiff file is also working.

org.mwc.cmap.plotViewer/src/org/mwc/cmap/plotViewer/editors/CorePlotEditor.java
```diff
@@ -553,14 +553,17 @@ protected String getFixedFilePath(final String fileName) {
 			}
 			else if(input instanceof FileStoreEditorInput) {
 				//if the file is dragged from outside workspace, get the location of the plot file 
+				String tmpFilePath;
 				try {
 					File localFile = new File(((FileStoreEditorInput)input).getURI().toURL().getFile());
-					if(localFile!=null) {
```
IanMayo: `localFile` can't be null - it must contain a File object (or exception will have been thrown)

IanMayo:
@ayeshare - there's still another comment to handle: https://github.com/debrief/debrief/pull/4972/files#r493629854

IanMayo (approved):
Tested, and it's working fine with sample plot file in `sample_data`

IanMayo (changes requested):
It appears that @helenayele is getting an error on this (as reported in Slack)

ayeshare:
> It appears that @helenayele is getting an error on this (as reported in Slack)
@helenayele 
@IanMayo She had some difficulties in setting the absolute and relative path.  After setting the path correctly she has not found any new issues

IanMayo (approved):
Reported working by Ayesha: https://github.com/debrief/debrief/pull/4972#issuecomment-698824459

PR #4978, opened 2020-09-26
4973 structure change compose file

### LINKED ISSUE #4973: GeoPDF - Change structure of the Compose file
Currently the compose file is created using Line, Labels, Markers and first point structure. I propose changing this structure to render FixWrappers independently. It will give the flexibility of changing the color/shape to each point individually, to represent a final result more similar to the original Debrief file.

### LINKED ISSUE #4977: Export to GeoPDF - Phase 5 tracks with multiple colors
It is possible to shade a Debrief track in different colors.  

The colors may be used to indicate when one platform is in contact with another, or to separate times when a platform is going above 10 knots.

It would be useful if these colors could be presented in the GeoPDF export

Note: this change may require the change in structure defined in #4973

saulhidalgoaular:
Fix #4973 
Fix #4977


IanMayo:
@saulhidalgoaular  - I've just done an export of `PlotWithGeoTiff.dpf` from the `sample_data` folder. I had to delete & reload the TIFF, but that's being fixed elsewhere.

The time controller is configured to have one minute time steps.  But, the export is at 10 minute time steps - so it's quite _Clunky_.

So, the time steps in the GeoPDF need to match the Time Controller, please.

IanMayo:
> @saulhidalgoaular - I've just done an export of `PlotWithGeoTiff.dpf` from the `sample_data` folder. I had to delete & reload the TIFF, but that's being fixed elsewhere.
> 
> The time controller is configured to have one minute time steps. But, the export is at 10 minute time steps - so it's quite _Clunky_.
> 
> So, the time steps in the GeoPDF need to match the Time Controller, please.

@saulhidalgoaular  - I think I've fixed the above, we should use small time step, not large time step.

IanMayo (approved):
Working - let's get it into a release.

PR #4982, opened 2020-09-30
4966 multiple tiff geopdf

### LINKED ISSUE #4966: Export to GeoPDF - Phase 3: Multiple TIFF images
It is most likely that the analyst will just be displaying a single TIFF images.

But:
1. Other TIFFs may be visible outside the current viewport
2. The analyst may wish to display the content of multiple TIFFs in the export GeoPDF.

For 1) we can loop through all visible GeoTiffs, and discount any that don't overlap the current viewport.  

But, if multiple are present, then under 2) we should pass them all to the `GDAL_Warp` command.

saulhidalgoaular:
Fix #4966 
Not finished, I still need to discard the tif out of the viewport


IanMayo (approved):
Multiple export working fine.

PR #4985, opened 2020-10-01
4984 Removing any special character from the geojson path

### LINKED ISSUE #4984: GeoPDF - Check Special characters in the track names while generating the Geojson
Analysts had trouble yesterday. couldn't find geojson file. it was for track titled F/V SOMETHING.
Debrief interpreted the / as a folder separator.  We need to make the track name "safe" for use in a file path

### LINKED ISSUE #4986: Geo-PDF Warn users about performance issue when the track has more than 100 steps.

saulhidalgoaular:
Fix #4984 
Fix #4986


saulhidalgoaular:
It can be tested with the following file.
[boat1 - Copy.zip](https://github.com/debrief/debrief/files/5310650/boat1.-.Copy.zip)

IanMayo (approved):
Working fine, pushing release, so we can get feedback

PR #4990, opened 2020-10-01
4988 Using milliseconds for the temporal file name generated

### LINKED ISSUE #4988: GeoPDF export. Issue with duplicated timestamp
I'm playing with exporting to GeoPDF.  I've configured a scenario that breaks the export.

It's a strange scenario, but it gives us an opportunity to make things more robust.

Steps to reproduce:
* open `PlotWithGeoTiff.dpf`
* resample `COLLINGWOOD` to 30 second positions
* Export GeoPDF

I think the export fails because multiple data-files have the same name. This is because of the date format in use.

I _think_ that if the date format is modified to include seconds, the export works.

IMHO the issue could be fixed by using the DTG (long millis) in the filename, rather than the formatted string.

saulhidalgoaular:
Fix #4988


IanMayo (approved):
Seen working.

PR #4994, opened 2020-10-01
4992 GeoPDF - Using Speed Interval from Time Controller 

### LINKED ISSUE #4992: GeoPDF Export - step forward using Time Controller interval
When the user `Plays` a GeoPDF we control the time interval between each step forward (probably a second).

We should instead take this value from the TimeController `Step Interval` property.

Then `Play` can be faster or slower than one second.

saulhidalgoaular:
Fix #4992

### DIFF
--- a/org.mwc.debrief.core/src/org/mwc/debrief/core/actions/ExportAsGeoPDFHandler.java
+++ b/org.mwc.debrief.core/src/org/mwc/debrief/core/actions/ExportAsGeoPDFHandler.java
@@ -70,6 +70,7 @@ public class ExportAsGeoPDFHandler extends CoreEditorAction {
 				if (pref != null) {
 					configuration.setStepDeltaMilliSeconds(pref.getSmallStep().getMillis());
 					configuration.setDateFormat(pref.getDTGFormat());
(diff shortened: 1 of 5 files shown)


helenayele:
After I export "UKPlotWithGeoTiff.dpf" to pdf using File > Export into GeoPdf action, the pdf file cannot be opened and got the following messages while trying to open it in Adobe Reader and Edge browser.
[Image from Gyazo](https://gyazo.com/5ad2c16f0c27f4eea0cd719f95647639)
[Image from Gyazo](https://gyazo.com/c690001d90fc282e1174e07f72f727ea)
[Image from Gyazo](https://gyazo.com/4d2c02e01d3a643dc523111410eb3674)
Is there something I miss?

IanMayo:
@saulhidalgoaular - I also get this error when exporting the UK plot.

The export works fine for `PlotWithGeoTiff.dpf`.

I'm interested in what's going wrong.

IanMayo (approved):
Tested, working well.

Witness `Play` running at 1 sec, 5 sec, and 1/3 sec intervals.

PR #4995, opened 2020-10-01
4991 Put all the temporary files inside a folder together

### LINKED ISSUE #4991: GeoPDF Export - put all files for one export into single folder
Currently in the export we don't delete the files used in the export.

But, they all sit in the top-level of `temp`.

I think it would be better if we started by creating a unique folder in `temp`, then put our working files into there.  That will make it easier for us (or analysts) to package up a set of data-files for subsequent analysis.

saulhidalgoaular:
Fix #4991


IanMayo (approved):
Tested. Working fine.

PR #5005, opened 2020-10-09
4992 check fixwrapper visible

### LINKED ISSUE #4992: GeoPDF Export - step forward using Time Controller interval

### LINKED ISSUE #4996: GeoPDF - Don't include elements that are not in the Time Range
Elements that are not visible in the viewport shouldn't be included as a layer (background or other layers).

Also in the Time Period we should use the union of the track time periods, but only if they are visible, intercepting it with the Time Filter of the Time Controller.

saulhidalgoaular:
Fix #4992 
Fix #4996


IanMayo:
@helenayele  - I think you're aware of both of the issues resolved here.  Would you mind testing please?

helenayele (changes requested):
1. In this fix, we only export the visible part of the track
[Image from Gyazo](https://gyazo.com/1edd37d99d4a0aac54d1cee3b2387707)
2. But the end time on exported pdf and on time controller is not the same.
See the end time on time controller: it is 5:26
[Image from Gyazo](https://gyazo.com/cfa94e2edbf1a326fa61571e6e419a8e)

The end time on exported pdf looks is 5:20
[Image from Gyazo](https://gyazo.com/a919dbe6906d2f7049b8ff6bf15a2b2b)

can you please fix this?

saulhidalgoaular:
Good morning. Could you specify the steps to generate that PDF?
Somehow I tried the same, but it works as expected in my computer.

image

helenayele:
### Steps to reproduce ;

1. Open the PlotWithGeoTiff.dpf file
2. Change small step to 5 minutes
3.Export to pdf file

PR #5006, opened 2020-10-09
4993 - GeoPDF - Check if the label name exists before creating it.

### LINKED ISSUE #4993: GeoPDF Export - failing with time steps under one minute
I have `PlotWithGeoTiff.dpf` open, in the `origin/4988_renaming_geojson_temporal_files` branch.

I want to experiment with time steps shorter than one minute.  This track only has positions at one minute, so that probably won't work.  So, I resampled both tracks to 30 second intervals (to give tracks with 47 & 53 positions).

I did export to GeoPDF, all worked fine.

Then I switched small step to 30 seconds, and exported again.  The export completed, but the file wouldn't open.  I didn't see any GDAL errors.

I switched back to small step of one minute, and it worked ok.

This may be easier to explore if we package up the source files into one folder first, as requested in #4991

saulhidalgoaular:
Fix #4993


saulhidalgoaular:
Good afternoon @IanMayo .

Origin of the problem was that multiple layers had the same name. I am now checking if the name was previously used, and if it was, I add a suffix. Example:

"95/12/12 05:17"
"95/12/12 05:17_2"
"95/12/12 05:17_3"
....

It would produce a PDF like the attached.

[156.pdf](https://github.com/debrief/debrief/files/5357128/156.pdf)

If you prefer a different suffix/alternative, I am ready to implement it.

IanMayo:
Hello @helenayele  - another one to test, please. Hopefully there's enough detail in the issue for you to reproduce it.

helenayele (changes requested):
Debrief goes to not responding state while generating pdf following these steps:

1. Open plotwithgeotiff.dpf file
2. change small steps and step interval  to milliseconds as follows
[Image from Gyazo](https://gyazo.com/932a0a4af822e52e64465df063edc10b)
3. it goes to not responding state
[Image from Gyazo](https://gyazo.com/c469d6cd4122e06b42d20289d91c272b)

helenayele:
But if I give 30seconds it generates the pdf file

IanMayo:
> step interval to milliseconds as follows

That's not going to happen :-)

But, maybe @saulhidalgoaular  should include test of time period vs step size, to ensure we aren't going to try to create something with over 200 points.

Hmm, maybe we do something like that already.

saulhidalgoaular:
Good morning.

I have tested what @helenayele  mention, and yes, I receive the message of performance problems, because it is over 100 steps.

image

@IanMayo . Initially we had agreed to show the message when the result PDF had more than 200 steps, but in some cases 200 was still too slow and we started showing the warning message from 100 steps.

helenayele:
Okay so what is the minimum values of the small step and step interval it is expected to work?

saulhidalgoaular:
Hello @helenayele .

There is no minimum small step interval.

PDF generates a list of "steps" for each tick of the clock. Idea is not to have more than 100 steps. So, the formula should be like 
ideally timePeriod / stepInterval < 100. Step Interval can be small, but you will need a small timePeriod too then.

helenayele:
Okay @saulhidalgoaular , with the below time steps values
[Image from Gyazo](https://gyazo.com/4a816f2c5455f0be01ddb35a67e4de26)
I can generate the pdf, but while opening it, it goes to not responding state. And after waiting for 2 - 3 minutes the pdf file is opened. But I can't play the generated pdf , it just goes to not responding state.
[Image from Gyazo](https://gyazo.com/291985ad5c3b4abc96a5837e59e30803)

IanMayo:
I think issue #5043 relates to the issue Helen had above.

helenayele (approved):
Okay, so may be the last issue I have reported will not be related to the main issue this PR is fixing. So I will create new issue for it #5045.
So I will approve this PR.

PR #5010, opened 2020-10-28
5008 pepys duplicate names

### LINKED ISSUE #5008: Pepys Import - duplicate names
> The Cobham jets parser had an issue where two jets had the same name but different tail numbers. Debrief importer couldn't differentiate between the two and shaded the screen.

I guess we need one of two strategies:
* when importing tracks, use platform name and identifier (pennant) to generate the track name
* when importing tracks, loop through the platforms. If there are any duplicate names (but different GUID) append the identifier.

saulhidalgoaular:
Fix #5008

### DIFF
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Comment.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Comment.java
@@ -71,7 +71,7 @@ public class Comment implements AbstractBean, TreeStructurable {
 			_layers.addThisLayer(dest);
 		}
 
-		final NarrativeEntry entry = new NarrativeEntry(getPlatform().getName(), new HiResDate(getTime()),
+		final NarrativeEntry entry = new NarrativeEntry(getPlatform().getTrackName(), new HiResDate(getTime()),
 				getContent());
 
 		// ok, can we provide a track color for it?
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Contact.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Contact.java
@@ -87,7 +87,7 @@ public class Contact implements AbstractBean, TreeStructurable {
 
 		if (target != null && target instanceof BaseLayer) {
 			final BaseLayer folder = (BaseLayer) target;
-			final Editable found = folder.find(getPlatform().getName());
+			final Editable found = folder.find(getPlatform().getTrackName());
 			if (found != null && found instanceof TrackWrapper) {
 				final TrackWrapper track = (TrackWrapper) found;
 
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Datafile.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/Datafile.java
@@ -70,10 +70,6 @@ public class Datafile implements AbstractBean {
 					assertTrue("Datafiles - Reference Name", exist);
 				}
 
-				for (final Datafile datafile : list) {
-					assertTrue("Datafiles - Concatenated reference ", datafile.getDatafile().getName()
-							.equals(datafile.getReference().substring(datafile.getReference().indexOf('.'))));
-				}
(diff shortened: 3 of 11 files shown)


IanMayo:
@saulhidalgoaular  = do we really need to add the two `test` databases to the repo?

Were you able to reproduce the error, and thus be confident the issue is fixed?  I guess it should be able to "mangle" the data in `DB Browser for SQLite".  I _presume_ we've got `Collingwood` and `Nelson` in there.  The `Platforms` table entries could be modified so they're both `Collingwood`, but to provide different identifiers.  Then, just load the `Collingwood` tracks, and verify they aren't being handled as one.

I hope that all makes sense :-)

Oh, if you are able to reproduce the above, then a screenshot of the data in Debrief would be useful.

IanMayo (changes requested):
Changes requested in earlier comment.

helenayele:
Hey @saulhidalgoaular  , would you please attach the database to test this PR ?

saulhidalgoaular:
Good afternoon.

I am attaching a sample  very much like you have specified Ian. I didn't find Collingwood in the tracks, but I did the same with DOLPHIN_FR.

I have modified DOLPHIN_GB to DOLPHIN_FR, and the identifier to 124, so we will have 2 Platforms with the same name, but different identifier.

image

After importing, it is properly identified as a different Track.

image

I am uploading here the database with the corrupted data sample to be able to reproduce the steps easier.

[pepys.sqlite.tar.gz](https://github.com/debrief/debrief/files/5506354/pepys.sqlite.tar.gz)
Steps:
1. Load the database attached.
2. Select in the Database Import (Pepys module), the DOLPHIN_FR database.
3. Press Import.
4. Close the Pepys Module.
5. Confirm that there are 2 different tracks.

Helen, Please, follow the screenshots I have attached to reproduce the steps easier.

helenayele (approved):
Thanks @saulhidalgoaular  for the detailed steps.

Following the steps I get two tracks imported with different names as follows:
[Image from Gyazo](https://gyazo.com/55c50c21c7bba6fd6f5b58249de5309b)

and before the fix, it looks like:
[Image from Gyazo](https://gyazo.com/ea12a5c52799aa66cf7bd94b2db0152b)

So approving...

IanMayo (changes requested):
Thanks @saulhidalgoaular 

From looking at the screenshots, the `Pepys Import Dialog` has a single entry for `Dolphin FR`.  I guess we need to include the identifier in the query and the listing - so the analysts see the presence of both platforms,

saulhidalgoaular:
> Thanks @saulhidalgoaular
> 
> From looking at the screenshots, the `Pepys Import Dialog` has a single entry for `Dolphin FR`. I guess we need to include the identifier in the query and the listing - so the analysts see the presence of both platforms,

oh, I understand. Ok, Sure. Changing.

saulhidalgoaular:
Good afternoon. I have added the identifier to the platform name.

image

helenayele (commented):
the identifiers are shown like this
[Image from Gyazo](https://gyazo.com/70728d6ea00c180214bc545babbbfc44)
But we can't query the database by these identifiers
[Image from Gyazo](https://gyazo.com/70728d6ea00c180214bc545babbbfc44)

saulhidalgoaular:
Hello @helenayele 

Well spotted. I have added it as filterable.

helenayele:
Thanks @saulhidalgoaular , I can now filter with the identifier like this
[Image from Gyazo](https://gyazo.com/6d67d44f2a024083b54e14fbbf4ceeee)

But if I concat the state name like this, it doesn't filter 
[Image from Gyazo](https://gyazo.com/122a43461fe36f9a974073e56793b61f)

saulhidalgoaular:
Oh, I see. The thing is that we are mixing different concepts. That box has been created to search in the database query, filtering the results according to the text there. However, there is no single database field with the name and identifier together. 

It could be implemented, but maybe it is not trivial. I need to specify (in the mini orm) a way to concatenate fields to create a new field as "filterable".

I don't know. I could implement it if it is really needed. What do you think @IanMayo ?

IanMayo:
No, we can live without it, @saulhidalgoaular .  It's something that is quite unlikely to happen.

helenayele (approved):
Okay @saulhidalgoaular , Approved.

IanMayo (approved):

PR #5033, opened 2020-11-26
5026 Clear Area button in Pepys

### LINKED ISSUE #5026: Need to be able to clear spatial filter in `Pepys Import` dialog
We can populate the spatial filter by typing in values, or by using the current viewport.

But, it's cumbersome to clear the filter.

image

We should add a "Clear" button at the bottom left of the `Area` panel.

saulhidalgoaular:
Fix #5026


helenayele (dismissed):
Clear action is added and works well.
[Image from Gyazo](https://gyazo.com/f680196bedb6ccb6d37a8f04680e1d4a)
Approving...

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/view/PepysImportView.java
```diff
@@ -68,6 +68,7 @@
 	private Button importButton;
 	private Button testConnectionButton;
 	private Button useCurrentViewportButton;
+	private Button clearAreaButton;
```
IanMayo, line 71: @saulhidalgoaular  - minor question.  This really isn't important, I'm just "thinking out loud".

It caught my eye that these fields weren't final, because  I'm sure the GUI elements won't get deleted/edited.

I looked at the code.  They can't be final because they're not created in the constructor.  But, they are created in a method that only gets called from the constructor.

So, if the UI was actually built in the constructor they could be final.

What do you think about the pros/cons of doing that?

saulhidalgoaular: Good afternoon @IanMayo 

Very good observation. I consider you are correct, and controls should always be final. In other case, we are in risk of losing the bindings created by the controller.

I will change it and I would also remove access from the view to the model (removing the model from the view constructor), because it is conceptually not very precise.

helenayele:
@saulhidalgoaular , one minor change, can we put Clear button after Use Viewport ? Because usually users use Clear button after using Use Viewport button.

saulhidalgoaular:
> @saulhidalgoaular , one minor change, can we put Clear button after Use Viewport ? Because usually users use Clear button after using Use Viewport button.

Good morning @helenayele .
Yes, I could move to the clear button to the right, but the thing is that it would contradict the original requirements:
https://github.com/debrief/debrief/issues/5026#issue-751597549
> We should add a "Clear" button at the bottom left of the Area panel.

IanMayo:
>  contradict the original requirements:

I don't mind if it moves :-D

saulhidalgoaular:
image

helenayele (approved):
Now clear buttons moves to the right of the Viewport button.
[Image from Gyazo](https://gyazo.com/7879b4f522819e4b1cba44b43ef86c8f)

approving..

PR #5049, opened 2021-02-08
5025 custom query orm

### LINKED ISSUE #5025: Pepys Database Import crashing on `Apply`
Users are connecting to a PostGIS database.

They apply a time filter, and press `Apply`.  Debrief hangs.  Nothing gets written to the Error Log.

We should put some (more) error trapping around the `Apply` button operation.

1. The problem _could_ be the volume of data. 
2. Or, the problem _could_ be something specific to the currently loaded data.

The database contains around 150 datafiles.  To help with option 1) we should create/load - say, 200 datafiles to PostGIS, and then connect to it. (Ian can do that).

For option 2) we should inspect the code,  include some discrete logging statements, and check we have exception handling.

saulhidalgoaular:
Fix #5025

### DIFF
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/ModelConfiguration.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/ModelConfiguration.java
@@ -135,7 +135,11 @@ public class ModelConfiguration implements AbstractConfiguration {
 	@Override
 	public void apply() throws Exception {
 		validate();
-		currentItems = TreeUtils.buildStructure(this);
+		if (ALGORITHM_TYPE.FAST_MODE.equals(getAlgorithmType())) {
+			currentItems = TreeUtils.buildStructureFastMode(this);
+		}else if (ALGORITHM_TYPE.LEGACY.equals(getAlgorithmType())) {
+			currentItems = TreeUtils.buildStructure(this);
+		}
 		updateTree();
 		setSearch("");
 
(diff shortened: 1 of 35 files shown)


org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/ModelConfiguration.java
```diff
@@ -135,7 +135,11 @@ public void addPropertyChangeListener(final PropertyChangeListener l) {
 	@Override
 	public void apply() throws Exception {
 		validate();
-		currentItems = TreeUtils.buildStructure(this);
+		if (ALGORITHM_TYPE.FAST_MODE.equals(getAlgorithmType())) {
```
IanMayo, line 138: Hello @saulhidalgoaular - I think a `switch` statement would be tidier here - it will check we handle all the possibilities. Right?

IanMayo:
The maximum number of positions/data-points that Debrief can handle is probably around 200,000.

It would be sensible to have a page-size of 10,000 - in order to have a responsive UI, or even to display a progress marker.

Hmm, so we probably should do a `count()` query before the import, so we can then know how far through the progress bar we are.  I'll ask @arunlal-v about that.

Screen Shot 2021-02-26 at 13 49 44

IanMayo:
We aren't loading `Contact` items.

I've had a look at the logic.   The importer is looking for the layer with the datafile name:
image

But, tracks aren't put into folders with the datafile name - they're going directly in at the top level:
image

I've modified the logic to look for the track (instead of the datafile):
image

But, it's failing because the Sensor doesn't know about the parent platform:
image

(I'll check with the users about whether they want the tracks put into data-file named folders.)

IanMayo:
Ok. The ORM isn't setting the datafile reference. I've tried to get into it, but it's going to take ages.

I think we're better off if you sort it :-)

IanMayo (approved):
Tested & working on MS-Windows

PR #5052, opened 2021-03-09
added a minor change to remove restriction to build only tagged release

### LINKED ISSUE #5050: Create deployable instance in PR builds
In another project I maintain, a deployable instances gets created as an artefact:
image

I believe this is performed by this code:
image

So, if a build fails to pass the test for "Release", it would be good if we could create it as an artefact.  Well, ideally we would make all of the files we normally release into these artefacts.

ayeshare:
Fixes #5050


ayeshare:
@helenayele The artifacts will be found after clicking the link in github actions tab.

IanMayo:
Thanks @ayeshare 

The artefact is much larger than normal Debrief releases:
image

Could we name the artefact with `Debrief_Win64_date_branch, like this please:
```
Debrief_Win64_20210311_ 5050_downloadablerelease_allbuilds
```

IanMayo:
Aah, I see why the downloadable distribution is larger.

For this we just need the 500Mb `DebriefNG_Windows64Bit.zip`

image

IanMayo:
> Aah, I see why the downloadable distribution is larger.
> 
> For this we just need the 500Mb `DebriefNG_Windows64Bit.zip`
> 
> image
> 
> Instead of a zip of the 3 windows deliverables, could we just have the one DebriefNG Zip, renamed according to the _new_ naming convention please?

@ayeshare - I think I've fixed the above issue. Checkout my change here: https://github.com/debrief/debrief/pull/5052/commits/89e3529f87259d604ba790e230af3ce34e71d01d

IanMayo:
@ayeshare - one last issue. It doesn't look like the correct branch name is being used.  It appears to have the PR number (5052 instead of 5050) with the word "merge".

IanMayo:
@ayeshare - it looks like we aren't getting the branch in the filename.
image

PR #5057, opened 2021-03-24
5053 speed units

### LINKED ISSUE #5053: GPX course being read in as radians
The GPX importer is mistakenly using Radians for course.  The course data is actually provided in Degrees.

It has also become apparent that the values for speed (in m/sec) are being treated as yards/sec.

To store this data,  the data is converted as follows:
```
trackPoint.setSpeed(MWC.Algorithms.Conversions.Mps2Kts(speedMS/3));
```

IanMayo:
Fixes #5053


IanMayo (commented):
Validated through inspection.

PR #5083, opened 2021-04-29
5079 Count of the Import next to the "Import" button

### LINKED ISSUE #5079: Pepys Import - show count of selected items
Within (or next to) the `Import` button it would be useful to display the number of selected records.

Then analysts can develop best practices such as only performing import if there are less than 200,000 items selected.

If this affects the size of the `import` button we should be prepared to display up to 7 digits.

saulhidalgoaular:
Fix #5079


IanMayo:
Very good, thanks @saulhidalgoaular 

ImportWarning

IanMayo (approved):
Nice work Saul

PR #5085, opened 2021-05-03
5078 pepys using store procedures

### LINKED ISSUE #5078: Move to PostGres stored functions
A set of functions have been created in the `tracstor` postgres database:
image

These are quite similar to the existing ones in `pepys\xxx.sql`

Using these functions brings these benefits:
* When a query is run, PostGres has to take some time to plan how to most effectively run the query.  For a stored function, it just does this the first time - it runs straightaway next time
* We can change the structure/strategy of the query without requiring a Debrief change
* We have opportunities to control access to the database, by hiding all tables, and letting some users see these functions

saulhidalgoaular:
Fix #5078


saulhidalgoaular:
In the state query, the result doesn't contain the sourceId that indicates to which reference this state belongs.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/ModelConfiguration.java
```diff
@@ -314,13 +319,33 @@ public ALGORITHM_TYPE getAlgorithmType() {
...
+				return ALGORITHM_TYPE.FAST_MODE_STORED_PROC;
 			}
 		}
 		// LEGACY AS DEFAULT.
 		return ALGORITHM_TYPE.LEGACY;
 	}
 
+	@Override
+	public String getCommentQuery(final ALGORITHM_TYPE algorithType) {
+		if (getAlgorithmType() == ALGORITHM_TYPE.FAST_MODE) {
```
IanMayo, line 331: @saulhidalgoaular - we're receiving the algorithm type as a parameter, but also retrieving it here.

I guess we should either delete the parameter, or use it instead of `getAlgorithmType()`

IanMayo: I think a Switch statement would give better type checking in these :
```
	@Override
	public String getCommentQuery(final QUERY_STRATEGY algorithmType) {
		switch (algorithmType) {
		case FAST_MODE:
			return "/comments.sql";
		case FAST_MODE_STORED_PROC:
			return "/commentsProc.sql";
		default:
			Application.logError2(ToolParent.ERROR, "Unexpected query strategy" + algorithmType, null);
			return null;
		}
	}

```

IanMayo (changes requested):
Hello @saulhidalgoaular - thanks for this.

I've added a request to the original issue, to allow users to switch `call stored functions` on and off.

I've also added a coding suggestion here in the PR, which shortens the code and introduces a bit more type-safety.

org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/ModelConfiguration.java
saulhidalgoaular: > @saulhidalgoaular - we're receiving the algorithm type as a parameter, but also retrieving it here.
> 
> I guess we should either delete the parameter, or use it instead of `getAlgorithmType()`

oh, you are absolutely right. I have it twice.

IanMayo:
Have _Strange_ issue.  When I ask to use stored functions, it says the file path is invalid.  I know the file path isn't invalid.
screencast

saulhidalgoaular:
> Have _Strange_ issue @saulhidalgoaular . When I ask to use stored functions, it says the file path is invalid. I know the file path isn't invalid.
> screencast

Hello @IanMayo .

There is a bug related with the path for the master pptx template. You can skip that error selecting again the template. I will create a ticket for it

IanMayo (approved):
Great. thanks.

PR #5097, opened 2021-05-13
5064 Enabling transparency for snail draw

### LINKED ISSUE #5064: Avoid snail trail fading to when white not on white background
The current snail trail fades to the background colour of the plot.
image

This looks unusual when rendered over a blue ocean, like we have with the Natural Earth backdrop.
image

It would be better if:
a) fade to light blue when Natural Earth is loaded
b) fade to transparent

Option b) would be a better solution, since it would work for a range of backdrop types.

The fix to this issue is probably in `SnailDrawTrack.PhasedColorDef`, and involves increasing the alpha on the color instead of making the color a steadily closer to the background color.

Note: this workaround for the issue has been passed to analysts: https://www.youtube.com/watch?v=yxx19gwoGfY

saulhidalgoaular:
Fix #5064

### DIFF
--- a/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
+++ b/org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
@@ -1218,6 +1218,8 @@ public class SWTCanvasAdapter implements CanvasType, Serializable, Editable, Ext
 
 		if (theCol != _currentColor) {
 			_currentColor = theCol;
+			// Let's save the alpha in case the color has.
+			_theDest.setAlpha(theCol.getAlpha());
 
 			// transfer the color
 			final Color swtCol = ColorHelper.getColor(theCol);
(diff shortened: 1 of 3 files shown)


org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
```diff
@@ -1218,6 +1218,8 @@ public void setColor(final java.awt.Color theCol) {
 
 		if (theCol != _currentColor) {
 			_currentColor = theCol;
+			// Let's save the alpha in case the color has.
```
IanMayo, line 1221: Do you need to finish off the above sentence @saulhidalgoaular ?

IanMayo:
This is a good piece of work @saulhidalgoaular 

I've looked at it a few times, and was expecting a more complex solution :-)

org.mwc.cmap.core/src/org/mwc/cmap/core/ui_support/swt/SWTCanvasAdapter.java
saulhidalgoaular: Do you mean to improve the grammar? I will change it in a sec.

saulhidalgoaular: Solved in https://github.com/debrief/debrief/pull/5097/commits/445c6946693ffaa72be4a2121d44000db67ac7e9

saulhidalgoaular:
transparency

Tested on Windows

IanMayo (approved):
Good implementation, thanks.

PR #5101, opened 2021-05-14
Fixed by specially handling range.

### LINKED ISSUE #5086: When viewing Sensor Cuts in the Grid View there isn't a column for range
We probably need to add range as below.  But, it may not reflect `WorldDistance`.  If we can't provide a WorldDistance editor, then we need to provide another accessor that handles distances in metres.

image

We also need to handle the fact that the range could be null.

ayeshare:
Fixes #5086


org.mwc.debrief.legacy/src/Debrief/Wrappers/SensorContactWrapper.java
```diff
@@ -269,6 +269,7 @@ public SensorContactInfo(final SensorContactWrapper data) {
 						prop("Visible", "whether this sensor contact data is visible", FORMAT),
 						prop("Frequency", "the frequency measurement for this data item", OPTIONAL),
 						prop("Bearing", "bearing to target", SPATIAL),
+						prop("Range", "range to centre of solution", SPATIAL),
```
IanMayo, line 272: The label here should be `range to contact`

IanMayo:
I tried to edit a range using the GridEditor, but it threw this exception:
image

Note: when we use GridEditor to view a series of Positions, a helper function converts between the Location editor/helper and the WorldLocation field.  That may provide a useful example of how to handle storing the edited value.
image

IanMayo:
@ayeshare - the build is failing.

We can't change the behaviour of the `getRange()` method. Returning null means something (it means there isn't a range). There is probably lots of code that depends on this behaviour.

If we need to modify the behaviour, we should introduce a new getter/setter that _wraps_ `getRange()/setRange()`.  The griddable properties should refer to the new getter/setter, then the existing `getRange()` will work as expected.

IanMayo:
One last change, please.

You may (or may not) be aware of this:
* SensorContacts are typically plotted as lines from the track
* if the range is `null`, then they are plotted out to the edge of the visible viewport
* if the range is non-null, then they are plotted out to that distance.

For the above, users need to be able to set the range to null.

There are no cases where the range is zero metres.  So, I propose that in `setEditableRange`, if the value of the new range is zero then it's actually a null that gets stored.  Similarly, when rendering data, if the range is null, then zero metres is used as the value for `EditableRange`

IanMayo (approved):

PR #5104, opened 2021-05-14
5099 preference bug master pttx template

### LINKED ISSUE #5099: Error in path validation for the pptx master template.
In the preference section, there is a bug that after changing any property, it doesn't recognize the original pptx master template path like a valid path. 

More information about it here:
https://github.com/debrief/debrief/pull/5085#issuecomment-840558806

saulhidalgoaular:
Fix #5099


IanMayo:
Thanks @saulhidalgoaular  - just to help my understanding, could you give me a quick overview of what the issue is?

It's not obvious from the code.

saulhidalgoaular:
> Thanks @saulhidalgoaular - just to help my understanding, could you give me a quick overview of what the issue is?
> 
> It's not obvious from the code.

Hello @IanMayo 
Issue was that the code was not checking properly who triggered the event. Instead it was checking the property _field_editor_value_ , which is part of FieldEditor (Master class of all the editors), so, it was always returning true.

In the change now I am checking that the event is triggered by the field that belongs to the file selector.

Issue #4894, opened 2020-06-18
Operation failed Adding item to series error while inserting a row for the second time


helenayele:
I get RuntimeException thrown while inserting a row of data from grid editor for the second time without editing the first added row.

### Steps:

1. Open boat1.rep file
2. Open Grid editor view
3. From Outline view, select 120500.00 layer, this time you will get data displayed in grid editor.
4. Go to the bottom or anywhere you want and select a row , and right click  and select Insert Row context menu
5. This time it adds a duplicate row. as of the selected row as new row.
6. Try to select the previous selected row and select again insert row context menu. This time it didn't add other row instead throws the below exception.

[![Image from Gyazo](https://i.gyazo.com/82579be53415fc49728e89ddb01733a6.gif)](https://gyazo.com/82579be53415fc49728e89ddb01733a6)

### Expected result:
Instead of logging the exception silently, it should inform the user that there is a duplicate row and can't insert this row again .

java.lang.RuntimeException: Operation failed Adding item to series
	at org.mwc.cmap.grideditor.table.actons.AbstractViewerAction.handleExecutionException(AbstractViewerAction.java:43)
	at org.mwc.cmap.grideditor.table.actons.AbstractViewerAction.run(AbstractViewerAction.java:60)
	at org.eclipse.jface.action.Action.runWithEvent(Action.java:474)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:580)
	at org.eclipse.jface.action.ActionContributionItem.lambda$4(ActionContributionItem.java:412)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4131)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1056)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:3944)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3547)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1173)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1062)
[...]

PR #5105, opened 2021-05-15
5051 orm performance improvement

### LINKED ISSUE #5051: Pepys import performance
I'm importing a datafile with 4 points:
image

The data takes 9 seconds to load:
screencast
It would be useful to understand what is happening in this time, and to see if any bottlenecks can be resolved.

We do have our licences for Yourkit Java Profiler, if you think that will help.

This larger file (10k) points takes 14 seconds to load:
image

saulhidalgoaular:
Fix #5051

### DIFF
--- a/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
+++ b/org.mwc.cmap.legacy/src/MWC/GUI/Plottables.java
@@ -665,6 +665,15 @@ public class Plottables implements Plottable, Serializable, PlottablesType, Prop
 	}
 
 	/**
+	 * @return all items in our list greater or equal to fromElement and less than
+	 *         toElement
+	 */
+	public SortedSet<Editable> subSet(final Editable fromElement, final boolean fromInclusive, final Editable toElement,
+			final boolean toInclusive) {
+		return _thePlottables.subSet(fromElement, fromInclusive, toElement, toInclusive);
+	}
+
+	/**
 	 * @return all items in our list greater or equal to fromElement
 	 */
 	public SortedSet<Editable> tailSet(final Editable fromElement) {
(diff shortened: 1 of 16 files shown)


saulhidalgoaular:
Good morning @IanMayo .

There is something I saw that I would like you to confirm it is conceptually correct.

In the bug https://github.com/debrief/debrief/issues/4894 , I realized that the grid was trying to insert an Item with the same time stamp (because it was a copy of one of the items), so it was throwing an exception. So, I have created the class `FixWrapperCollisionCheck` to modify the item that is being inserted, changing the time by adding 1 second in case it has the exact same time. All good.

However, I have found in the sample database some datafiles that contains Fixes with the exact timestamp. Currently the `FixWrapperCollisionCheck` interprets it like an error, and "correct" it by adding 1 second later to the fix. Is this behavior the expected? CAME2_UK has 11521 repeated timestamps, which are being added with a difference of +1 sec.

Debrief without the CollisionCheck would simply ignore the insertion.

saulhidalgoaular:
oh @IanMayo , I didn't mention it, but the import time has been reduced from 7 minutes to 1.2 seconds.

IanMayo:
> oh @IanMayo , I didn't mention it, but the import time has been reduced from 7 minutes to 1.2 seconds.

WHAT?  7 minutes to 1.2 seconds!  wow!

IanMayo:
> CAME2_UK has 11521 repeated timestamps
I thought that track was only around 11521 points long

Anyway - I'm happy with your strategy in general.  But, could we add one millisecond instead of one second?

For a track with one second updates, the handler will end up incrementing the time of every point.  Aaah, that would make it look like there were 11k duplicates 🤣

saulhidalgoaular:
Hello @IanMayo .

Sorry, I was wrong, I wasn't adding 1 second, I was adding 1 millisecond already.

IanMayo (approved):
Amazing improvement, thanks @saulhidalgoaular

PR #5106, opened 2021-05-15
5100 Wrong class to load resource.

### LINKED ISSUE #5100: Potential source code error
> ./org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/model/bean/custom/ContactFastMode.java
> 805 >--->--->---final String query = OSUtils.readFile(ContactFastMode.class, getStateQuery(getAlgorithmType()));$
> The contact bean class is used to read pepys."States" records. This might be a valid reuse scenario. But, if it's an oversight, we need to rectify it.

> The above one looks to be a case where a wrong bean file (pepys."Contacts") is used to read pepys."States" records. But as said, this could be a valid reuse but with confusing names.

There's a chance that States & Contacts are being confused.  Could you check, please?

saulhidalgoaular:
Fix #5100


saulhidalgoaular:
It was not harmful, but I have fixed it.

IanMayo (approved):

PR #5110, opened 2021-05-18
added preferences to save date time

### LINKED ISSUE #5096: Pepys Import dialog - reuse dates
The Pepys Import Dialog includes date fields to set the start/end filters.

image

These are currently set to a 20 year period. The start date is _always_ wrong for our analysts.

It would be better if the start/end dates were persistent, stored across Debrief sessions.

I guess this means storing them in prefs.  Hmm, maybe it's actually in memento - I can't remember.

ayeshare:
Fixes #5096

### DIFF
--- a/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/controller/PepysImportController.java
+++ b/org.mwc.debrief.pepys/src/main/java/org/mwc/debrief/pepys/controller/PepysImportController.java
@@ -333,6 +333,8 @@ public class PepysImportController {
 								updateAreaView2Model(model, view);
 								model.apply();
 								view.getImportButton().setEnabled(false);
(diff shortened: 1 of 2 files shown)


IanMayo:
Thanks @ayeshare , couple of things:
1. Once you're happy it's working you should delete the `system.out` debug lines, please
2. If an exception occurs, it should be logged via `CorePlugin.logError` - so that it gets into the user's Error Log and we can find out more details
3. There is a date-format for parsing the data preference, but not for writing it.  I _think_ the `DateFormat` should be specified/created somewhere, and that format be used for both reading and writing.

IanMayo (changes requested):
Changes requested [here](https://github.com/debrief/debrief/pull/5110#issuecomment-843129667)

IanMayo:
Hello @ayeshare - just had some more feedback.  They'd like to use 24-hour times, please.  Does the editor control support 24-hour clocks?

IanMayo (approved):

PR #5130, opened 2021-08-17
5122 Adding missing files to the binary version

### LINKED ISSUE #5122: Pepys database connection issue
Analysts are pulling data into Debrief from Pepys, using the Pepys Import tool.  They're in version 3.1.30, dated `2021/05/30`

The `Test connection` function works, but an error is thrown when the `Apply` button is clicked for filtering.

[exact message to follow]

I've looked at the error log.  A "normal" series of database import tracking logging message are shown, including a reference to using the `Fast Mode` algorithm.

At the top of the error log is a NPE in `OSUtils.getInputStreamResource (OSUtils.java#56)`

Note: we know that the Pepys database is up & running - since it could be verified in PGAdmin.  
Note: there is a current issue with `Pepys-Admin`, which is showing an Alembic migration error.  But, Debrief doesn't use Alembic.

saulhidalgoaular:
Fix #5122

### DIFF
--- a/org.mwc.debrief.pepys/build.properties
+++ b/org.mwc.debrief.pepys/build.properties
@@ -18,4 +18,8 @@ bin.includes = META-INF/,\
                measurements.sql,\
                contacts.sql,\
                comments.sql,\
-               states.sql
+               states.sql,\
+               statesProc.sql,\
+               measurementsProc.sql,\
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
Briliant. And obvious (with hindsight  😉  )

PR #5138, opened 2021-10-04
5119 stacked dots click

### LINKED ISSUE #5119: Unable to select points on very large Stacked Dots plot
When doing single-sided reconstruction, analysts are trying to minimise the errors on the Stacked Dots plot.

While doing this, they may select some data points to delete, if the errors are really large.

Analysts have found it's not possible to select data-points on the far-right of the plot.

Note the unusual workbench layout in the following animation. The users use huge monitors, this layout is a way of reproducing the width of the plot.

screencast

More [detailed](https://i.gyazo.com/d30602ee54cdeda1b4138a7b990d0cb1.mp4) video.

The issue only arises when then the graph is really, really wide, something like 2000 pixels wide.

That makes me think the code to convert from screen to data coords isn't quite right - and some scaling factor is wrong.

The issue may be in `BaseStackedDotsView.chartMouseClicked()` - possibly in the logic for `findSelectedPlot`.

IanMayo:
Fixes #5119


IanMayo:
Superceded by #5139

PR #5147, opened 2022-01-26
5146 eclipse update

### LINKED ISSUE #5125: Issue with EASE JS Scripting
Debrief is now building, with EASE 0.8.0 (updated from 0.6.0).

But, when I try to run a script like `filter_matching_symbols`, the JS code calls over, reporting that `loadModule` is not defined:
image

This could have been a change in EASE.

If we can successfully implement Python support, then it would be fine to deprecate JS support, and leave this issue unresolved.

### LINKED ISSUE #5146: Update Debrief to Eclipse 2021-12
Update that is expected to have this bugfix
https://bugs.eclipse.org/bugs/show_bug.cgi?id=571990
and then having Eclipse Ease working again

saulhidalgoaular:
Fix #5146 
Fix #5125


saulhidalgoaular:
Sure. I will remove them.

.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
IanMayo, line 1: Hmm, this is introducing a new top-level project.  Did you mean to do that @saulhidalgoaular ?

saulhidalgoaular: This is a good question. I deleted it many times, but the Eclipse was generating it again automatically. I saw this behaviour for first time in Eclipse 4.22 (current update). At the end, I decided to push it, so it will not be generated again. I will dedicate tonight some time to research in detail what is the idea of this file.

IanMayo: In a Slack conversation we agreed to include the new `.project` file.

IanMayo (approved):
CI build working, great.

Issue #5148, opened 2022-02-11
Add new bookmark function, to also store viewport


IanMayo:
From the time-controller drop-down menu we can add a bookmark for the current DTG.

It would be useful to also add the ability to bookmark both the DTG _and_ the viewport.

So - add a new type of bookmark which also stores the viewport bounds.

When the user clicks on it - move to both the specified DTG and viewport.

ayeshare:
https://github.com/debrief/debrief/pull/5153

PR #5151, opened 2022-03-04
5148 - changes to add dtgandviewport as bookmark

ayeshare:
Supports #5148 

image (1)
image (2)


IanMayo:
Thanks @ayeshare 

I think it would be useful for the users to see the type of each bookmark.  I _guess_ we could modify `description` to be [DTG] or [DTG]+[Viewport].  The detailed date-time values would stay in the `location` column.

IanMayo:
Hello @ayeshare - nice work so far. But, there seems to be some inconsistency in what happens when bookmarks are clicked.  I have two that are just for DTG and two for DTG+Viewport.  When I'm zoomed out all the way the viewport ones work as expected. But, when I'm zoomed in, they don't seem to work.  But, changing DTG moves the viewport.

Video attached.
video

IanMayo:
Closing. Superceded by #5153

PR #5153, opened 2022-03-09
#5148 - Store viewport with bookmark

ayeshare:
supports #5148 

Supercedes #5151, which was in wrong repo.


IanMayo:
Hello @ayeshare - it doesn't always zoom in on the viewport. but I _think_ I've spotted a pattern of behaviour.  If we have a bookmark with time and viewport, then if we are already at that time, then it doesn't zoom in.

Hmm, or it moves to time, but only changes viewport on next update. I've tried to capture it here:
7170b76fa8eba1dabe53da3f900a439b

PR #5178, opened 2023-07-10
[WIP] 5177 update eclipse

theanuradha:


IanMayo:
Thanks for this @theanuradha - I can see it's a complex piece of work.

I just noticed this in `debriefng.product`.  I guess we need to update the JRE versions for `linux` and `win` too, right?
image

PR #5184, opened 2024-02-14
5183 corrupt ais

### LINKED ISSUE #5183: Robustly handle corrupted AIS messages
The AIS parser is dropping the whole file if one line is corrupt.

It would be better to log a warning, and carry on importing.

IanMayo:
Fixes #5183


IanMayo:
Merging this, so we can include it in a new release, and test.

PR #5203, opened 2024-12-05
Keep existing units when end of vector dragged.

### LINKED ISSUE #5202: Vector units lost on drag
A user has a Vector on the plot.  The range is specified in a particular set of units (such as `km`).

When they drag the `end` of the vector the range is re-calculated on drop.  But, the range switches to units in degrees - the original units are lost.

Debrief should keep the original units used to specify the Distance.

IanMayo:


IanMayo:
Close PR. Wrong fix pulled in.

PR #5211, opened 2025-04-01
misc: fix spelling in Java files

jbampton:


IanMayo:
Thanks John.  I won't be merging these typo fixes because of either:
* they occur in third-party code that we aren't responsible for
* they do not affect user functionality.

